PR #46, opened 2018-02-13
"Groups" is ambiguous

mnot:
It first reads as if MLS allows groups to communicate with each other, which I don't *think* is the intent.


beurdouche:
Thanks Mark ;)

PR #48, opened 2018-03-17
Remove GroupInitKey from Handshake; add to GroupAdd

bifurcation:
In implementing MLS, I found that having a `GroupInitKey` attached to all `Handshake` messages was almost never useful.  There's also a dangerous ambiguity between whether the `GroupInitKey` for a `Handshake` represents the state of the group before or after applying the change in the message.  "After" is the answer you want if you want to use historical `Handshake` messages as a way of distributing `GroupInitKey`s, but that requires emulating the change when signing the message -- and doing the emulation might rely on the `GroupInitKey`!

Instead of all that complication, this PR just removes the `GroupInitKey` from `Handshake` and adds it to `GroupAdd`, where its construction and utility are both obvious.

### DIFF
--- a/draft-barnes-mls-protocol.md
+++ b/draft-barnes-mls-protocol.md
@@ -817,12 +817,6 @@ identity key and a fresh leaf key pair).
 * The frontier of the identity tree, as a sequence of hash values
 * The frontier of the ratchet tree, as a sequence of public keys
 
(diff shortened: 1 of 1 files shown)


bifurcation:
This PR is overtaken by the changes in #59

PR #54, opened 2018-07-16
Message Protection

beurdouche:
This is one of the options regarding Message Protection discussed in Montreal...

### DIFF
--- a/draft-barnes-mls-protocol.md
+++ b/draft-barnes-mls-protocol.md
@@ -1463,6 +1463,33 @@ a change of the Group Secret. Hence this change MUST be applied before encryptin
 any new Application message. This is required for obvious confidentiality reasons
 regarding who can encrypt and decrypt Application messages.
 
+Updating the Application secret and deriving the associated AEAD key and IV is
+performed according the the following Application key schedule:
+
+## Application Key Schedule {#key-schedule-application}
+
+~~~~~
+           Application Secret [n-1]
+                     |
+                     +--> HKDF-Expand-Label(.,"mls app key", [sender]_id, key_length)
+                     |    = Application AEAD Key [n-1]
+                     |
+                     +--> HKDF-Expand-Label(.,"mls app iv", [sender]_id, key_length)
+                     |    = Application AEAD IV [n-1]
+                     |
+                     V
+           HKDF-Expand-Label(.,"mls app upd", App_Context, Hash.length)
+                     |
+                     V
+           Application Secret [n]
+                     |
+                     +--> HKDF-Expand-Label(.,"mls app key", [sender]_id, key_length)
+                     |    = Application AEAD Key [n]
+                     |
+                     +--> HKDF-Expand-Label(.,"mls app iv", [sender]_id, key_length)
+                          = Application AEAD IV [n]
+~~~~~
+
 ## Updating the Application Secret
 
 Each Application secret MUST be updated after each message to provide
@@ -1512,10 +1539,10 @@ The traffic keying material is generated from an input traffic secret value usin
   Hash([sender]_public_identity_key)
 
 [sender]_write_key =
-  HKDF-Expand-Label(Application Secret,"mls app key", [sender]_id, key_length)
+  HKDF-Expand-Label(Application_Secret,"mls app key", [sender]_id, key_length)
 
 [sender]_write_iv  =
-  HKDF-Expand-Label(Application Secret,"mls app iv", [sender]_id, iv_length)
+  HKDF-Expand-Label(Application_Secret,"mls app iv", [sender]_id, iv_length)
 ~~~~
 
 The identity of the sender MUST be binded to the keys and IV used for
--- a/draft-barnes-mls-protocol.md
+++ b/draft-barnes-mls-protocol.md
@@ -957,7 +957,7 @@ proceeds as shown in the following diagram:
                Init Secret [n-1] (or 0)
                      |
                      V
-Update Secret -> HKDF-Extract = Group Secret
+Update Secret -> HKDF-Extract = Epoch Secret
                      |
                      +--> Derive-Secret(., "mls add", ID, Epoch, Msg)
                      |       |
@@ -1535,8 +1535,9 @@ Application Secret changes.
 
 ## Message Encryption and Decryption
 
-The Group participants should use the Application AEAD Keys to AEAD encrypt
-and decrypt their Application message as follows:
+The Group participants should use the Application AEAD Keys associated with
+the negotiated MLS ciphersuite to AEAD encrypt and decrypt their
+Application messages as follows:
 
 ~~~~~
     struct {
@@ -1553,9 +1554,25 @@ and decrypt their Application message as follows:
 
 ~~~~~
 
-The classical AEAD primitives MUST be AES_GCM and CHACHA20_POLY1305.
-These SHOULD be replaced by a PQ scheme when the Group Key Establishment mechanism
-provides strong security guarantees against a Quantum adversary.
+[[ OPEN ISSUE: Should the padding be required for Application messages ?
+It is unclear if it is necessary or not for the Handshake messages.
+Can an adversary get more that the position of a participant in the tree
+without padding ?
+
+Padding of Application messages MUST be enforced to provide some resistance
+to traffic analysis techniques over encrypted traffic.
+{{?CLINIC=DOI.10.1007/978-3-319-08506-7_8}}
+{{?HCJ16=DOI.10.1186/s13635-016-0030-7}}
+
+While MLS is less suceptible to serve the same payload multiple time across
+a lot of ciphertexts than traditionnal web servers, it might still provide
+the attacker enough information to mount an attack. If Alice asks Bob:
+"When are we going to the movie ?" the answer "Wednesday" might be leaked
+by the ciphertext length to an adversary expecting Alice to provide Bob
+with a day of the week at some point in their discussion. To improve protection
+against timing side-channels, if the padding mechanism is used it MUST be
+implemented in a "constant-time" at the MLS layer and above. ]]
+
 
 # Security Considerations
 
--- a/draft-barnes-mls-protocol.md
+++ b/draft-barnes-mls-protocol.md
@@ -1483,7 +1483,7 @@ Application_Secret_N+1 =
 ~~~~
 
 The Application context provided together with the previous Application secret
-is used to bind the content of the message with the next key and add some freshness.
+is used to bind the Application messages with the next key and add some freshness.
 
 For MLS, the usability probably requires to keep the Application secrets and
 keys for a certain amount of time to retain the ability to decrypt messages
@@ -1491,10 +1491,11 @@ possibly in transit while the updating being done.
 Note that keeping these secrets will considerably weaken the cryptographic
 security guarantees expected at the protocol level.
 
-[[OPEN ISSUE: Should the App_Context be a hash of the Application Data ?
-Hashing all the data is obviously very costly, an other solution could be
-to add a pseudo-random contribution to each message and only hash these.
-It would be surprising if the message counter was enough.]]
+[[OPEN ISSUE: It might be enough to use the message counter for App_Context
+to be enough, this would be more conveniant.
+Hashing all the data is obviously very costly and prevents from encrypt in
+parallel, an other solution could be to add a pseudo-random contribution to
+each message and only hash these. ]]
 
 ## Application AEAD Key Calculation
 
--- a/draft-barnes-mls-protocol.md
+++ b/draft-barnes-mls-protocol.md
@@ -1557,21 +1557,27 @@ Application messages as follows:
 [[ OPEN ISSUE: Should the padding be required for Application messages ?
 It is unclear if it is necessary or not for the Handshake messages.
 Can an adversary get more that the position of a participant in the tree
-without padding ?
+without padding ? Should the base ciphertext block length be negotiated or
+is is reasonnable to allow leaking an interval for the length of the plaintext
+by allowing to send a variable number of ciphertext blocks ? ]]
 
-Padding of Application messages MUST be enforced to provide some resistance
-to traffic analysis techniques over encrypted traffic.
+Padding of Application messages SHOULD be enforced to provide some resistance
+against traffic analysis techniques over encrypted traffic.
 {{?CLINIC=DOI.10.1007/978-3-319-08506-7_8}}
 {{?HCJ16=DOI.10.1186/s13635-016-0030-7}}
-
-While MLS is less suceptible to serve the same payload multiple time across
+While MLS might be less suceptible to serve the same payload multiple time across
 a lot of ciphertexts than traditionnal web servers, it might still provide
 the attacker enough information to mount an attack. If Alice asks Bob:
 "When are we going to the movie ?" the answer "Wednesday" might be leaked
 by the ciphertext length to an adversary expecting Alice to provide Bob
-with a day of the week at some point in their discussion. To improve protection
-against timing side-channels, if the padding mechanism is used it MUST be
-implemented in a "constant-time" at the MLS layer and above. ]]
+with a day of the week at some point in their discussion.
+
+Similarily to TLS 1.3, the MLS messages MUST be padded before AEAD encryption
+with zero-valued bytes. Upon AEAD decryption, the length field of the plaintext
+is used to compute the number of bytes to be removed from the plaintext to get the
+correct data.
+If the padding mechanism is used to improve protection against timing side-channels,
+it MUST be implemented in a "constant-time" at the MLS layer and above.
 
 
 # Security Considerations
--- a/draft-barnes-mls-protocol.md
+++ b/draft-barnes-mls-protocol.md
@@ -957,19 +957,18 @@ proceeds as shown in the following diagram:
                Init Secret [n-1] (or 0)
                      |
                      V
-Update Secret -> HKDF-Extract = Epoch Secret
+Update Secret -> HKDF-Extract = Group Secret
                      |
-                     |
-                     +--> Derive-Secret(., "msg", ID, Epoch, Msg)
-                     |       = message_master_secret
-                     |
-                     +--> Derive-Secret(., "add", ID, Epoch, Msg)
+                     +--> Derive-Secret(., "mls add", ID, Epoch, Msg)
                      |       |
                      |       V
                      |    Derive-Key-Pair(.) = Add Key Pair
                      |
+                     +--> Derive-Secret(., "mls app", ID, Epoch, Msg)
+                     |       = Application Secret
+                     |
                      V
-               Derive-Secret(., "init", ID, Epoch, Msg)
+               Derive-Secret(., "mls init", ID, Epoch, Msg)
                      |
                      V
                Init Secret [n]
@@ -1435,44 +1434,128 @@ all arrive at the following state:
 
 # Message Protection
 
-[[ OPEN ISSUE: This section has initial considerations about message
-protection.  This issue clearly needs more specific recommendations,
-possibly a protocol specification in this document or a separate
-one. ]]
+The primary purpose of the handshake protocol is to enable an authenticated
+group key exchange among participants.  In order to protect Application messages sent
+among those participants, an application will need to specify how messages
+are protected. The message protection layer is somewhat independant of the chosen
+group key-establishment mechanism as long as this mechanism provides the
+expected cryptographic security properties for the Message Protection Layer.
 
-The primary purpose of this protocol is to enable an authenticated
-group key exchange among participants.  In order to protect messages sent among
-those participants, an application will need to specify how messages are protected.
+All Application messages SHOULD be protected with an Authenticated-Encryption
+with Associated-Data (AEAD) encryption scheme using AEAD keys derived from the
+Application secret (ie. see the key schedule figure).
+Note that "Authenticated" in this context is not mean messages are known to
+be sent by a specific participant but only from a legitimate member of the group.
+To obtain non-repudiability, Handshake messages MUST use asymmetric signatures
+to strongly authenticate the sender of a message, Application messages SHOULD
+use such signatures to provide the same property.
 
-For every epoch, the root key of the ratcheting tree can be used to
-derive key material for symmetric operations such as encryption/AEAD and MAC;
-AEAD or MAC MUST be used to ensure that the message originated from a member
-of the group.
+For each change to the Group root secret of the ratcheting tree, participants
+MUST derive new Application secret.
+The main reason for that is to make sure that only the intended participants,
+according to the latest group update, will be able to derive the authenticated
+encryptions keys from the Application secret.
 
-In addition, asymmetric signatures SHOULD be used to authenticate the sender
-of a message.
+Application secrets can be independantly refreshed on a regular basis. This is
+for two main reasons: the first is to improve FS and PCS guarantees for the messages if
+an Application AEAD key gets compromised; the second is to avoid key exhaustion
+from AEAD encrypting more data under the same keys than expected by the encryption scheme.
+The policy to update this secret is currently left to the application but it
+is RECOMMENDED for this secret to be updated with every message.
+Since Application AEAD keys are automatically updated at each Group operation,
+it is not expected for the Clients to be very agressive on independant updates.
+In all cases, participant MUST refresh their Application secret before
+exhaustion of the associated AEAD keys.
 
-In combination with server-side enforced ordering, data from previous messages
-is used (as a salt when hashing) to:
+Note that each change to the Group through a Handshake message will cause
+a change of the Group Secret. Hence this change MUST be applied before encrypting
+any new Application message. This is required for obvious confidentiality reasons
+regarding who can encrypt and decrypt Application messages.
 
- * add freshness to derived symmetric keys
- * cryptographically bind the transcript of all previous messages with the current
-   group shared secret
+[[OPEN ISSUE: This Key Schedule being the one of the most important part of the protocol
+its security should be formalized and be as close as possible of the TLS 1.3 Key Schedule.]]
 
-Possible candidates for that are:
+[[OPEN ISSUE: Should we maintain a Transcript hash of the Application Data ?
+It would be surprising if the message counter was enough.]]
 
- * the key used for the previous message (hash ratcheting)
- * the counter of the previous message (needs to be known to new members of the group)
- * the hash of the previous message (proof that other participants saw the same history)
+## Updating the Application Secret
 
-The requirement for this is that all participants know these values.
-If additional clear-text fields are attached to messages (like the counter), those
-fields MUST be protected by a signed message envelope.
+As each Application Secret MUST be updated to provide Application messages
+with better FS and PCS guarantees. The next generation of Application Secret
+is computed by generating an Application_Secret_N+1 from Application_Secret_N
+as described in this section then re-deriving the traffic keys as described in
+the previous section.
 
-Alternatively, the hash of the previous message can also be included as an additional
-field rather than change the encryption key. This allows for a more flexible approach,
-because the receiving party can choose to ignore it (if the value is not known, or if
-transcript security is not required).
+The next-generation Application_Secret_N+1 can happen independantly from an update
+triggered by a Group Secret change and is computed as:
+
+~~~~
+Application_Secret_N+1 =
+    HKDF-Expand-Label(Application_Secret_N,"mls app upd","", Hash.length)
+~~~~
+
+Once Application_Secret_N+1 and its associated traffic keys have
+been computed, implementations SHOULD delete Application_Secret_N and
+its associated traffic keys as soon as possible. For MLS, the usability
+probably requires to keep the Application secrets and keys for a certain
+amount of time to retain the ability to decrypt messages possibly in transit
+while the updating being done. Note that keeping these secrets will considerably
+weaken the cryptographic security guarantees expected at the protocol level.
+
+## Application AEAD Key Calculation
+
+The Application AEAD keying material is generated from the following
+input values:
+
+- The Application Secret value
+- A purpose value indicating the specific value being generated and by who.
+- The length of the bytestream being generated
+
+The traffic keying material is generated from an input traffic secret value using:
+
+~~~~
+[sender]_write_key =
+    HKDF-Expand-Label(Application Secret,[sender]+"key","",key_length)
+
+[sender]_write_iv  =
+    HKDF-Expand-Label(Application Secret,[sender]+"iv","",iv_length)
+~~~~
+
+\[sender] denotes the identifier of the Group participant and + the
+concatenation operator.
+
+Note that since, the Application keys are dependant on the identity of
+the sender, the label caracterizing this identity MUST be part of the
+Associated-Data when performing AEAD encryption.
+As previously noted, it is RECOMMENDED to refresh the application secret
+and the associated keys after each message.
+
+All the traffic keying material is recomputed whenever the underlying
+Application Secret changes.
+
+## Message Encryption and Decryption
+
+The Group participants should use the Application AEAD Keys to AEAD encrypt
+and decrypt their Application message as follows:
+
+~~~~~
+    struct {
+        uint32 length;
+        opaque content[length];
+        uint8 zeros[length_of_padding];
+    } MLSPlaintext;
+
+    struct {
+        uint32 key_id;
+        uint32 length;
+        opaque encrypted_data[MLSCiphertext.length];
+    } MLSCiphertext;
+
+~~~~~
+
+The classical AEAD primitives MUST be AES_GCM and CHACHA20_POLY1305.
+These SHOULD be replaced by a PQ scheme when the Group Key Establishment mechanism
+provides strong security guarantees against a Quantum adversary.
 
 # Security Considerations
 
--- a/draft-barnes-mls-protocol.md
+++ b/draft-barnes-mls-protocol.md
@@ -965,7 +965,7 @@ Update Secret -> HKDF-Extract = Epoch Secret
                      |    Derive-Key-Pair(.) = Add Key Pair
                      |
                      +--> Derive-Secret(., "mls app", ID, Epoch, Msg)
-                     |       = Application Secret
+                     |    = Application Secret [0]
                      |
                      V
                Derive-Secret(., "mls init", ID, Epoch, Msg)
@@ -1447,9 +1447,15 @@ To obtain non-repudiability, Handshake messages MUST use asymmetric signatures
 to strongly authenticate the sender of a message; Application messages SHOULD
 use the signature scheme defined by the ciphersuite to provide the same property.
 
-The Application secret MUST be rotated for each Application message, independently
+Each participant maintains his own chain of Application secrets, where the first
+one is derived based on a secret chained from the Epoch secret.
+As shown in {{#key-schedule}}, the initial Application secret is binded to the
+identity of each participant to avoid collisions and allow support for decryption
+of reordered messages.
+
+Subsequent Application secrets MUST be rotated for each Application message, independently
 from the Group secret updates, in order to provide stronger cryptographic
-security properties for messages. This secret is used to derive AEAD encryption
+security properties for messages. This secret is then used to derive AEAD encryption
 keys and IVs used to encrypt and decrypt Application messages.
 Since Application AEAD keys are also automatically updated at each Group operation,
 the AEAD key exhaustion bound applies on a per message basis.
@@ -1464,7 +1470,7 @@ regarding who can encrypt and decrypt Application messages.
 
 ## Updating the Application Secret
 
-Each Application secret MUST be updated after each message to provide
+The Application secret MUST be updated after each message to provide
 better cryptographic security guarantees, hence:
 
 - Senders MUST use the generation N+1 of the application secret, where N is
@@ -1496,25 +1502,21 @@ input values:
 
 - The Application Secret value;
 - A purpose value indicating the specific value being generated;
-- An identifier associated to the identity of the sender;
 - The length of the key being generated.
 
 The traffic keying material is generated from an input traffic secret value using:
 
 ~~~~
-[sender]_id =
-  Hash([sender]_public_identity_key)
-
 [sender]_write_key =
-  HKDF-Expand-Label(Application_Secret,"mls app key", [sender]_id, key_length)
+  HKDF-Expand-Label(Application_Secret,"mls app key", "", key_length)
 
 [sender]_write_iv  =
-  HKDF-Expand-Label(Application_Secret,"mls app iv", [sender]_id, iv_length)
+  HKDF-Expand-Label(Application_Secret,"mls app iv", "", iv_length)
 ~~~~
 
-The identity of the sender MUST be binded to the keys and IV used for
-AEAD encryption. This identifier should the the Hash of the public identity
-key of the used for signatures.
+Note, that because the identity of the participant using the keys to send data
+is included in the initial Application Secret, all successive updates to the
+Application secret will implicitely inherit this ownership.
 
 All the traffic keying material is recomputed whenever the underlying
 Application Secret changes.
@@ -1524,13 +1526,16 @@ be summerized as the following Application key schedule:
 
 ## Application Key Schedule {#key-schedule-application}
 
+Each participant Application secret chain looks as follows after the initial
+derivation:
+
 ~~~~~
            Application Secret [n-1]
                      |
-                     +--> HKDF-Expand-Label(.,"mls app key", [sender]_id, key_length)
+                     +--> HKDF-Expand-Label(.,"mls app key", "", key_length)
                      |    = Application AEAD Key [n-1]
                      |
-                     +--> HKDF-Expand-Label(.,"mls app iv", [sender]_id, key_length)
+                     +--> HKDF-Expand-Label(.,"mls app iv", "", iv_length)
                      |    = Application AEAD IV [n-1]
                      |
                      V
@@ -1539,10 +1544,10 @@ be summerized as the following Application key schedule:
                      V
            Application Secret [n]
                      |
-                     +--> HKDF-Expand-Label(.,"mls app key", [sender]_id, key_length)
+                     +--> HKDF-Expand-Label(.,"mls app key", "", key_length)
                      |    = Application AEAD Key [n]
                      |
-                     +--> HKDF-Expand-Label(.,"mls app iv", [sender]_id, key_length)
+                     +--> HKDF-Expand-Label(.,"mls app iv", "", iv_length)
                           = Application AEAD IV [n]
 ~~~~~
 
--- a/draft-barnes-mls-protocol.md
+++ b/draft-barnes-mls-protocol.md
@@ -1460,7 +1460,7 @@ keys and IVs used to encrypt and decrypt Application messages.
 Since Application AEAD keys are also automatically updated at each Group operation,
 the AEAD key exhaustion bound applies on a per message basis.
 In all cases, a participant MUST NOT encrypt more than expected by the AEAD scheme
-with IV and keys generated from the same Application secrets.
+with Nonces and keys generated from the same Application secrets.
 
 Note that each change to the Group through a Handshake message will cause
 a change of the Group Secret. Hence this change MUST be applied before encrypting
@@ -1469,7 +1469,7 @@ regarding who can encrypt and decrypt Application messages.
 
 ## Application Key Schedule {#key-schedule-application}
 
-Updating the Application secret and deriving the associated AEAD key and IV can
+Updating the Application secret and deriving the associated AEAD key and nonce can
 be summerized as the following Application key schedule:
 Each participant Application secret chain looks as follows after the initial
 derivation:
@@ -1477,8 +1477,8 @@ derivation:
 ~~~~~
            application_secret_N-1
                      |
-                     +--> HKDF-Expand-Label(.,"mls app iv", [sender], iv_length)
-                     |    = write_iv_N-1_[sender]
+                     +--> HKDF-Expand-Label(.,"mls app nonce", [sender], nonce_length)
+                     |    = write_nonce_N-1_[sender]
                      |
                      +--> HKDF-Expand-Label(.,"mls app key", [sender], key_length)
                      |    = write_key_N-1_[sender]
@@ -1489,8 +1489,8 @@ derivation:
                      V
            application_secret_N
                      |
-                     +--> HKDF-Expand-Label(.,"mls app iv", [sender], iv_length)
-                     |    = write_iv_N_[sender]
+                     +--> HKDF-Expand-Label(.,"mls app nonce", [sender], nonce_length)
+                     |    = write_nonce_N_[sender]
                      |
                      +--> HKDF-Expand-Label(.,"mls app key", [sender], key_length)
                           = write_key_N_[sender]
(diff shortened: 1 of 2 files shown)


draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+
+All the traffic keying material is recomputed whenever the underlying
+Application Secret changes.
+
+
+## Message Encryption and Decryption
+
+The Group participants should use the AEAD algorithm associated with
+the negotiated MLS ciphersuite to AEAD encrypt and decrypt their
+Application messages and sign them as follows:
```
ekr: See above about should, must, etc.

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+
+~~~~~
+    struct {
+        uint32 length;
+        opaque content[length];
+        opaque signature[signature_length];
```
ekr: How do I know how long signature_length is? There's no guarantee that all algorithms will have fixed-length sigs.

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+        uint8 zeros[length_of_padding];
+    } MLSPlaintext;
+
+    struct {
+        opaque group<0..2^32-1>;
```
ekr: a 4 GB group identifier seems very long. Is this supposed to be sonething different?

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+        uint32 epoch_secret;
```
ekr: Should this just be |epoch|

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+        uint32 generation;
+        uint32 sender;
+        uint32 length;
+        opaque encrypted_content[MLSCiphertext.length];
```
ekr: Typically you would do this as opaque encrypted_content<0..2^32-1>;

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
-field rather than change the encryption key. This allows for a more flexible approach,
-because the receiving party can choose to ignore it (if the value is not known, or if
-transcript security is not required).
+The Group identifier and epoch allows a device to know which Group secret
+should be used and from which Epoch secret to start computing other secrets
+and keys. The participant identifier is used to derive the participant
+Application secret chain from the initial shared Application secret.
+The application message-counter field is used to determine which Application
+secret should be used from the chain to compute the correct AEAD keys
+before performing decryption.
```
ekr: This field does not seem to appear. Is it now called |generation|

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+
+The signature field, allows to privide strong authentication of the
+plaintext and avoid Group participants to impersonate other participants.
+Note, that this non-repudiability property does not necessarily contradict
+deniability (see {{authentication}}).
```
ekr: It looks to me like this structure would let me transplant messages between groups, right? I get that nonrepudiation has that structure in general, but it seems worse in this case, because without signing the metadata I can transplant things undetectably.

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+
+[[ OPEN ISSUE: Should the padding be required for Handshake messages ?
+Can an adversary get more that the position of a participant in the tree
+without padding ? Should the base ciphertext block length be negotiated or
+is is reasonnable to allow leaking an interval for the length of the plaintext
```
ekr: reasonable.

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+by allowing to send a variable number of ciphertext blocks ? ]]
+
+Padding of Application messages SHOULD be enforced to provide some resistance
+against traffic analysis techniques over encrypted traffic.
```
ekr: What does enforced mean here? That the receiver or the sender?

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+{{?CLINIC=DOI.10.1007/978-3-319-08506-7_8}}
+{{?HCJ16=DOI.10.1186/s13635-016-0030-7}}
+While MLS might be less suceptible to serve the same payload multiple time across
+a lot of ciphertexts than traditionnal web servers, it might still provide
```
ekr: traditional

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+"When are we going to the movie ?" the answer "Wednesday" might be leaked
+by the ciphertext length to an adversary expecting Alice to provide Bob
+with a day of the week at some point in their discussion.
+
+Similarily to TLS 1.3, the MLS messages MUST be padded before AEAD encryption
+with zero-valued bytes. Upon AEAD decryption, the length field of the plaintext
+is used to compute the number of bytes to be removed from the plaintext to get the
+correct data.
+If the padding mechanism is used to improve protection against timing side-channels,
+it MUST be implemented in a "constant-time" at the MLS layer and above.
```
ekr: How is it implemented above?

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+
+### Delayed and Reordered Application messages
+
+Since each MLSCiphertext contains the Group identifier, the epoch and a
+message counter, a participant receiving Application messages out of order
+is able to compute the correct AEAD decryption keys if he kept the Application
+secret long enough.
```
ekr: or the leaf keys, right?

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+
+For usability, MLS might require to keep the Application secrets for a
+certain amount of time to retain the ability to decrypt delayed or out of
+order messages, possibly still in transit while a decryption is being done.
+Note that keeping these secrets will considerably weaken the cryptographic
+security guarantees expected at the protocol level.
```
ekr: This is pretty vague. At least leave a placeholder for a real discussion.

draft-barnes-mls-protocol.md
```diff
@@ -1435,44 +1434,219 @@ all arrive at the following state:
...
+
+An old application secret will only be re-used if a participant fails to receive
+messages for some time (which would tell it to use a newer generation), or
+if it decides to send without first syncing up on the latest messages sent
+to the group. It should thus only be necessary for clients to retain a small
+number of application secrets, say five to ten, to deal with relatively
+transient delivery failures.
```
ekr: This whole section feels kind of rushed. What do (say) Wire and Allo use for keeping old keys?

draft-barnes-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-barnes-mls-protocol.md
beurdouche: Went for `opaque signature<0..2^32-1>;`

draft-barnes-mls-protocol.md
beurdouche: Changed to a 32 byte identifier for now

draft-barnes-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-barnes-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-barnes-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-barnes-mls-protocol.md
beurdouche: Nice catch, as you said the signature must cover the metadata (typically gid, epoch, generation).
I rephrased

draft-barnes-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-barnes-mls-protocol.md
beurdouche: I rephrased to something like "Application messages should be padded to provide some resistance to traffic analysis"

draft-barnes-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-barnes-mls-protocol.md
beurdouche: Added a TODO.

draft-barnes-mls-protocol.md
beurdouche: I rephrased and expanded to say that anything that can be used to re-derive keys should disappear as well.

draft-barnes-mls-protocol.md
beurdouche: Added a discussion item.

draft-barnes-mls-protocol.md
beurdouche: Javascript lol ?

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1434,218 @@ all arrive at the following state:
...
+Application messages MUST be protected with the Authenticated-Encryption
+with Associated-Data (AEAD) encryption scheme associated with the MLS ciphersuite.
+Note that "Authenticated" in this context is not mean messages are known to
+be sent by a specific participant but only from a legitimate member of the group.
+To authenticate a message from a particular member, signatures are required.
+Handshake messages MUST use asymmetric signatures
+to strongly authenticate the sender of a message; Application messages SHOULD
+use the signature scheme defined by the ciphersuite to provide the same property.
+
+Each participant maintains his own chain of Application secrets, where the first
```
raphaelrobert: his -> their

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1434,218 @@ all arrive at the following state:
...
-transcript security is not required).
+The Group identifier and epoch allows a device to know which Group secret
+should be used and from which Epoch secret to start computing other secrets
+and keys. The participant identifier is used to derive the participant
+Application secret chain from the initial shared Application secret.
+The application generation field is used to determine which Application
+secret should be used from the chain to compute the correct AEAD keys
+before performing decryption.
+
+The signature field, allows to privide strong authentication of the
```
raphaelrobert: provide

beurdouche:
@raphaelrobert Fixed ! Thanks Raphaël !

draft-ietf-mls-protocol.md
```diff
@@ -134,6 +134,37 @@ shared keys with costs that scale as the log of the group size.  The
 use of Merkle trees to store identity information allows strong
 authentication of group membership, again with logarithmic cost.
 
+##  Change Log
+
+RFC EDITOR PLEASE DELETE THIS SECTION.
+
+
+(*) indicates changes to the wire protocol which may require implementations
```
ekr: Doesn't basically everything here change the wire protocol

draft-ietf-mls-protocol.md
```diff
@@ -954,25 +985,24 @@ Given these inputs, the derivation of secrets for an epoch
...
                      |       V
-                     |    Derive-Key-Pair(.) = Add Key Pair
+                     |    Derive-Key-Pair(.) = add_key_pair
+                     |
+                     +--> Derive-Secret(., "mls app", ID, Epoch, Msg)
+                     |    = application_secret_[0]
                      |
                      V
-               Derive-Secret(., "init", ID, Epoch, Msg)
+               Derive-Secret(., "mls init", ID, Epoch, Msg)
```
ekr: I would just have "mls" as part of the basic label in HKDF-Expand_Label

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+key schedule is used to derive encryption keys for the Message Protection Layer.
+
+Application messages MUST be protected with the Authenticated-Encryption
+with Associated-Data (AEAD) encryption scheme associated with the MLS ciphersuite.
+Note that "Authenticated" in this context does not mean messages are known to
+be sent by a specific participant but only from a legitimate member of the group.
+To authenticate a message from a particular member, signatures are required.
+Handshake messages MUST use asymmetric signatures
+to strongly authenticate the sender of a message; Application messages SHOULD
+use the signature scheme to provide the same property.
```
ekr: I think you want to say here that you should use signatures, but the rules about what algorithm are used are kind of separate.

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+
+Each participant maintains their own chain of Application secrets, where the first
+one is derived based on a secret chained from the Epoch secret.
+As shown in {{key-schedule}}, the initial Application secret is bound to the
+identity of each participant to avoid collisions and allow support for decryption
+of reordered messages.
+
+Subsequent Application secrets MUST be rotated for each Application message, independently
```
ekr: each message sent.

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+from the Group secret updates, in order to provide stronger cryptographic
+security properties for messages. To later encrypt and decrypt Application messages,
+we use this secret to derive AEAD encryption key and nonce.
```
ekr: I'm not sure what this says.

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+In all cases, a participant MUST NOT encrypt more than expected by the AEAD scheme
+with Nonces and keys generated from the same Application secrets.
+
+[[OPEN ISSUE: The AEAD key exhaustion bound applies on a per Application
+Secret basis. This depends on the number of messages per Application secrets
+that we allow. Should we set a limit ?
+]]
```
ekr: I'm not following this. The rules here require one AEAD per key/nonce pair.

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+derivation:
 
-The primary purpose of this protocol is to enable an authenticated
-group key exchange among participants.  In order to protect messages sent among
-those participants, an application will need to specify how messages are protected.
+~~~~~
+           application_secret_N-1
+                     |
+                     +--> HKDF-Expand-Label(.,"nonce", "", nonce_length)
+                     |    = write_nonce_N-1_[sender]
```
ekr: If you use the nonce directly like this rather than as an IV mask on message seq, then you absolutely can never use the same keys twice.

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+chained from the initial one to the next generation of Epoch secret. While this
+seems safe because cryptographic operations using the application secrets can't
+affect the group init_secret, it remains to be proven correct. ]]
+
+### Updating the Application Secret
+
+When the Application Secret is updated:
+
+- Senders MUST use the generation N+1 of the application secret, where N is
+the last generation they received.
```
ekr: There is no good reason for this restriction. It doesn't make the key schedule any easier.

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+- Recipients SHOULD delete older generations of application secret and as soon
+as possible, within usability bounds.
+
+These rules imply that in most circumstances, an application secret will be
+used for exactly one message. Recipients MUST NOT enforce
+a one-message-per-application-secret limit.
```
ekr: This is still pretty confused. Because the application secrets are per-sender, and the nonce is derived directly, you absolutely must use each application secret once.

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
- * add freshness to derived symmetric keys
- * cryptographically bind the transcript of all previous messages with the current
-   group shared secret
+~~~~
+write_nonce_[sender] =
+  HKDF-Expand-Label(Application_Secret,"mls app nonce", [sender], nonce_length)
 
-Possible candidates for that are:
+write_key_[sender] =
+  HKDF-Expand-Label(Application_Secret,"mls app key", [sender], key_length)
```
ekr: This section is in conflict with the previous section.

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+The Group identifier and epoch allows a device to know which Group secret
+should be used and from which Epoch secret to start computing other secrets
+and keys. The participant identifier is used to derive the participant
+Application secret chain from the initial shared Application secret.
+The application generation field is used to determine which Application
+secret should be used from the chain to compute the correct AEAD keys
+before performing decryption.
+
+The signature field, allows to provide strong authentication of the
+plaintext and avoid Group participants to impersonate other participants.
```
ekr: This is ungrammatical/

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+The signature MUST cover the metadata information about the current state
+of the group (group identifier, epoch, generation and sender) to avoid
+cross-group attacks.
```
ekr: This is not a normative requirement, because the normative requirement is the structure below. Just move this to afterwards and then use it as an explanation for why.

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+"When are we going to the movie ?" the answer "Wednesday" might be leaked
+by the ciphertext length to an adversary expecting Alice to provide Bob
+with a day of the week at some point in their discussion.
+
+Similarily to TLS 1.3, the MLS messages MUST be padded before AEAD encryption
+with zero-valued bytes. Upon AEAD decryption, the length field of the plaintext
+is used to compute the number of bytes to be removed from the plaintext to get the
+correct data.
+If the padding mechanism is used to improve protection against timing side-channels,
+it MUST be implemented in a "constant-time" at the MLS layer and above.
```
ekr: 1. This isn't a reasonable requirement because "above" is the entire app.
2. The sender is the one who decides whether to pad but the receiver is the one you want to have constant time, so this is not an enforceable must.

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+
+[[ TODO: Describe implementation ? ]]
+
+### Delayed and Reordered Application messages
+
+Since each MLSCiphertext contains the Group identifier, the epoch and a
+message counter, a participant receiving Application messages out of order
+is able to compute the correct AEAD decryption keys if he kept long enough,
+or is able to recompute, the associated Application Secret.
```
ekr: This is ungrammatical.

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+
+For usability, MLS might require to keep the Application secrets for a
+certain amount of time to retain the ability to decrypt delayed or out of
+order messages, possibly still in transit while a decryption is being done.
+Note that keeping these secrets will considerably weaken the cryptographic
+security guarantees expected at the protocol level.
```
ekr: Why?

draft-ietf-mls-protocol.md
```diff
@@ -1435,44 +1465,230 @@ all arrive at the following state:
...
+
+[[TODO: Describe here or in the Architecture spec the details. Depending
+on which Secret or key is kept alive, the security guarantees will vary.]]
+
+An old application secret will only be re-used if a participant fails to receive
+messages for some time (which would tell it to use a newer generation), or
```
ekr: No, see above.

draft-ietf-mls-protocol.md
beurdouche: I removed this because it was a leftover of the non-monotonic handling of the counters for the group-chaining version.

draft-ietf-mls-protocol.md
beurdouche: Fixed. Thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: Reordered and made a minor rewrite to eliminate more grammatical mistakes.

draft-ietf-mls-protocol.md
beurdouche: Fair enough, I slightly changed the paragraph and moved to say that the padding removal SHOULD be implemented in constant time.

draft-ietf-mls-protocol.md
beurdouche: Thanks, I reworked the phrasing.

draft-ietf-mls-protocol.md
beurdouche: I updated the text, actually we MUST NOT keep the previous application secret when we chain, we have to retain the key/nonce for an out-of-order message that we didn't receive yet otherwise I am quite sure we loose FS.

draft-ietf-mls-protocol.md
beurdouche: I removed that part, it was due to the group-chaining variant.

draft-ietf-mls-protocol.md
```diff
@@ -1551,14 +1541,13 @@ affect the group init_secret, it remains to be proven correct. ]]
...
-the last generation they received.
+- Senders MUST monotonically increment the generation of their
+  Application secret.
 - Recipients SHOULD delete older generations of application secret and as soon
 as possible, within usability bounds.
 
 These rules imply that in most circumstances, an application secret will be
-used for exactly one message. Recipients MUST NOT enforce
-a one-message-per-application-secret limit.
+used for exactly one message. Note that while
```
ekr: Why in most cases? It should never happen.

ekr: This text generally seems like it needs work:

The rules are these:

1. Senders MUST only use each secret once
2. Receivers MUST delete an application secret once it has been used to successfully decrypt something.
3. Receivers SHOULD delete application secrets once any newer application secret has been used but MAY keep the corresponding leaf keys around for some reasonable period.

draft-ietf-mls-protocol.md
```diff
@@ -1475,27 +1471,21 @@ with Associated-Data (AEAD) encryption scheme associated with the MLS ciphersuit
...
-with Nonces and keys generated from the same Application secrets.
-
-[[OPEN ISSUE: The AEAD key exhaustion bound applies on a per Application
-Secret basis. This depends on the number of messages per Application secrets
-that we allow. Should we set a limit ?
-]]
+Subsequent Application secrets MUST be rotated for each message sent in
+order to provide stronger cryptographic security guarantees. The Application
+Key Schedule use this rotation to generate fresh AEAD encryption key and nonce
+provides to encrypt and decrypt future Application messages.
```
ekr: This is ungrammatical.

draft-ietf-mls-protocol.md
```diff
@@ -1638,6 +1621,11 @@ cross-group attacks.
     } MLSSignatureContent;
 ~~~
 
+The signature MUST cover the metadata information about the current state
+of the group (group identifier, epoch, generation and sender's Leaf index)
+to avoid Group participants to impersonate other participants and in order
+to prevent cross-group attacks.
```
ekr: My point is that this wasn't normative. It's just descriptive. So the MUST should not be capitalized.

draft-ietf-mls-protocol.md
```diff
@@ -1658,38 +1646,32 @@ the attacker enough information to mount an attack. If Alice asks Bob:
...
-with zero-valued bytes. Upon AEAD decryption, the length field of the plaintext
-is used to compute the number of bytes to be removed from the plaintext to get the
-correct data.
-If the padding mechanism is used to improve protection against timing side-channels,
-it MUST be implemented in a "constant-time" at the MLS layer and above.
+Similarily to TLS 1.3, if padding is used, the MLS messages MUST be padded
+before AEAD encryption with zero-valued bytes. Upon AEAD decryption,
+the length field of the plaintext is used to compute the number of bytes
+to be removed from the plaintext to get the correct data.
+As the padding mechanism is used to improve protection against timing
```
ekr: Is it used for that? IT's more about message size side chanels.

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: I rewrote this part. Thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: Yeah, tired I guess : ) Fixed, Thanks !

draft-ietf-mls-protocol.md
ekr: OK, but we should remove all of these

draft-ietf-mls-protocol.md
ekr: Well, this way you would get key reuse.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,226 @@ all arrive at the following state:
...
 
-The requirement for this is that all participants know these values.
-If additional clear-text fields are attached to messages (like the counter), those
-fields MUST be protected by a signed message envelope.
+- Senders MUST only use the Application Secret once and monotonically
+  increment the generation of their secret.
+
+- Receivers SHOULD delete an Application Secret once it has been used to
+  derive the corresponding AEAD key and nonce and MAY keep those around
+  for some reasonable period.
```
ekr: This doesn't seem quite right. You MUST discard those as soon as used.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,226 @@ all arrive at the following state:
...
+- The length of the key being generated.
+
+The traffic keying material is generated from an input traffic secret value using:
+
+~~~~
+write_nonce_[sender] =
+  HKDF-Expand-Label(Application_Secret,"nonce", "", nonce_length)
+
+write_key_[sender] =
+  HKDF-Expand-Label(Application_Secret,"key", "", key_length)
```
ekr: This seems to duplicate the daigram above.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,226 @@ all arrive at the following state:
...
+
+All the traffic keying material is recomputed whenever the underlying
+Application Secret changes.
+
+
+## Message Encryption and Decryption
+
+The Group participants MUST use the AEAD algorithm associated with
+the negotiated MLS ciphersuite to AEAD encrypt and decrypt their
+Application messages and sign them as follows:
```
ekr: This duplicates a statement above

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,226 @@ all arrive at the following state:
...
+
+~~~~~
+    struct {
+        uint32 length;
+        opaque content[length];
```
ekr:
```
opaque content<0..2^32-1>
```

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,226 @@ all arrive at the following state:
...
+        opaque signature<0..2^16-1>;
```
ekr: How is this actually going to work? There might be multiple signing keys, right?

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,226 @@ all arrive at the following state:
...
+~~~
+    struct {
+        uint8  group[32];
+        uint32 epoch;
+        uint32 generation;
+        uint32 sender;
+        uint32 length;
+        opaque content[length];
+    } MLSSignatureContent;
+~~~
```
ekr: You don't say how this structure is to be used.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,226 @@ all arrive at the following state:
...
+The signature field allows strong authentication of messages:
+
+~~~
+    struct {
+        uint8  group[32];
+        uint32 epoch;
+        uint32 generation;
+        uint32 sender;
+        uint32 length;
+        opaque content[length];
```
ekr:
```
opaque content<0..2^32-1>;
```

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,226 @@ all arrive at the following state:
...
+Application secret chain from the initial shared Application secret.
+The application generation field is used to determine which Application
+secret should be used from the chain to compute the correct AEAD keys
+before performing decryption.
+
+The signature field allows strong authentication of messages:
+
+~~~
+    struct {
+        uint8  group[32];
```
ekr: I would make a type for this.

draft-ietf-mls-protocol.md
beurdouche: Removed it.

draft-ietf-mls-protocol.md
beurdouche: I didn't find a good way to remove the duplication atm. Let's do this in the next iteration, someone more skilled in english than me should polish the text after anyway...

draft-ietf-mls-protocol.md
beurdouche: Fixed

draft-ietf-mls-protocol.md
beurdouche: We only have a single identity key per participant at the moment.

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: I rephrased to "The signature used in the MLSPlaintext is computed over the MLSSignatureContent
which covers the metadata ..."

draft-ietf-mls-protocol.md
beurdouche: I don't think that it is necessary to do it now; it seems quite self contained anyway...

draft-ietf-mls-protocol.md
```diff
@@ -906,7 +933,7 @@ key of the non-updated child and the ephemeral public key
 transmitted in the message.
 
 
-## Key Schedule
+## Key Schedule {#key-schedule}
```
bifurcation: This tag isn't necessary.  Kramdown automatically makes a tag of this character from the content of the heading.

draft-ietf-mls-protocol.md
```diff
@@ -954,25 +981,24 @@ Given these inputs, the derivation of secrets for an epoch
...
+                     |
+                     +--> Derive-Secret(., "app", ID, Epoch, Msg)
+                     |    = application_secret_[0]
                      |
                      V
                Derive-Secret(., "init", ID, Epoch, Msg)
                      |
                      V
-               Init Secret [n]
+               init_secret [n]
```
bifurcation: If you're going to use `application_secret_[0]`, this should be `init_secret_[n]`.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,213 @@ all arrive at the following state:
...
+
+~~~~~
+    struct {
+        opaque content<0..2^32-1>;
+        opaque signature<0..2^16-1>;
+        uint8 zeros[length_of_padding];
+    } MLSPlaintext;
+
+    struct {
+        uint8  group[32];
```
bifurcation: This isn't necessary.  In practice, it will be provided by application context, and it will be authenticated by including it under the message signature.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,213 @@ all arrive at the following state:
...
+        uint32 epoch;
+        uint32 generation;
+        uint32 sender;
+        uint32 length;
```
bifurcation: This is redundant; the vector provides its own length.

draft-ietf-mls-protocol.md
bifurcation: Assuming we're going to land #59, It seems like we can just use the `GroupState`:

```
struct {
    GroupState state; // includes epoch, group_id
    uint32 sender;
    uint32 generation;
    opaque content<0..2^32-1>;
} MLSSignatureContent;
```

If we agree that's the right path, we can just land this as-is, and I'll update #59 to patch it.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,213 @@ all arrive at the following state:
...
+Similarly to TLS 1.3, if padding is used, the MLS messages MUST be padded
+before AEAD encryption with zero-valued bytes. Upon AEAD decryption,
+the length field of the plaintext is used to compute the number of bytes
+to be removed from the plaintext to get the correct data.
+As the padding mechanism is used to improve protection against traffic
+analysis, removal of the padding SHOULD be implemented in a "constant-time"
+manner at the MLS layer and above to prevent timing side-channels that
+would provide attackers with information on the size of the plaintext.
+
+[[ TODO: Describe implementation ? ]]
```
bifurcation: I don't think this is necessary.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,213 @@ all arrive at the following state:
...
+    } MLSPlaintext;
+
+    struct {
+        uint8  group[32];
+        uint32 epoch;
+        uint32 generation;
+        uint32 sender;
+        uint32 length;
+        opaque encrypted_content<0..2^32-1>;
+    } MLSCiphertext;
```
bifurcation: For consistency with the prose above, I might call this an `Application` message, and `MLSPlaintext -> ApplicationPlaintext`.  It may be that once we get to handshake encryption, we'll want a record layer that's more TLS-like (and thus usable for encrypting both Application and Handshake messages.  But for now, this is specific to `Application` messages.

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: This is because you think of it as an ID for the conversation, while I think of it as an ID over the conversation AND a specific set of users. Also, I don't think we should depend on the Application for this but the opposite.

draft-ietf-mls-protocol.md
beurdouche: Ack, fixed.

draft-ietf-mls-protocol.md
beurdouche: Let's patch it up on your side. I am still not convinced by the `GroupState` thing, so let's wait and see what others think :)

draft-ietf-mls-protocol.md
beurdouche: Removed, Thanks !

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,210 @@ all arrive at the following state:
...
+each participant's Application secret chain looks as follows after the initial
+derivation:
 
-The primary purpose of this protocol is to enable an authenticated
-group key exchange among participants.  In order to protect messages sent among
-those participants, an application will need to specify how messages are protected.
+~~~~~
+           application_secret_N-1
+                     |
+                     +--> HKDF-Expand-Label(.,"nonce", "", nonce_length)
```
bifurcation: Actually, it occurs to me that you have a simultaneous-send issue here -- two senders can chose to transmit based on the same application secret, and then you have nonce reuse, which I hear is bad.

At the IETF meeting, we had discussed two potential approaches here:

1. Per-sender hash ratchets
2. A single hash ratchet for the group folding in sender information at key/nonce generation time

You need to do one of those here.  I continue to prefer the second one, and it is a smaller delta on what you have here.  You would just need to change the "" in the nonce/key derivations to some encoding of the sender ID.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,210 @@ all arrive at the following state:
...
+                     |    = write_nonce_N-1
+                     |
+                     +--> HKDF-Expand-Label(.,"key", "", key_length)
+                     |    = write_key_N-1
+                     |
+                     V
+           Derive-Secret(., "app upd","")
```
bifurcation: Is there a reason this is `Derive-Secret` instead of `Hkdf-Expand-Label` [as in TLS 1.3](https://tools.ietf.org/html/rfc8446#section-7.2)?

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,210 @@ all arrive at the following state:
...
+                     |
+                     V
+           application_secret_N
```
bifurcation: The diagram below this should be deleted.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,210 @@ all arrive at the following state:
...
+                     |
+                     +--> HKDF-Expand-Label(.,"nonce", "", nonce_length)
+                     |    = write_nonce_N
+                     |
+                     +--> HKDF-Expand-Label(.,"key", "", key_length)
+                          = write_key_N
+
```
bifurcation: Extra blank line

draft-ietf-mls-protocol.md
beurdouche: No they can't do it, maybe it is not clear enough but this is the per-sender chaining here, so the Participant application secret is used...

draft-ietf-mls-protocol.md
beurdouche: This is a secret to secret derivation so there is no need to expand yes ?

draft-ietf-mls-protocol.md
beurdouche: "each participant's Application secret chain looks as follows after the initial
derivation:"

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,220 @@ all arrive at the following state:
...
-group key exchange among participants.  In order to protect messages sent among
-those participants, an application will need to specify how messages are protected.
+~~~
+           application_secret_[0]
+                     |
+                     V
+           Derive-Secret(., "app upd", [sender])
+                     |
+                     V
+           application_secret_[1]
```
bifurcation: Thanks, this is clearer.  Two minor edits / quetions:

However, the subscripts should indicate that this is a fanout.  I would propose having the root be something like `application_secret` (no subscript), then have the specific application secrets be double-subscripted, e.g., `app_secret_[sender]_[0]`.

You might want to use a separate label (app sender?) for this stage as opposed to the linear updates.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,220 @@ all arrive at the following state:
...
-of a message.
+Updating the Application secret and deriving the associated AEAD key and nonce can
+be summarized as the following Application key schedule where
+each participant's Application secret chain looks as follows after the initial
+derivation:
 
-In combination with server-side enforced ordering, data from previous messages
-is used (as a salt when hashing) to:
+~~~~~
+           application_secret_[N-1]
```
bifurcation: Same comment about subscripting

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks ! : )

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,219 @@ all arrive at the following state:
...
+~~~~~
+
+The Application context provided together with the previous Application secret
+is used to bind the Application messages with the next key and add some freshness.
+
+[[OPEN ISSUE: Context is left empty for now, it might be that using only
+the message counter is enough, this would be more convenient.
+Hashing all the data is obviously very costly and prevents from encrypt in
+parallel, an other solution could be to add a pseudo-random contribution to
+each message and only hash these. ]]
```
ekr: This is basically unreadable. I guess it's OK because it's in an OPEN ISSUE, but...

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,219 @@ all arrive at the following state:
...
+
+### Updating the Application Secret
+
+The following rules apply to an Application Secret:
 
-The primary purpose of this protocol is to enable an authenticated
-group key exchange among participants.  In order to protect messages sent among
-those participants, an application will need to specify how messages are protected.
+- Senders MUST only use the Application Secret once and monotonically
+  increment the generation of their secret.
```
ekr: You probably need to sharpen this point becauseit's a much more important security consideration than the next two.

draft-ietf-mls-protocol.md
ekr: Please make an open issue marker.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,219 @@ all arrive at the following state:
...
+~~~
 
-Alternatively, the hash of the previous message can also be included as an additional
-field rather than change the encryption key. This allows for a more flexible approach,
-because the receiving party can choose to ignore it (if the value is not known, or if
-transcript security is not required).
+The signature used in the MLSPlaintext is computed over the MLSSignatureContent
+which covers the metadata information about the current state
+of the group (group identifier, epoch, generation and sender's Leaf index)
+to avoid Group participants to impersonate other participants and in order
```
ekr: "to avoid..." -> "to prevent Group participants from impersonating"

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,219 @@ all arrive at the following state:
...
+to prevent cross-group attacks.
+
+[[ TODO: A preliminary formal security analysis has yet to be performed on
+this authentication scheme.]]
+
+[[ OPEN ISSUE: Should the padding be required for Handshake messages ?
+Can an adversary get more that the position of a participant in the tree
```
ekr: "more than"

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,219 @@ all arrive at the following state:
...
+without padding ? Should the base ciphertext block length be negotiated or
+is is reasonable to allow leaking an interval for the length of the plaintext
```
ekr: "leaking an interval" isn't clear.

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,219 @@ all arrive at the following state:
...
+by allowing to send a variable number of ciphertext blocks ? ]]
+
+Application messages SHOULD be padded to provide some resistance
+against traffic analysis techniques over encrypted traffic.
+{{?CLINIC=DOI.10.1007/978-3-319-08506-7_8}}
+{{?HCJ16=DOI.10.1186/s13635-016-0030-7}}
+While MLS might be less susceptible to serve the same payload multiple time across
```
ekr: "less susceptible to serve" is ungrammatical

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,219 @@ all arrive at the following state:
...
+a lot of ciphertexts than traditional web servers, it might still provide
+the attacker enough information to mount an attack. If Alice asks Bob:
+"When are we going to the movie ?" the answer "Wednesday" might be leaked
+by the ciphertext length to an adversary expecting Alice to provide Bob
```
ekr: leaked to the adversary by the ciphertext length

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,219 @@ all arrive at the following state:
...
+with a day of the week at some point in their discussion.
+
+Similarly to TLS 1.3, if padding is used, the MLS messages MUST be padded
+before AEAD encryption with zero-valued bytes. Upon AEAD decryption,
```
ekr: padding with zero-valued bytes before AEAD encryption

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,219 @@ all arrive at the following state:
...
+the length field of the plaintext is used to compute the number of bytes
+to be removed from the plaintext to get the correct data.
+As the padding mechanism is used to improve protection against traffic
+analysis, removal of the padding SHOULD be implemented in a "constant-time"
+manner at the MLS layer and above to prevent timing side-channels that
```
ekr: at the MLS layer and above layers

draft-ietf-mls-protocol.md
```diff
@@ -1435,48 +1461,219 @@ all arrive at the following state:
...
+would provide attackers with information on the size of the plaintext.
+
+### Delayed and Reordered Application messages
+
+Since each Application message contains the Group identifier, the epoch and a
+message counter, a participant can receive messages out of order.
+Participants can decrypt these messages if they are able to retrieve
+or recompute the correct AEAD decryption key from currently stored
+cryptographic material.
```
ekr: If they are able to retrieve ... participants can decrypt

draft-ietf-mls-protocol.md
beurdouche: I added a mention to Forward Secrecy here.

draft-ietf-mls-protocol.md
beurdouche: Done.

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, Thanks !

PR #58, opened 2018-09-10
Remove UserAdd

bifurcation:
I would like to remove the UserAdd functionality that's in the current document, and the GroupInitKey construct that supports it.  We will need to support a flow where a device can join without an explicit invite.  But that mechanism will also have to be limited,  since it leaks information out of the group.  It will be cleaner to get a GroupAdd-only system working first, then figure out how to add this back in.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -944,7 +944,7 @@ Update Secret -> HKDF-Extract = Epoch Secret
 
 In order to facilitate asynchronous addition of participants to a
 group, it is possible to pre-publish initialization keys that
-provide some public information about a user or group.  UserInitKey
+provide some public information about a user.  UserInitKey
 messages provide information about a potential group member, that a group member can use to
 add this user to a group asynchronously.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1178,18 +1097,17 @@ A group member generates this message using the following steps:
   direct path between the new leaf and the new root
 
 The public key of the ephemeral key pair is placed in the
-`ephemeral` field of the GroupAdd message.  The computed direct path
+`ephemeral` field of the Add message.  The computed direct path
 is placed in the `add_path` field.
```
suhasHere: nit: should we include info about  identity frontier and ratcher_frontier inclusion in the above para ?

draft-ietf-mls-protocol.md
```diff
@@ -982,12 +946,7 @@ In order to facilitate asynchronous addition of participants to a
 group, it is possible to pre-publish initialization keys that
 provide some public information about a user or group.  UserInitKey
```
suhasHere: since we no longer have groupInitKeys, should we remove the "or group" part of the sentence above ?

draft-ietf-mls-protocol.md
raphaelrobert: I agree that those definitions would make sense.

bifurcation: Done, though note that this whole section will get overwritten by #59  :)

PR #59, opened 2018-09-10
Authn v2

bifurcation:
This PR reflects a concrete scheme on top of #58 that implements the authentication framework discussed on the list recently.  There are a few major changes:

* Defines a `GroupState` structure that is incorporated into the key schedule and signatures
* Adds a `Welcome` message that provisions a new member with roster and tree information
* Removes the Merkle tree over group identities (and related definitions), since all participants now keep the full roster

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -712,16 +712,16 @@ enum {
 
 struct {
     opaque identity<0..2^16-1>;
-    SignaturePublicKey public_key; 
+    SignaturePublicKey public_key;
 } BasicCredential;
 
 struct {
     CredentialType credential_type;
     select (credential_type) {
-        case basic: 
+        case basic:
             BasicCredential;
 
-        case x509:  
+        case x509:
             opaque cert_data<1..2^24-1>;
     };
 } Credential;
@@ -738,7 +738,7 @@ struct {
   uint32 epoch;
   Credential roster<1..2^24-1>;
   PublicKey tree<1..2^24-1>;
-  opaque transcript_hash<0..255>;
+  GroupOperation transcript<0..255>;
 } GroupState;
 ~~~~~
 
@@ -771,13 +771,8 @@ operations:
 * The `group_id` field is constant
 * The `epoch` field increments by one for each GroupOperation that
   is processed
-* The `transcript_hash` is updated by a GroupOperation message
-  `operation` in the following way (where H is the hash function
-  for the ciphersuite in use):
-
-~~~~~
-transcript_hash_{n+1} = H(transcript_hash_{n} || operation)
-~~~~~
+* The `transcript` is updated by a GroupOperation message
+  `operation` by appending `operation` to `transcript`
 
 ## Direct Paths
 
@@ -960,15 +955,14 @@ enum {
 } GroupOperationType;
 
 struct {
-    HandshakeType msg_type;
-    select (Handshake.msg_type) {
+    GroupOperationType msg_type;
+    select (GroupOperation.msg_type) {
         case init:      Init;
         case add:       Add;
         case update:    Update;
         case remove:    Remove;
     };
 } GroupOperation;
-
 
 struct {
     uint32 prior_epoch;
@@ -1042,10 +1036,10 @@ current state using the Add message:
 struct {
   opaque group_id<0..255>;
   uint32 epoch;
-  opaque init_secret<0..255>;
-  opaque transcript_hash<0..255>;
   Credential roster<1..2^24-1>;
   PublicKey tree<1..2^24-1>;
+  GroupOperation transcript<0..255>;
+  opaque init_secret<0..255>;
   opaque leaf_secret<0..255>;
 } Welcome;
 ~~~~~
@@ -1071,7 +1065,7 @@ member:
 
 ~~~~~
 struct {
-    DirectPath add_path<1..2^16-1>;
+    DirectPath path<1..2^16-1>;
     UserInitKey init_key;
 } Add;
 ~~~~~
@@ -1113,7 +1107,7 @@ regard to the participant's prior leaf private key.
 
 ~~~~~
 struct {
-    DirectPath update_path;
+    DirectPath path;
 } Update;
 ~~~~~
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -173,7 +173,7 @@ Identity Key:
   message.
 
 Terminology specific to tree computations is described in
-{{binary-trees}}.
+{{ratchet-trees}}.
 
 We use the TLS presentation language {{!RFC8446}} to
 describe the structure of protocol messages.
@@ -339,23 +339,14 @@ A              B     ...      Z          Directory       Channel
 ~~~~~
 
 
-# Binary Trees
+# Ratchet Trees
 
-The protocol uses two types of binary tree structures:
-
-  * Merkle trees for efficiently committing to a set of group participants.
-  * Ratchet trees for deriving shared secrets among this group of
-    participants.
-
-The two trees in the protocol share a common structure, allowing us to maintain
-a direct mapping between their nodes when manipulating group membership. The
-`nth` leaf in each tree is owned by the `nth` group participant.
+The protocol uses "ratchet trees" for deriving shared secrets among
+a group of participants.
 
 ## Terminology
 
-We use a common set of terminology to refer to both types of binary tree.
-
-Trees consist of various different types of _nodes_. A node is a
+Trees consist of _nodes_. A node is a
 _leaf_ if it has no children, and a _parent_ otherwise; note that all
 parents in our Merkle or ratchet trees have precisely
 two children, a _left_ child and a _right_ child. A node is the _root_
@@ -405,66 +396,40 @@ For example, in the below tree:
           /        \
          /          \
      ABCD            EFG
-    /    \          /   \
-   /      \        /     \
-  AB      CD      EF      \
- /  \    /  \    /  \      \
-A    B  C    D  E    F      G
+    /    \          /  \
+   /      \        /    \
+  AB      CD      EF    |
+ / \     / \     / \    |
+A   B   C   D   E   F   G
+
+                    1 1 1
+0 1 2 3 4 5 6 7 8 9 0 1 2
 ~~~~~
 
-We extend both types of tree to include a concept of "blank" nodes;
-which are used to replace group members who have been removed. We
-expand on how these are used and implemented in the sections below.
+Each node in the tree is assigned an _index_, starting at zero and
+running from left to right.  A node is a leaf node if and only if it
+has an even index.  The indices for the nodes in the above tree are
+as follows:
+
+* 0 = A
+* 1 = AB
+* 2 = B
+* 3 = ABCD
+* 4 = C
+* 5 = CD
+* 6 = D
+* 7 = ABCDEFG
+* 8 = E
+* 9 = EF
+* 10 = F
+* 11 = EFG
+* 12 = G
 
 (Note that left-balanced binary trees are the same structure that is
 used for the Merkle trees in the Certificate Transparency protocol
 {{?I-D.ietf-trans-rfc6962-bis}}.)
 
-## Merkle Trees
-
-Merkle trees are used to efficiently commit to a collection of group members.
-We require a hash function, denoted H, to construct this tree.
-
-Each node in a Merkle tree is the output of the hash function,
-computed as follows:
-
-* Leaf nodes: `H( 0x01 || leaf-value )`
-* Parent nodes: `H( 0x02 || left-value || right-value)`
-* Blank leaf nodes: `H( 0x00 )`
-
-The below tree provides an example of a size 2 tree, containing identity keys
-`A` and `B`.
-
-~~~~~
-             * H(2 || H(1 || A) || H(1 || B))
-            / \
-           /   \
-H(1 || A) *     * H(1 || B)
-~~~~~
-
-In Merkle trees, blank nodes appear only at the leaves.  In computation of
-intermediate nodes, they are treated in the same way as other nodes.
-
-### Merkle Proofs
-
-A proof of a given leaf being a member of the Merkle tree consists of the value
-of the leaf node, as well as the values of each node in its copath. From these
-values, its path to the root can be verified; proving the inclusion of the leaf
-in the Merkle tree.
-
-In the below tree, we denote with a star the Merkle proof of membership for
-leaf node `A`. For brevity, we notate `Hash(0x02 || A || B)` as `AB`.
-
-~~~~~
-      ABCD
-    /      \
-  AB        CD*
- /  \      /  \
-A   B*    C    D
-~~~~~
-
-
-## Ratchet Trees
+## Ratchet Tree Nodes
 
 Ratchet trees are used for generating shared group secrets. In this
 section, we describe the structure of a ratchet tree, along with two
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -695,6 +659,46 @@ tree management to behave as for a balanced tree for programming simplicity.
...
+state of the group:
+
+~~~~~
+struct GroupState {
+  opaque group_id<...>;
+  uint32 size;
+  uint32 epoch;
+  Credential roster<...>;
+  PublicKey tree<...>;
+  opaque transcript_hash<...>;
```
raphaelrobert: `transcript_hash` is not defined anywhere AFAICT

draft-ietf-mls-protocol.md
```diff
@@ -695,6 +659,46 @@ tree management to behave as for a balanced tree for programming simplicity.
...
+* The `tree` field contains the public keys corresponding to the
+  nodes of the ratchet tree for this group.  The length of this
+  vector MUST be `2*size + 1`.
+
+When a new member is added to the group, an existing member of the
+group provides the new member with a Welcome message.  The Welcome
+message provides the information the new member needs to initialize
+its GroupState.
+
+Other group operations will have different affects on the group
```
raphaelrobert: affects -> effects

draft-ietf-mls-protocol.md
```diff
@@ -281,45 +281,9 @@ A              B              C          Directory            Channel
 
 Subsequent additions of group members proceed in the same way.  Any
 member of the group can download an InitKey for a new participant
-and broadcast a GroupAdd which the current group can use to update
+and broadcast an Add which the current group can use to update
```
suhasHere: nit: "an Add message"

draft-ietf-mls-protocol.md
```diff
@@ -1155,88 +994,81 @@ including only itself, then adding the initial participants.  This
...
+The Welcome message contains the information that the new member
+needs to initialize a GroupState object that can be updated to the
+current state using the Add message:
+
+~~~~~
+struct Welcome {
+  uint32 size;
+  opaque group_id<0..255>;
+  uint32 size;
+  uint32 epoch;
```
suhasHere: size repeated twice

draft-ietf-mls-protocol.md
```diff
@@ -655,80 +584,57 @@ public-key encryption to the public key K of the secret value S):
...
-  DH(A, DH(CD))
-   /      \
-  A       DH(CD)
- / \      /  \
-A   _    C    D
+struct GroupState {
+  opaque group_id<0..255>;
+  uint32 size;
+  uint32 epoch;
+  Credential roster<1..2^24-1>;
```
suhasHere: Wondering where is __Credential__ defined ?

draft-ietf-mls-protocol.md
```diff
@@ -655,80 +584,57 @@ public-key encryption to the public key K of the secret value S):
...
+  in the group, whether or not these slots are currently occupied.
+* The `epoch` field represents the current version of the group key.
+* The `roster` field contains credentials for the occupied slots in
+  the tree, including the identity and public key for the holder of
+  the slot.  The length of the `roster` vector MUST be equal to the
+  value of the `size` field; the n-th entry in the vector represents
+  the identity for the n-th slot, if any.
+* The `tree` field contains the public keys corresponding to the
+  nodes of the ratchet tree for this group.  The length of this
+  vector MUST be `2*size + 1`.
```
suhasHere: nit .. for readers beneifit , might be useful to explain the size formula ?

draft-ietf-mls-protocol.md
```diff
@@ -655,80 +584,57 @@ public-key encryption to the public key K of the secret value S):
...
+
+When a new member is added to the group, an existing member of the
+group provides the new member with a Welcome message.  The Welcome
+message provides the information the new member needs to initialize
+its GroupState.
+
```
suhasHere: Wonder how the Welcome message race is resolved when multiple existing members end up adding the same new member?

draft-ietf-mls-protocol.md
```diff
@@ -655,80 +584,57 @@ public-key encryption to the public key K of the secret value S):
...
+Different group operations will have different effects on the group
+state.  These effects are described in their respective subsections
+of {{handshake-messages}}.  The following rules apply to all
+operations:
+
+* The `group_id` field is constant
+* The `epoch` field increments by one on each change
```
suhasHere: reword .. each change --> each update to group state ?

draft-ietf-mls-protocol.md
```diff
@@ -655,80 +584,57 @@ public-key encryption to the public key K of the secret value S):
...
- / \      /  \
-A   _    C    D
+struct GroupState {
+  opaque group_id<0..255>;
+  uint32 size;
+  uint32 epoch;
+  Credential roster<1..2^24-1>;
+  PublicKey tree<1..2^24-1>;
+  opaque transcript_hash<0..255>;
+}
```
suhasHere: transcript_hash description is missing below

draft-ietf-mls-protocol.md
```diff
@@ -913,17 +813,15 @@ functions as defined in {{!RFC5869}}, as well as the functions
...
 Where HkdfLabel is specified as:
 
 struct {
     uint16 length = Length;
-    opaque label<7..255> = "mls10 " + Label;
-    opaque group_id<0..2^16-1> = ID;
-    uint32 epoch = Epoch;
-    opaque message<1..2^16-1> = Msg
+    opaque label<6..255> = "mls10 " + Label;
+    GroupState state = State;
```
suhasHere: Trying to understand the need for entire group state in generating the Label ?

draft-ietf-mls-protocol.md
```diff
@@ -1283,18 +1113,18 @@ The sender of a Remove message generates it as as follows:
 
 * Generate a fresh leaf key pair
 * Compute its direct path in the current ratchet tree, starting from
-  the deleted leaf (Note: In ART, this requires knowing the deleted
-  node's copath)
+  the deleted leaf
 
-An existing participant receiving a Delete message first verifies
+An existing participant receiving a Remove message first verifies
 the signature on the message, then verifies its identity proof
```
suhasHere: Remove message type isn't defined.

draft-ietf-mls-protocol.md
```diff
@@ -655,80 +584,57 @@ public-key encryption to the public key K of the secret value S):
...
-If two sibling nodes are both \_, their parent value also becomes \_.
-
-Blank nodes effectively result in an unbalanced tree, but allow the
-tree management to behave as for a balanced tree for programming simplicity.
-
+The fields in this state have the following semantics:
+
+* The `group_id` field is an application-defined identifier for the
+  group.
+* The `size` field represents the total number of key-exchange slots
```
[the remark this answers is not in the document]
bifurcation: Actually, this can be eliminated, since it can be derived from the length of the `tree` vector.

draft-ietf-mls-protocol.md
```diff
@@ -655,80 +584,57 @@ public-key encryption to the public key K of the secret value S):
...
+  in the group, whether or not these slots are currently occupied.
+* The `epoch` field represents the current version of the group key.
+* The `roster` field contains credentials for the occupied slots in
+  the tree, including the identity and public key for the holder of
```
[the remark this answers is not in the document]
bifurcation: "Public key" == "Signature public key".  ("Identity" is whatever opaque string the application assigns)  Note that because this state can change, we're not committed to using long-term identity keys.

draft-ietf-mls-protocol.md
```diff
@@ -655,80 +584,57 @@ public-key encryption to the public key K of the secret value S):
...
+  the slot.  The length of the `roster` vector MUST be equal to the
+  value of the `size` field; the n-th entry in the vector represents
```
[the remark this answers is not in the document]
bifurcation: I don't know why you think this is implementation specific.  In the current implementation, there is an "owner" of each leaf in the tree, and the participants need to agree on the leaf-to-owner assignment.  So you can compress your storage locally, but you need to be able to reconstruct this vector.

draft-ietf-mls-protocol.md
```diff
@@ -655,80 +584,57 @@ public-key encryption to the public key K of the secret value S):
...
+its GroupState.
+
+Different group operations will have different effects on the group
+state.  These effects are described in their respective subsections
+of {{handshake-messages}}.  The following rules apply to all
+operations:
+
+* The `group_id` field is constant
+* The `epoch` field increments by one on each change
+* The `transcript_hash` is updated by a KeyExchange message
```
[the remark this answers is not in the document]
bifurcation: I'll change `KeyExchange` to something else, maybe `GroupOperation`.

draft-ietf-mls-protocol.md
```diff
@@ -913,17 +813,15 @@ functions as defined in {{!RFC5869}}, as well as the functions
 defined below:
 
 ~~~~~
-Derive-Secret(Secret, Label, ID, Epoch, Msg) =
+Derive-Secret(Secret, Label, State, Length) =
```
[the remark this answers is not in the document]
bifurcation: I don't know what you mean by log.  Anything you hash in here is something the endpoints have to agree on, or else they end up at different keys.  So the idea is to include everything we want the endpoints to agree on.

draft-ietf-mls-protocol.md
```diff
@@ -1067,70 +912,64 @@ messages are exchanged throughout the lifetime of a group, whenever
...
-The input to the signature computations
-comprises the entire handshake message except for the signature
-field.
-
-The Merkle tree head to be used for validating the inclusion
-proof MUST be one that the recipient trusts to represent the current
-list of participant identity keys.
+An MLS handshake message encapsulates a specific "key exchange" message that
+accomplishes a change to the group state. It also includes a
+signature by the sender of the message over the GroupState object
```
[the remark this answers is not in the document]
bifurcation: See the authentication discussion on the list.

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
beurdouche: I think I disagree with the fact that you need to use the entire group state every time to make sure everybody agrees on the same thing. Intuitively the previous definition encompassed all group operations through the Handshake messages (`Msg`), which includes the signatures, which is enough to get the weak authentication (group-membership) property. I think it is a good practice to include incrementally what we need to be able to prove the cryptographic security properties we want and not the opposite which is just gonna be annoying for proofs. I would strongly be in favor of keeping the previous definition for now.

draft-ietf-mls-protocol.md
beurdouche: Same comment here, I prefer the previous definition for now...

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
beurdouche: Maybe I misread but I couldn't find anything about that there.

draft-ietf-mls-protocol.md
bifurcation: Good catch, I forgot to do that.

draft-ietf-mls-protocol.md
bifurcation: Added an open issue.

draft-ietf-mls-protocol.md
```diff
@@ -924,16 +882,11 @@ proceeds as shown in the following diagram:
 Update Secret -> HKDF-Extract = Epoch Secret
                      |
                      |
-                     +--> Derive-Secret(., "msg", ID, Epoch, Msg)
+                     +--> Derive-Secret(., "msg", GroupState_n)
```
beurdouche: If we don't use the `_[N]` subscript to give the precise version of the secrets, maybe just use `GroupState` instead of `GroupState_n` ?

draft-ietf-mls-protocol.md
```diff
@@ -993,61 +945,63 @@ messages are exchanged throughout the lifetime of a group, whenever
...
-authentication (of identities, rather than keys), and will need to
-be done in such a way as to prevent unknown key share attacks. ]]
+The high-level flow for processing a Handshake message is as
+follows:
+
+1. Verify that the `prior_epoch` field of the Handshake message
+   is equal the `epoch` field of the current GroupState object.
 
-[[ OPEN ISSUE: The Add and Delete operations create a "double-join"
+2. Use the `key_exchange` message to produce an updated GroupState
```
beurdouche: `key_exchange` is now called `GroupOperation`

draft-ietf-mls-protocol.md
```diff
@@ -809,6 +693,91 @@ the corresponding ART ciphersuite.
...
+## Group State
+
+Each participant in the group maintains a representation of the
+state of the group:
+
+~~~~~
+struct {
+  opaque group_id<0..255>;
+  uint32 epoch;
+  Credential roster<1..2^24-1>;
```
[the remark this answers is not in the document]
raphaelrobert: While `2^16` is roughly the order of magnitude until which MLS should work well, I agree that it should rather be `2^32` instead of `2^24`.

draft-ietf-mls-protocol.md
```diff
@@ -993,61 +945,63 @@ messages are exchanged throughout the lifetime of a group, whenever
...
+2. Use the `key_exchange` message to produce an updated GroupState
+   object incorporating the proposed changes.
+
+3. Look up the public key for slot index `signer_index` from the
+   roster in the current GroupState object (before the update).
+
+4. Use that public key to verify the `signature` field in the
+   Handshake message, with the updated GroupState object as input.
+
+5. If the signature fails to verify, discared the updated GroupState
```
raphaelrobert: discared -> discard

draft-ietf-mls-protocol.md
```diff
@@ -1074,54 +1028,82 @@ the O(N) complexity of direct initialization. ]]
 
 ## Add
 
-An Add message is sent by a group member to add a new participant
-to the group.
+In order to add a new member to the group, an existing member of the
+group must take two actions:
+
+1. Send a Welcome message to the new member
+2. Send an Add message to the group (including the new member)
```
raphaelrobert: The order is probably wrong, since the `Welcome` message should only be sent if the `Add` message succeeded. We should think some more about synchronicity issues that we want to avoid. 
Ideally the `Add` and `Welcome` message are on atomic operation in which all old and new members are notified at the same time. However since the existing member now waits for the `Add` to be successful before sending the (rather large) `Welcome`, there might be edge cases when the `Welcome` is never sent because the inviting member went offline too early.
For now I think we should keep this as an open issue. Going forward we should consider improvements, such as splitting the `Welcome` into a light header that gets reliably dispatched at the same time with the `Add` and a larger message containing the roster and the tree.

draft-ietf-mls-protocol.md
```diff
@@ -1074,54 +1028,82 @@ the O(N) complexity of direct initialization. ]]
...
-the identity tree held by the participant. The participant then updates
-its state as follows:
+the signature on the message,  then updates its state as follows:
 
-* Compute the new participant's leaf key pair by combining the leaf
-  key in the UserInitKey with the prior epoch add key pair
-* Update the group's identity tree and ratchet tree with the new
-  participant's information
+* Increment the size of the group
+* Verify the signature on the included UserInitKey
```
raphaelrobert: Obviously if the verification fails, none of the changes should be applied.

draft-ietf-mls-protocol.md
```diff
@@ -809,6 +693,91 @@ the corresponding ART ciphersuite.
...
+that the holder of the credential will use to sign MLS messages.
+Credentials MAY also include information that allows a relying party
+to verify the identity / signing key binding.
+
+~~~~~
+enum {
+    basic(0),
+    x509(1),
+    (255)
+} CertificateType;
```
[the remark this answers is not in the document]
bifurcation: Done.

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
bifurcation: I don't think this is as much of an implementation issue as you do; all the decoders out there just read the relevant bytes into a uint32_t and carry on.  But I'll change it for the moment; we can try to economize on bytes later.

draft-ietf-mls-protocol.md
bifurcation: Note that there's an open issue for this a few lines further down.  I agree that this is a tricky sync to get right.  Shoving the Welcome message in the Add message might simplify this, but the cost of sending an O(N) size message to everyone seems steep.

PR #62, opened 2018-09-19
Use nonces in ECIES derived from the shared secret 

bifurcation:
This matches the behavior in TLS, and saves space (due to not having to send the nonce).

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -850,17 +850,28 @@ encrypted secret values are computed as follows:
 * Generate an ephemeral DH key pair (x, x\*G) in the DH group
   specified by the ciphersuite in use
 * Compute the shared secret Z with the node's other child
-* Generate a fresh nonce N
+* Derive a key and nonce as described below
 * Encrypt the node's secret value using the AEAD algorithm specified
   by the ciphersuite in use, with the following inputs:
-  * Key: A key derived from Z as specified by the ciphersuite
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -852,18 +854,29 @@ encrypted secret values are computed as follows:
...
+  * Nonce: The nonce derived from Z
   * Additional Authenticated Data: The empty octet string
   * Plaintext: The secret value, without any further formatting
 * Encode the ECIESCiphertext with the following values:
   * ephemeral\_key: The ephemeral public key x\*G
-  * nonce: The random nonce N
   * ciphertext: The AEAD output
 
+~~~~~
+key = HKDF-Expand(Secret, ECIESLabel("key"), Length)
```
beurdouche: HKDF Expand Label ?

bifurcation: No, just HKDF-Expand.  HKDF-Expand-Label isn't defined here (only in TLS), and in any case, we're using a different label.

beurdouche: Ack, this is good for me then...

draft-ietf-mls-protocol.md
```diff
@@ -166,6 +166,9 @@ draft-01
 
 - Added an appendix with example code for tree math
 
+- Changed the ECIES mechanism used by TreeKEM so that it uses nonces
```
beurdouche: Though I think this is already the case...

PR #63, opened 2018-09-21
Allow bigger sizes for PQ primitives

beurdouche:
By discussing with MSR, I realized that some NIST PQ candidates can't be used easily by TLS because signatures can be bigger that `2^16` bytes.

Since we already consume a lot for bandwidth, I am not concerned at all about two more bytes to encode the length of signatures on each ciphertext so I suggest we move to `opaque signature <0..2^32-1>` instead to be future-proof.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -749,7 +749,7 @@ enum {
 } CredentialType;
 
 struct {
-    opaque identity<0..2^16-1>;
+    opaque identity<0..2^32-1>;
     SignaturePublicKey public_key;
 } BasicCredential;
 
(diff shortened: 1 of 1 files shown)


raphaelrobert:
Having more flexibility with the payload size sounds reasonable.
One day having signatures that big if however another kind of challenge.

bifurcation:
@beurdouche if you could fix the merge conflicts, I'm OK with this.

ekr:
As I said in the WG meeting, I don't think this is a good idea. If we end up having primitives this big, we are going to have lots of problems and will probably want a better architecture (e.g., signatures that batch multiple messages).

bifurcation:
I agree with EKR here.  PQ is probably going to require different designs, not just bigger crypto values.  If someone wants to make a concrete proposal for shimming in a PQ scheme here, which would run into these limits, I would be open to reconsidering this.

beurdouche:
FYI, even though this is closed this will be the basis of PQ experiment in real products, just with a different protocol version.
Ps: Please don't delete other people branches.

PR #64, opened 2018-09-26
Change from AES-GCM to AES-SIV

galadran:
[Mailing List Post](https://mailarchive.ietf.org/arch/msg/mls/gvszahIKpQ5CYJG-ARwi9MGURGo)

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -703,13 +703,13 @@ infinity (O), (2) verify that for Y = (x, y) both integers are in the correct
 interval, (3) ensure that (x, y) is a correct solution to the elliptic curve equation.
 For these curves, implementers do not need to verify membership in the correct subgroup.
 
-### TreeKEM with Curve25519, SHA-256, and AES-128-GCM
+### TreeKEM with Curve25519, SHA-256, and AES-SIV-256
 
 This ciphersuite uses the following primities:
 
 * Hash function: SHA-256
 * Diffie-Hellman group: Curve25519 {{!RFC7748}}
-* AEAD: AES-128-GCM
+* AEAD: AES-SIV-256
 
 DH and Derive-Key-Pair operations are performed in the same way as
 the corresponding ART ciphersuite.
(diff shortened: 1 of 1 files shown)


ekr:
I'm not sure I see a lot of value here. The current discipline is one key per message, so is there a special reason to believe that clients will be careless about nonces.

galadran:
> I'm not sure I see a lot of value here. The current discipline is one key per message, so is there a special reason to believe that clients will be careless about nonces.

Hi ekr, 

Using one key per message just lifts the problem to a different layer, we're relying on implementers getting the state handling right and never reusing a key. (Since if they reuse a key, they will reuse a nonce). 

Verifying correct nonce/key usage is difficult. Using implicit nonces and testing for compatibility gives us an assurance that most uses are correct, but says little about edge cases which are hard to test exhaustively. With AES-GCM, these edge cases lead to an immediate loss of authenticity and possible total compromise. With AES-SIV, these edge cases have very minimal impact as the adversary only learns if the two plaintexts were equal.

ekr:
Yes, my question is: is there a special reason to be believe that MLS implementations will get this wrong as opposed to (say) TLS 1.3 or SRTP implementations, both of which use implicit nonces.

galadran:
Yes.

- MLS is many party as opposed to two 
- MLS is highly stateful in order to achieve PFS and PCS
- MLS is asynchronous
- MLS sessions are expected to continue despite a device suddenly losing power 
- MLS is a completely new and complex protocol

I already described some of the situations in which SIV offers significant benefits over implicit nonces on the mailing list.

Finally, it seems premature to declare implicit nonces to have fixed all the nonce issues in TLS. Let's give TLS 1.3 another decade (or two) before declaring it perfect.

ekr:
On Thu, Sep 27, 2018 at 12:58 AM Dennis Jackson <notifications@github.com>
wrote:

>
>    - MLS sessions are expected to continue despite a device suddenly
>    losing power
>
> This last is the only one of these that seems particularly relevant, given
that each sender uses its own keys. I'd be interested in hearing from
implementors whether they consider this kind of state loss to be an issue.


>
> Finally, it seems premature to declare implicit nonces to have fixed all
> the nonce issues in TLS. Let's give TLS 1.3 another decade (or two) before
> declaring it perfect.
>
This misrepresents what I said, which was nothing like claiming that TLS
1.3 nonce handling was perfect.

With that said, I would again refer you to SRTP , which has used implicit
nonces for a very long time. Do you have any evidence that nonce reuse has
been a problem here?

galadran:
> With that said, I would again refer you to SRTP , which has used implicit nonces for a very long time. Do you have any evidence that nonce reuse has been a problem here?

I have no idea. Let's assume that SRTP has no nonce reuse problems. What is the conclusion?  

I've described some of the differences between synchronous, essentially stateless, 2-party protocols and MLS. I've referenced a catastrophic nonce reuse attack which implicit nonces would not have mitigated (KRACK on WPA2). There are clear scenarios where SIV provides a tangible security benefit _in addition to_  implicit nonces. 

There are reasons not to use SIV in applications like TLS, as SIV only encrypts at ~70% of the speed of GCM. However, I don't think they apply to us considering the public key operations we will be doing anyway. Equally in 2014 SIV was less widely supported, since then it has found its way into OpenSSL and BoringSSL.

grittygrease:
Following up on this. This suggestion was discussed at the first interim and there was not a lot of enthusiasm for making the change.

PR #65, opened 2018-09-27
Fix typos.

Bren2010:


beurdouche:
@Bren2010 Hi Brendan ! Would you mind please revert the last of the 3 changes. Even though the sentence is not amazing and your change correct, I find it better to keep explicit the fact that the root should not be included, instead of having to infer it from the fact it has no siblings :) Thanks !

Bren2010:
You don't have to infer it -- direct paths are defined to never include the root node in the sentence just before.

PR #66, opened 2018-10-01
Remove ART

bifurcation:
There was agreement at the MLS interim of September 2018 to focus further development on TreeKEM, not ART.  This PR implements that change.

PR #67, opened 2018-10-01
Remove double-join with the "partial tree" approach

bifurcation:
This PR updates the procedures for Add and Remove to remove double-joins.  The cost is a potential increase in the size of GroupOperation messages in the case of incomplete trees.

Depends on #66

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -746,10 +746,18 @@ state of the group:
 
 ~~~~~
 struct {
+  uint8 present;
+  switch (present) {
+    case 0: struct{};
+    case 1: T value;
+  }
+} optional<T>;
+
+struct {
   opaque group_id<0..255>;
   uint32 epoch;
-  Credential roster<1..2^32-1>;
-  PublicKey tree<1..2^32-1>;
+  optional<Credential> roster<1..2^32-1>;
+  optional<PublicKey> tree<1..2^32-1>;
   opaque transcript_hash<0..255>;
 } GroupState;
 ~~~~~
@@ -1177,17 +1185,16 @@ the signature on the message, then verifies its identity proof
 against the identity tree held by the participant.  The participant
 then updates its state as follows:
 
-* Update the roster by replacing the credential in the removed slot
-  with the credential from the sender's slot (i.e., the sender of
-  the Remove takes over the removed slot)
+* Update the roster by setting the credential in the removed slot to
+  the null optional value
 * Update the ratchet tree by replacing nodes in the direct
   path from the removed leaf using the information in the Remove message
 * Update the ratchet tree by setting to blank all nodes in the
   direct path from the removed leaf to the root
 
 The update secret resulting from this change is the secret for the
-root node of the ratchet tree after both updates.
-
+root node of the ratchet tree after the second step (after the third
+step, the root is blank).
 
 # Sequencing of State Changes {#sequencing}
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -523,7 +523,42 @@ X, then the tree will have the following structure.
 X    B     C    D
 ~~~~~
 
-### Ratchet Tree Updates
+## Blank Nodes and Resolution
+
+A node in the tree may be _blank_, indicating that no value is
+present at that node.  The _resolution_ of a node is an ordered list
+of non-blank nodes that collectively cover all non-blank descendants
+of the node.  The nodes in a resolution are ordered according to
+their indices.
+
+* The resolution of a non-blank node is a one element list
+  containing the node itself
+* The resolution of a blank leaf node is the empty list
+* The resolution of a blank intermediate node is the result of
+  concatinating the resolution of its left child with the resolution
+  of its right child, in that order
+
+For example, consider the following tree, where the "\_" character
+represents a blank node:
+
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -386,7 +388,7 @@ a group of participants.
 
 Trees consist of _nodes_. A node is a
 _leaf_ if it has no children, and a _parent_ otherwise; note that all
-parents in our Merkle or ratchet trees have precisely
```
josephlhall: potentially dumb question, apologies: is it still a ratchet tree if no ART?

draft-ietf-mls-protocol.md
```diff
@@ -490,46 +491,14 @@ contains up to three values:
...
 
 ~~~~~
-DH(DH(AB), DH(CD))
-    /      \
- DH(AB)    DH(CD)
- /  \      /  \
-A    B    C    D
+parent_secret = Hash(child_secret)
+parent_private, parent_public = Derive-Key-Pair(parent_secret)
 ~~~~~
```
josephlhall: Don't need the twiddles in the markdown if it's not pre-formatted?

draft-ietf-mls-protocol.md
```diff
@@ -554,43 +523,67 @@ X, then the tree will have the following structure.
...
+ / \     / \
+A   _   C   D
+
+0 1 2 3 4 5 6
+~~~~~
+
+In this tree, we can see all three of the above rules in play:
+
+* The resolution of node 5 is the list [CD]
+* The resolution of node 2 is []
```
josephlhall: "node 2 is [] (the empty list)"

draft-ietf-mls-protocol.md
```diff
@@ -554,43 +523,67 @@ X, then the tree will have the following structure.
...
+* The resolution of node 3 is [A, CD]
```
josephlhall: "node 3 is the list [A, CD]"

draft-ietf-mls-protocol.md
```diff
@@ -1013,7 +1052,6 @@ struct {
   PublicKey tree<1..2^32-1>;
   GroupOperation transcript<0..2^32-1>;
   opaque init_secret<0..255>;
```
raphaelrobert: This leaks the group secret of the current epoch to the new participant.

raphaelrobert: If we go for the solution proposed on the mailing list, this should be the hash of the current group secret and the other members should perform the hash as they process the Add message.

bifurcation: You guys have an off-by-one error :)  This is the `init_secret` at the **output** of the current epoch's key schedule, not the input.  So it can't be used to derive the current epoch secret.

The whole reason that we have an `init_secret` (as opposed to chaining from one `epoch_secret` to the next) is so that you can give it to someone to allow them to initialize without giving away the current epoch secrets.  

@beurdouche not sure what you mean about going backward.  Key schedule is all HDKFing along, so it's not reversible.

bifurcation: In any case, I'll add some notes to clarify.

draft-ietf-mls-protocol.md
```diff
@@ -533,21 +568,22 @@ direct path from a leaf to the root. Other participants in the group
...
-To perform an update, the sender transmits a node by sending the public key
-for the node and an encrypted version of the secret value for the
-node.  The secret value is encrypted in such a way that it can be
-decrypted only by holders of the private key for one of its
-children, namely the child that is not in the direct path being
-transmitted.  That is, each node in the direct path is encrypted
-for holders of the private key for a node in the corresponding
-copath.  For leaf nodes, no encrypted secret is transmitted.
+To perform an update, the sender transmits a node by sending the
+public key for the node and one or more encrypted copies of the
```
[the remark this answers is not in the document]
bifurcation: Suppose you've got the following structure:

```
   D2
  /  \
 /    \
D1    Sib
```

... where `D1` and `D2` are on the direct path, and `Sib` is not.  Then the encrypted value that's transmitted is `(pk(D2), Enc(pk(Sib); secret(D2)))`.  

I'll revise to try to make this clearer.

draft-ietf-mls-protocol.md
beurdouche: Well I understand how you read it but the diagram is ill-numbered then, because the `init_secret` is actually at the *input* of the the key schedule. The very first one is the `init_secret[0] = {0}` and is actually needed.

beurdouche: I don't think you understand us regarding the Group operation transcript. If it is not a hash, it contains all group operation messages since the beginning of the group. You can trivially process all these and compute all keys for all previous application messages hence trivially defeating all secrecy.  It also works for future messages as long as the participant you compromise doesn't update. This needs to be a hash, I don't understand why you would think this is safe. @raphaelrobert ?

bifurcation: Oh, I think it is safe because in order to process the GroupOperation messages, you need to have one of the corresponding private keys, which means you need to be in the group!

I'm happy to convert the transcript to a hash (there's no reason for anyone to have the whole thing, unlike the roster and the tree).  But I don't think it's necessary for this to be secure.

bifurcation: cf #73

PR #72, opened 2018-10-19
Add explicit key confirmation (alternate approach)

bifurcation:
This PR is an alternative approach to #71.  Instead of using an explicit MAC, we derive a confirmation value in the key schedule and just publish that.  Since the key schedule already folds in the GroupState (and thus the transcript of GroupOperations), the only gap here is that the signature on the Handshake message isn't included.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -140,6 +140,8 @@ draft-02
 
 - Removed ART (\*)
 
+- Added explicit key confirmation (\*)
+
 draft-01
 
 - Initial description of the Message Protection mechanism. (\*)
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -853,14 +855,16 @@ update_secret -> HKDF-Extract = epoch_secret
                      +--> Derive-Secret(., "app", GroupState_[n])
                      |    = application_secret
                      |
+                     +--> Derive-Secret(., "confirm", GroupState_[n])
+                     |    = key_confirmation
```
beurdouche: Can't we send the next `init_secret`instead ?

bifurcation: That seems bad, since the `init_secret` feeds into the next `epoch_secret`.  It does get mixed with the `update_secret` there, but it would still would make me even more worried than this proposal does.

beurdouche: Indeed it is preferable to avoid it. I was just wondering if we can reuse something we already have...

bifurcation:
Closing in light of WG discussion preferring the other approach.

PR #78, opened 2018-12-04
Change order of fields in UserInitKey struct

raphaelrobert:
algorithm is now before identity_key for easier parsing


bifurcation:
Agreed.  I also ran into this problem in my implementation.

PR #79, opened 2018-12-04
Add ciphersuites and signature schemes

bifurcation:
Need some actual values here for interop.  Signature scheme values are a subset of those in RFC 8446.

@raphaelrobert - Do you think we should go ahead and define the higher-security-level ciphersuites as well as signature algorithms?


raphaelrobert:
Looks good to me! We can always re-introduce Ed448 with X448 later.

PR #81, opened 2019-01-04
Bugfixes

bifurcation:
This PR fixes several small issues I encountered while bringing my implementation more up to date. 

* When I changed `GroupState` to allow blanks in the tree, I failed to update `Welcome` accordingly
* Credentials should dictate what signature scheme they will be used with, and once that's the case, we no longer need to indicate what algorithm is in use in cases where the signer's credential is known
* `UserInitKeys` should come with a credential, not just a public key
* The confirmation value is easier to parse as an opaque vector value than as an array whose size depends on the ciphersuite in use

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -729,8 +729,12 @@ Encryption keys are derived from shared secrets by taking the first
 A member of a group authenticates the identities of other
 participants by means of credentials issued by some authentication
 system, e.g., a PKI.  Each type of credential MUST express the
-holder's identity as well as the public key of a signature key pair
-that the holder of the credential will use to sign MLS messages.
+following data:
+
+* The public key of a signature key pair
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -748,6 +752,7 @@ enum {
 
 struct {
     opaque identity<0..2^16-1>;
+    SignatureScheme algorithm;
     SignaturePublicKey public_key;
 } BasicCredential;
```
beurdouche: Should that just be `Credential` ? Otherwise `BasicCredential` should probably be used in "UserInitKey".

bifurcation: `Credential` is the generic object (with a type selector).  `BasicCredential`is a specific type of credential that just lists the required info (without, say, a signature from a CA).

PR #82, opened 2019-01-04
Simplify authentication calculations

bifurcation:
Looking at the signature and confirmation calculations in the current draft and comparing them to TLS, things seemed a bit complicated.  In TLS, we have the following:

> CertificateVerify: A signature over the value Transcript-Hash(Handshake Context, Certificate)
> Finished: A MAC over the value Transcript-Hash(Handshake Context, Certificate, CertificateVerify) using a MAC key derived from the base key.

This PR implements an analogous computation for MLS, where the signature in a message covers the transcript hash in the latest group state (which includes any changes to the roster ~ Certificate), and the confirmation MAC covers the transcript hash, plus the signature (i.e., the contents of CertificateVerify).


draft-ietf-mls-protocol.md
```diff
@@ -1062,17 +1062,17 @@ follows:
...
+provisional GroupState object:
 
 ~~~~~
-struct {
-  GroupState state;     // Provisional group state
-  Handshake handshake;  // Handshake message, confirmation = 0
-} ConfirmationData;
+signature_data = GroupState.transcript_hash
+
+confirmation_data = GroupState.transcript_hash ||
```
beurdouche: Should that be `confirmation_data = signature_data || Handshake.signature` ?
(otherwise `signature_data` seems unused)

bifurcation: My intent was that "signature_data" would be the input to the signature calculation.  I'll revise to clarify that.

bifurcation:
Merging this despite CI failures because they seem to be spurious.

PR #84, opened 2019-01-11
Garbage collection

### LINKED ISSUE #86: Add-in-Place / Garbage Collection
Currently, the tree only ever grows:

* When you add someone, you have to add them at the right edge, even if there are blank leaves
* When you remove someone, the tree never shrinks

We should enable reclaiming of blank leaves and automatically reduce the size of the tree to the minimum necessary size.

bifurcation:
Right now, we blank out leaf nodes when we remove their holders, but we never reduce the size of hte tree or repopulate interior blanks with new members.  This PR adds those functions:

* Reduce the size of the tree and the roster when you remove from the right edge
* Add an `index` field to the `Add` message, to enable new members to be added to blank slots in the tree, not just at the right edge.

In addition to performing "garbage collection", this change should make it easier to do a state resync -- just remove the desync'ed endpoint and add them back in the same spot.

Fixes #86

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1269,6 +1269,8 @@ then updates its state as follows:
   the null optional value
 * Update the ratchet tree by replacing nodes in the direct
   path from the removed leaf using the information in the Remove message
+* Reduce the size of the roster and the tree until the rightmost
+  element roster element and leaf node are non-null
 * Update the ratchet tree by setting to blank all nodes in the
   direct path from the removed leaf to the root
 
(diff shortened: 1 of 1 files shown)


bifurcation:
@beurdouche Do you think this is ready to merge?  [Notes from the interim](https://github.com/mlswg/mls-protocol/issues/86#issuecomment-454579285) say it's OK as long as we can convince ourselves that truncation preserves the tree invariant. 

Proof sketch: The tree invariant is a universal statement over nodes in the tree and members in the group:  

image

Since subtrees are just subsets of the set of nodes in the tree, if the tree invariant holds for a given tree, then it also holds for any subtree of that tree.  So since truncation just moves from a tree to a subtree, the invariant continues to hold.

PR #106, opened 2019-01-18
Small edits: formatting, typos, and wording

rozbb:
I focused on 3 main things in this edit, which are reflected in the 3 commits:

1. Looking at the HTML-rendered output, there were some structs that were already backtick-literal and some that were not. For the sake of consistency and readability, I literalized all structs and variable names that occur in prose. I also chose to leave function names alone, since it felt appropriate that they be written in the font of the rest of the paragraph.
2. There were 3 obvious typos. The first one was "it its" where it meant "its", the second one was "participants" where it meant "participant's" (I'm sorry, this edit was actually in the formatting commit but it should've been in the typo commit), and the third was was listing "P-256" as the hashing algorithm for one of the ciphersuites.
3. There was a lot of wording like "direct path from the leaf to the root node" but the wording set up in section 5.1 is "direct path of the node", so I changed all instances of the former to the latter for consistency's sake. There were also some uses of "non-updated sibling" where the term "copath" could be used. I'm least confident about this change, since it sounds a bit obtuse, but it's nice to stick to the terminology that has been used previously.

Of course, I'm open to any and all disagreements. To whomever reads this: thank you for your time.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -848,7 +848,7 @@ struct {
 struct {
     DHPublicKey public_key;
     ECIESCiphertext node_secrets<0..2^16-1>;
-} RatchetNode
+} RatchetNode;
 
 struct {
     RatchetNode nodes<0..2^16-1>;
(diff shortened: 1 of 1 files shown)


beurdouche:
Thank you for the review, we'll look into it ASAP.

beurdouche:
This is done, I cherry picked 3 out of 4 commits, made minor corrections and merged the changes via (PR #111) opened an issue for the consistency of using backticks that I will discuss with other editors later (Issue #110). Thanks you, this was helpful !

Issue #95, opened 2019-01-14
Decouple curves from symmetric+hash identifiers


bifurcation:
* Enable, e.g., P-256 with both AES-GCM and ChaChaPoly
* Re-use code points from TLS

beurdouche:
We should discuss that today

PR #113, opened 2019-01-22
Algorithm agility and ciphersuites (Issue #95)

beurdouche:
DO NOT MERGE: Work in Progress

Discussion: Is this how we should define a ciphersuite -> (KEM,SYM,SIG,HASH) ?

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -689,10 +689,6 @@ enum {
     secp256r1(0x0017), secp384r1(0x0018), secp521r1(0x0019),
     x25519(0x001D), x448(0x001E),
 
-    /* Finite Field Groups (DHE) */
-    ffdhe2048(0x0100), ffdhe3072(0x0101), ffdhe4096(0x0102),
-    ffdhe6144(0x0103), ffdhe8192(0x0104),
-
     /* Reserved */
     (0xFFFF)
 } NamedGroup;
@@ -706,10 +702,6 @@ Elliptic Curve Groups (ECDHE)
 : Indicates support for the corresponding named curve, defined
   either in FIPS 186-4 {{DSS}} or in {{RFC7748}}.
 
-Finite Field Groups (FFDHE)
-: Indicates support of the corresponding finite field
-  group, defined in {{RFC7919}}.
-
 ~~~~~
 enum {
     /* ECDSA algorithms */
@@ -720,11 +712,6 @@ enum {
     /* EdDSA algorithms */
     ed25519(0x0807),
     ed448(0x0808),
-
-    /* RSASSA-PSS algorithms with public key OID RSASSA-PSS */
-    rsa_pss_pss_sha256(0x0809),
-    rsa_pss_pss_sha384(0x080a),
-    rsa_pss_pss_sha512(0x080b),
 
     /* Reserved */
     (0xFFFF)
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -662,17 +662,113 @@ Cryptographic algorithms are indicated using the following types:
 
 ~~~~~
 enum {
+    /* Elliptic Curve Groups (ECDHE) */
```
bifurcation: If we're going to decouple like this, then we should just incorporate the TLS algorithm identifiers by reference.

draft-ietf-mls-protocol.md
```diff
@@ -662,17 +662,113 @@ Cryptographic algorithms are indicated using the following types:
...
+  curve as defined in ANSI X9.62 {{X962}} and FIPS 186-4 {{DSS}}, and the
+  corresponding hash algorithm as defined in {{!SHS}}. The signature is
+  represented as a DER-encoded {{X690}} ECDSA-Sig-Value structure.
+
+EdDSA algorithms
+: Indicates a signature algorithm using EdDSA as defined in
+  {{RFC8032}} or its successors. Note that these correspond to the
+  "PureEdDSA" algorithms and not the "prehash" variants.
+
+RSASSA-PSS PSS algorithms
```
bifurcation: Let's not do RSA-PSS.  I don't think anyone is asking for it.

bifurcation:
In addition to the two comments inline, note that there's an additional complication here due to `Derive-Key-Pair` -- if we do this decoupling, then we need to also have a requirement that the hash function in use produces enough entropy for the curve in use.  E.g., you wouldn't want to use Ed448 with SHA-256.

katrielalex:
I am trying to fix up the CI and it looks like this PR is actually broken:

```
4:13:00 PM: Error: Unable to validate the XML document: draft-ietf-mls-protocol.xml
4:13:00 PM:  <string>: Line 726: IDREF attribute target references an unknown ID "SHS"
4:13:00 PM:  <string>: Line 736: IDREF attribute target references an unknown ID "SHS"
4:13:00 PM:  <string>: Line 788: IDREF attribute target references an unknown ID "SHS"
4:13:00 PM:  <string>: Line 730: IDREF attribute target references an unknown ID "RFC8032"
4:13:00 PM:  <string>: Line 694: IDREF attribute target references an unknown ID "RFC7919"
4:13:00 PM:  <string>: Line 733: IDREF attribute target references an unknown ID "RFC8017"
4:13:00 PM:  <string>: Line 724: IDREF attribute target references an unknown ID "ECDSA"
4:13:00 PM:  <string>: Line 691: IDREF attribute target references an unknown ID "DSS"
4:13:00 PM:  <string>: Line 725: IDREF attribute target references an unknown ID "DSS"
4:13:00 PM:  <string>: Line 727: IDREF attribute target references an unknown ID "X690"
4:13:00 PM:  <string>: Line 788: IDREF attribute target references an unknown ID "RFC7539"
4:13:00 PM: make: *** [draft-ietf-mls-protocol.htmltmp] Error 1
```

draft-ietf-mls-protocol.md
beurdouche: I actually agree with this and will remove it. I liked the idea of aligning on TLS but randomness issues related to this are not something we want.

beurdouche:
@bifurcation I'll try to work a bit more on this tomorrow. Anything controversial or missing ?

beurdouche:
Specifically, the next two subsections just before "Credentials" should probably be removed or merged.

bifurcation:
I think this one is well stale by now.  Still an issue that merits consideration, but needs a fresh PR.

PR #114, opened 2019-02-08
Use a KDF instead of a hash function to get keys for parent nodes

kkohbrok:
For now, I've only touched the sections until the "Cryptographic Objects" section, which seems to be outdated anyway. Once those are updated at some point, I'm happy to help moving from hashing to KDF-derivation there as well.


beurdouche:
Thanks Konrad ! : )

Issue #107, opened 2019-01-19
Application key schedule is busted


bifurcation:
There are at least two bugs in the message protection section that make it unimplementable:

1. It calls for `Derive-Secret(., "app sender", [sender])`, but the third argument to `Derive-Secret` as defined is a `GroupState`, not an octet string.

2. It refers to a function `HKDF-Expand-Label`, which is defined in TLS, but not here.

ISTM the simplest way to fix these problems would to define `HKDF-Expand-Label` and use it (1) as the basis for `Derive-Secret`, (2) for deriving sender root secrets, and (3) for deriving keys and nonces.

beurdouche:
Fixed by #123

Issue #90, opened 2019-01-13
Simplify the key schedule


bifurcation:
Currently, the whole GroupState object is folded into the key schedule, via the `DeriveSecret` function.  Could this be simplified?  For example:

* Hash the GroupState object once, and use it for all `DeriveSecret` calls, as opposed to passing it by value each time

* Hash in the roster and tree as a commitment (cf #87) rather than by value

Other suggestions welcome.

bifurcation:
I think we can get away with using the transcript hash if we put a little more information in the transcript.  In particular, the Welcome message needs to be included to assure that the new joiner and the group have the same view of the group state bits in the Welcome message.

PR #115, opened 2019-02-08
Key Schedule uses the transcript_hash instead of the raw group_state

beurdouche:
This is an attempt to solve #90 and #107.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1472,8 +1472,8 @@ to be used for its own sending chain:
            application_secret_[sender]_[0]
 ~~~
 
-Note that [sender] represent the uint32 bytes for the big-endian encoding
-the index of the participant in the ratchet tree.
+Note that [sender] represent the big-endian encoding of the uint32 word
+used to store the index of the participant in the ratchet tree.
 
 Updating the Application secret and deriving the associated AEAD key and nonce can
 be summarized as the following Application key schedule where
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -921,7 +921,7 @@ Where HkdfLabel is specified as:
 
 struct {
     uint16 length = Length;
-    opaque label<7..255> = "mls10 " + Label;
+    opaque label<6..255> = "mls10 " + Label;
     opaque context<0..255> = Context;
 } HkdfLabel;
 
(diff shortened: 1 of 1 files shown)


beurdouche:
Note that there is a need to handle messages that are not exactly Group Operations such as the Welcome message. (I think it should be considered as a Group Operation and be added in the Group State transcript hash)

draft-ietf-mls-protocol.md
```diff
@@ -944,23 +947,32 @@ Given these inputs, the derivation of secrets for an epoch
...
-                     |
-                     V
-               init_secret_[n]
-~~~~~
+                  init_secret_[N-1] (or 0)
+                        |
+                        V
+update_secret_[N-1] --> HKDF-Extract = epoch_secret_[N]
+                        |
+                        +--> Derive-Secret(., "app",
```
bifurcation: If you're going to change `Derive-Secret` to `HKDF-Expand-Label`, then you need to chance these instances as well.  Alternatively, you could just define `Derive-Secret` in terms of `HKDF-Expand-Label`, basically as an abbreviation to make this diagram flow better.

draft-ietf-mls-protocol.md
```diff
@@ -944,23 +947,32 @@ Given these inputs, the derivation of secrets for an epoch
...
+                        |                  GroupState_[N].transcript_hash)
```
bifurcation: I would prefer that we not make this change, i.e., that we set the context to the serialized group state.  Changing from the group state to the transcript hash has bigger implications than just changing how the hashes are arranged.  That is, let's do #107 in this PR, and #90 later.  Note that this would also obviate the OPEN ISSUE below.

draft-ietf-mls-protocol.md
```diff
@@ -1495,6 +1507,7 @@ chained from the initial one to the next generation of Epoch secret. While this
 seems safe because cryptographic operations using the application secrets can't
 affect the group init_secret, it remains to be proven correct. ]]
 
+
```
bifurcation: Nit: Extra blank line.

draft-ietf-mls-protocol.md
```diff
@@ -1454,14 +1466,14 @@ to be used for its own sending chain:
...
+           HKDF-Expand-Label(., "app sender", [sender], Hash.length)
                      |
                      V
            application_secret_[sender]_[0]
 ~~~
 
-Note that [sender] represent the uint32 value encoding the index
-of the participant in the ratchet tree.
+Note that [sender] represent the big-endian encoding of the uint32 word
+used to store the index of the participant in the ratchet tree.
```
bifurcation: You could note here that this is the same as the TLS encoding of `sender` as a `uint32`.  Which is probably in fact how I would implement it.

beurdouche:
Closing this in favor of #123 (which does not touch the way we pass the Group State to Derive Secret at all, we can do this in a different PR if need be.)

PR #116, opened 2019-02-08
Rename ECIES to HPKE

beurdouche:
This does not handle the removal of DH uses instead of KEM which will be done as part of Issue #95 / PR #113 ...

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -38,6 +38,7 @@ author:
     email: raphael@wire.com
 
 
+
 normative:
   X962:
        title: "Public Key Cryptography For The Financial Services Industry: The Elliptic Curve Digital Signature Algorithm (ECDSA)"
@@ -47,6 +48,7 @@ normative:
        seriesinfo:
          ANSI: X9.62
   IEEE1363: DOI.10.1109/IEEESTD.2009.4773330
+
 
 
 informative:
@@ -70,6 +72,20 @@ informative:
     author:
        - name: Trevor Perrin(ed)
        - name: Moxie Marlinspike
+
+  HPKE:
+       title: "Hybrid Public Key Encryption"
+       date: 2019
+       author:
+         -  ins: R. Barnes
+            name: Richard Barnes
+            organization: Cisco
+            email: rlb@ipv.sx
+         -
+            ins: K. Bhargavan
+            name: Karthik Bhargavan
+            organization: Inria
+            email: karthikeyan.bhargavan@inria.fr
 
 
 --- abstract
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -852,13 +856,13 @@ each node MUST be the parent of its predecessor.
 
 ~~~~~
 struct {
-    DHPublicKey ephemeral_key;
+    PublicKey ephemeral_key;
```
bifurcation: Since we're moving to a more KEM-like framework, it might be good to also change this from `ephemeral_key` to something like `encapsulated_key`.

Also, while you're working here, please go ahead and increase the size of the `ciphertext` field.

bifurcation: Also, the reason I had `DH` there was to distinguish from `SignaturePublicKey`.  Just having `PubilcKey` vs. `SignaturePublicKey` seems less clear.  What about `KEMPublicKey` or `HPKEPublicKey`?

draft-ietf-mls-protocol.md
```diff
@@ -886,20 +890,20 @@ encrypted secret values are computed as follows:
...
+nonce = HKDF-Expand(Secret, HPKELabel("nonce"), Length)
 
-Where ECIESLabel is specified as:
+Where HPKELabel is specified as:
 
 struct {
   uint16 length = Length;
   opaque label<12..255> = "mls10 ecies " + Label;
-} ECIESLabel;
+} HPKELabel;
```
bifurcation: I'm inclined to go ahead and change this out for a reference to the CFRG HPKE draft.  If that doesn't get adopted by CFRG for some reason, we can always revert.

draft-ietf-mls-protocol.md
```diff
@@ -1167,7 +1171,7 @@ object.
 
 [[ OPEN ISSUE: The Welcome message needs to be sent encrypted for
 the new member.  This should be done using the public key in the
-UserInitKey, either with ECIES or X3DH. ]]
+UserInitKey, either with HPKE or X3DH. ]]
```
bifurcation: This OPEN ISSUE can be deleted.  The Welcome Message is encrypted as of -03; I just neglected to remove this comment.

bifurcation:
I would be happier if this PR were moving the draft to use draft-barnes-cfrg-hpke, which is one of the changes I had in mind for -04.  Do you think that's premature?

draft-ietf-mls-protocol.md
beurdouche: Ok, I removed the description and added an informative reference to HPKE.

draft-ietf-mls-protocol.md
beurdouche: I removed this as we can point to the HPKE draft to get how the HPKECiphertext is internally organized.
There is still a sentence saying what the argument of HPKE Encrypt (node_secrets) are.

PR #120, opened 2019-02-17
WIP: Introduce a common framing layer

### LINKED ISSUE #101: Use common framing
We have several types of top-level objects right now:

* UserInitKey messages
* Welcome messages
* Handshake messages
* Application messages

We should have a common framing layer that describes the object type (and optionally provides encryption), to prevent confusion among these types.  This would be analogous to the [TLS record layer](https://tlswg.github.io/tls13-spec/draft-ietf-tls-tls13.html#rfc.section.5.1)

bifurcation:
This PR introduces a common framing layer that is used to encapsulate both Handshake and Application messages, encrypted or unencrypted.  The most significant impact is that in order to make signing logic common to both of these cases, the signature now covers only the message in question, as opposed to also covering the transcript.    

Fixes #101

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -963,7 +963,60 @@ update_secret -> HKDF-Extract = epoch_secret
 
 ## Application Key Schedule
 
-[[ TODO - Move most of the application section up here ]]
+The keys used to encrypt MLS messages during an epoch are derived
+from the epoch's application secret.  Each sender derives a chain of
+per-sender application secrets.  Each per-sender application secret
+is in turn used to generate a (key, nonce) pair that is used to
+encrypt a single message.
+
+Each step in the sequence of application secrets used by a sender is
+called a "generation".  The application secret for the first
+generation is derived from the epoch's application secret.
+
+~~~~~
+           application_secret
+                     |
+                     V
+           Derive-Secret(., "app sender", [sender])
+                     |
+                     V
+           application_secret_[sender]_[0]
+~~~~~
+
+Here the value [sender] represents the index of the sender in the
+ratchet tree, encoded as a uint32.  Each generation of the
+application secret is used to derive three values: A key and nonce
+to be used for encrypting a message, and the next generation of the
+sender's application secret.
+
+~~~~~
+           application_secret_[sender]_[N-1]
+                     |
+                     +--> HKDF-Expand-Label(.,"nonce", "", nonce_length)
+                     |    = write_nonce_[sender]_[N-1]
+                     |
+                     +--> HKDF-Expand-Label(.,"key", "", key_length)
+                     |    = write_key_[sender]_[N-1]
+                     V
+           Derive-Secret(., "app upd","")
+                     |
+                     V
+           application_secret_[sender]_[N]
+
+~~~~~
+
+Thus, the key and nonce used to encrypt a message are uniquely
+identified by a four-tuple of values (group_id, epoch, sender,
+generation).
+
+[[OPEN ISSUE: The HKDF context field is left empty for now.
+A proper security study is needed to make sure that we do not need
+more information in the context to achieve the security goals.]]
+
+[[ OPEN ISSUE: At the moment there is no contributivity of Application secrets
+chained from the initial one to the next generation of Epoch secret. While this
+seems safe because cryptographic operations using the application secrets can't
+affect the group init_secret, it remains to be proven correct. ]]
 
 ## Message Framing and Encryption
 
@@ -972,7 +1025,8 @@ structure.  (UserInitKey and Welcome messages are sent outside of
 the group, and have their own structure.)  The two main structures
 involved in this framing system are MLSPlaintext, representing a
 plaintext message, and MLSCiphertext, representing a message
-encrypted using an application key (see [[ TODO ]]).
+encrypted using an application key (see
+{{application-key-schedule}}).
 
 Applications may decide to transmit handshake messages either in
 plaintext or ciphertext (encrypted with an appropriate sender key).
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -143,6 +143,9 @@ draft-04
 
 - Using a KDF instead of a Hash in TreeKEM (\*)
 
+- Use a common framing / signing / encryption for both Handshake and
+  Application messages (\*)
+
 draft-03
 
 - Added ciphersuites and signature schemes (\*)
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1000,6 +1000,70 @@ struct {
 } UserInitKey;
 ~~~~~
 
+# Message Framing
+
+[[ TODO prose ]]
+
+~~~~~
+enum {
+    invalid(0),
+    handshake(1),
+    application(2),
+    (255)
+} ContentType;
+
+struct {
+    opaque content[MLSPlaintext.length];
+    uint8 signature[MLSInnerPlaintext.sig_len];
+    uint16 sig_len;
+    ContentType type;
+    uint8 zero\_padding[length\_of\_padding];
+} MLSInnerPlaintext;
+
+struct {
+    opaque group_id<0..255>;
+    uint32 epoch;
+    uint32 sender;
+    ContentType type;
+
+    select (MLSPlaintext.type) {
+        case handshake:   Handshake;
+        case application: Application;
+    }
+
+    opaque signature<1..2^16-1>;
+} MLSPlaintext;
+
+struct {
+    opaque group_id<0..255>;
+    uint32 epoch;
+    uint32 sender;
+    uint32 generation;
+    opaque cipertext<0..2^32-1>;
+} MLSCiphertext;
+~~~~~
+
+[[
+
+* Signature on MLSPlaintext is over all of the above
+* Encryption: MLSPlaintext -> MLSCiphertext
+  * Marshal MLSInnerPlaintext
+  * Encrypt -> ciphertext
+  * Form MLSCiphertext
+* Decryption: MLSCiphertext -> MLSPlaintext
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1000,6 +1000,70 @@ struct {
...
+    handshake(1),
+    application(2),
+    (255)
+} ContentType;
+
+struct {
+    opaque content[MLSPlaintext.length];
+    uint8 signature[MLSInnerPlaintext.sig_len];
+    uint16 sig_len;
+    ContentType type;
```
beurdouche: The signature must cover everything including the content type

draft-ietf-mls-protocol.md
```diff
@@ -1000,6 +1000,70 @@ struct {
...
+enum {
+    invalid(0),
+    handshake(1),
+    application(2),
+    (255)
+} ContentType;
+
+struct {
+    opaque content[MLSPlaintext.length];
+    uint8 signature[MLSInnerPlaintext.sig_len];
```
beurdouche: `sig_len` is undefined.

draft-ietf-mls-protocol.md
```diff
@@ -1000,6 +1000,70 @@ struct {
...
+    uint16 sig_len;
```
beurdouche: If we are gonna do this, I would prefer `opaque signature<0..2^16-1>`
otherwise if the ciphertext gives the length of the signature we don't need this.

draft-ietf-mls-protocol.md
```diff
@@ -1000,6 +1000,70 @@ struct {
...
+    uint32 epoch;
+    uint32 sender;
+    ContentType type;
+
+    select (MLSPlaintext.type) {
+        case handshake:   Handshake;
+        case application: Application;
+    }
+
+    opaque signature<1..2^16-1>;
```
beurdouche: I am confused now, why is there a second signature here ?
In both cases there is a signature that covers the content of `group_id`, `epoch`, `sender` and `content type` no ?

draft-ietf-mls-protocol.md
```diff
@@ -1000,6 +1000,70 @@ struct {
...
+} MLSPlaintext;
+
+struct {
+    opaque group_id<0..255>;
+    uint32 epoch;
+    uint32 sender;
+    uint32 generation;
```
beurdouche: While this is weakly-authenticated via the AEAD it should be in the MLSPlaintext header as well and covered by the signature (especially in the case of an Application message) to avoid forgery from an adversarial member who is not the sender.

draft-ietf-mls-protocol.md
```diff
@@ -1410,7 +1445,10 @@ all arrive at the following state:
  A    X    Y    D
 ~~~~~
 
-# Message Protection
+# Application Messages
+
+[[ XXX(RLB): It should be possible to simplify this section in light
```
beurdouche: I can do that when this PR is merged.

draft-ietf-mls-protocol.md
beurdouche: It is also worth noting that this field even if not technically needed in the case of a Handshake message might be able to use this to advertise a counter for application messages previously sent by the sender, hence allowing recipients to know if they received all app messages from the sender. We could limit that to the previous epoch or not... Maybe there is something there or not...

raphaelrobert:
As a general rule, I think we should not prefix structs with "MLS" as it adds no semantic value

draft-ietf-mls-protocol.md
```diff
@@ -1000,6 +1000,70 @@ struct {
...
+
+[[ TODO prose ]]
+
+~~~~~
+enum {
+    invalid(0),
+    handshake(1),
+    application(2),
+    (255)
+} ContentType;
```
kkohbrok: Why would a content ever be labelled invalid?

draft-ietf-mls-protocol.md
```diff
@@ -1000,6 +1000,70 @@ struct {
...
+    opaque signature<1..2^16-1>;
+} MLSPlaintext;
+
+struct {
+    opaque group_id<0..255>;
+    uint32 epoch;
+    uint32 sender;
+    uint32 generation;
+    opaque cipertext<0..2^32-1>;
+} MLSCiphertext;
```
kkohbrok: What do we need the sender field for in MLSCiphertext? We might even get some sort of identity hiding if we leave it out :-)

draft-ietf-mls-protocol.md
kkohbrok: I think this is the only signature we need. Then we can leave out the ContentType and the signature in the inner plaintext.

draft-ietf-mls-protocol.md
kkohbrok: What is the generation field for exactly?

draft-ietf-mls-protocol.md
beurdouche: On the contrary, the inner one is the one we use, you don't want to sign a large padding.
What is important is that the inner one must cover those fields as done the message protection section.

draft-ietf-mls-protocol.md
```diff
@@ -1000,6 +1000,70 @@ struct {
...
+    uint32 sender;
+    ContentType type;
+
+    select (MLSPlaintext.type) {
+        case handshake:   Handshake;
+        case application: Application;
+    }
+
+    opaque signature<1..2^16-1>;
+} MLSPlaintext;
```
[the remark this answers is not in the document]
beurdouche: We do not need to actually copy them twice indeed, only the outer one in `Ciphertext` is needed, but they need to be covered by both the AEAD and the internal signature.

draft-ietf-mls-protocol.md
```diff
@@ -1000,6 +1000,70 @@ struct {
...
+
+struct {
+    opaque group_id<0..255>;
+    uint32 epoch;
+    uint32 sender;
```
beurdouche: `sender` and `generation` should be inside `Plaintext` and not `Ciphertext`

draft-ietf-mls-protocol.md
```diff
@@ -1000,6 +1000,70 @@ struct {
...
+    opaque content[MLSPlaintext.length];
+    uint8 signature[MLSInnerPlaintext.sig_len];
+    uint16 sig_len;
+    ContentType type;
+    uint8 zero\_padding[length\_of\_padding];
+} MLSInnerPlaintext;
+
+struct {
+    opaque group_id<0..255>;
+    uint32 epoch;
```
beurdouche: `group_id` and `epoch` should be in `Ciphertext` and not in `Plaintext`

draft-ietf-mls-protocol.md
beurdouche: No, indeed we don't want it outside, that is also why the signature is inside the Plaintext.

draft-ietf-mls-protocol.md
kkohbrok: Ok, then we're assuming that `group_id`, `sender` and `epoch` are still part of the `Handshake` struct? I was wondering if we could somehow avoid redundancy here.

draft-ietf-mls-protocol.md
```diff
@@ -1000,6 +1000,70 @@ struct {
...
+    uint32 sender;
+    ContentType type;
```
beurdouche: My initial interpretation was incorrect but, if I am correct now, this should be removed.

draft-ietf-mls-protocol.md
beurdouche: I was misunderstanding Richard's idea at first. I believe the idea is to encrypt the `Plaintext` with different keys depending if it is `Handshake` or `Application`. In that scenario, you need the `sender` and the `generation` in the clear to know which key to pick in the case of an Application message. Note that @raphaelrobert and I are thinking on suggesting to encrypt these under the the group key to hide them if necessary, we'll discuss this at some point.

kkohbrok: Why do you need the sender, though? Both handshake and application keys derived from the groupkey should be sender-independent.

beurdouche: Which key do you pick when receiving an application message ? :)

kkohbrok: I'm confused. The key for handshake encryption is derived from the `group_key`, right? So why do we need to know the sender? We only need the `group_id` and the `epoch` to derive the key if I understand correctly.

draft-ietf-mls-protocol.md
bifurcation: The idea is that the ContentType is used to mark where the padding bytes (all zero) end, so it needs to be non-zero.

bifurcation:
I added some prose in the latest round of commits that should clarify what's going on here.  I think that should address most of the comments. 

@raphaelrobert - regarding the "MLS" prefix, I'm inclined to keep it, because (1) it mirrors TLS, and (2) just having Plaintext and Ciphertext seems odd to me.

draft-ietf-mls-protocol.md
```diff
@@ -948,6 +948,212 @@ update_secret -> HKDF-Extract = epoch_secret
...
+  type      |
+  content   |--+                     +--> ciphertext
+  signature |  |  MLSInnerPlaintext  |
+               |  | content       |  |
+               +--| signature     |--+
+                  | sig_len       |
+                  | type          |
+                                  |
+                  * zero_padding  |
+~~~~~
```
kkohbrok: Do I understand correctly that the fields that represent information that is shared between the different structs are not actually part of the encoded data structure but rather are just meant to be pointers to the respective "outer" struct? If that is the case, I think we should stay true to the principle of having the structs represent what is actually part of the data structure that is encoded and not what is composed locally to compute signatures. Please do correct me if I'm wrong in my understanding of the spec here.

bifurcation:
Closing in favor of #131

PR #123, opened 2019-02-21
Fix for Application Key Schedule and define HKDF-Expand-Label

beurdouche:
Fix for #107


beurdouche:
@bifurcation This might not be completely ideal, because of the group state but it currently does not change the key schedule, I am merging this because we need it for correct interop but feel free to get back to discuss a better way to do this in details.

PR #131, opened 2019-02-27
Legacy - Common Framing serving as a base for PR  #153 and #155

beurdouche:
Alternative to #120 to solve #101

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -935,6 +935,9 @@ proceeds as shown in the following diagram:
                      V
 update_secret -> HKDF-Extract = epoch_secret
                      |
+                     +--> Derive-Secret(., "handshake", GroupState_[n])
+                     |    = handshake_encryption_key
+                     |
                      +--> Derive-Secret(., "app", GroupState_[n])
                      |    = application_secret
                      |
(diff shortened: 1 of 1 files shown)


beurdouche:
Current todo before we can merge this: Proverif/Tamarin model. The intuition is that this is fine because the signature covers all the metadata, to be safe, the second AAD could take the ciphertext of the sender data + tag...

draft-ietf-mls-protocol.md
```diff
@@ -1003,69 +1012,119 @@ enum {
...
-} Plaintext;
+} MLSPlaintext;
 
 struct {
     opaque group_id<0..255>;
     uint32 epoch;
-    uint32 sender;
-    uint32 generation;
+    ContentType type;
+    opaque nonce_sender_data<0..255>;
```
bifurcation: Nit: I would call this `sender_data_nonce`

draft-ietf-mls-protocol.md
```diff
@@ -1003,69 +1012,119 @@ enum {
...
 
 struct {
-    ContentType type;
-    select (Plaintext.type) {
+    select (MLSPlaintext.type) {
         case handshake:   Handshake;
         case application: Application;
     }
-} Plaintext;
+} MLSPlaintext;
```
bifurcation: I would prefer if this struct had everything that was not necessary for encryption.  That way if you're not encrypting Handshake messages, you just send this struct instead.  So I would add back:

* Group ID
* Content type
* Signature

PR #132, opened 2019-02-28
Some propositions for minor changes in structure and wording of the Handshake section.

kkohbrok:
The handshake section felt a bit "flat" to me, so I introduced some more structure. Also, every section describing the handling of the individual message types mentioned verifying the signature, which was already covered in the general description of how to handle Handshake messages. I realize this Section might change somewhat with the generic framing approach, so feel free to ignore this particular PR, if you want to tackle these points there.

Update: I also found what I think is an small error in the instructions of how to process an Add message: The position of the newly added credential in the list of members has to match the position of the newly added leaf node in the tree. This necessitates that it happens _after_ the leaf node is added. The last commit updates the Add instructions accordingly.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1057,7 +1057,8 @@ follows:
    is equal the `epoch` field of the current GroupState object.
 
 2. Use the `operation` message to produce an updated, provisional
-   GroupState object incorporating the proposed changes.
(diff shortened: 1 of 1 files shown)


kkohbrok:
The text has changed quite a bit in the meantime and I don't think my comments are very relevant anymore.

Closing.

PR #134, opened 2019-03-06
Tree Hash

### LINKED ISSUE #90: Simplify the key schedule

bifurcation:
As discussed on the list, this PR replaces the explicit tree and roster inputs to the key schedule with a "tree hash" construct.

@beurdouche - Given this, do you think we need to keep the `transcript_hash` field around?  It seems like in TLS and similar, the transcript hash is there to ensure that the parties agree on things like identities, but we're encoding all that explicitly.

Fixes #90

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -841,7 +841,7 @@ struct {
 ## Tree Hashes
 
 To allow group members to verify that they agree on the
-cryptographic state of the group, this section define a scheme for
+cryptographic state of the group, this section defines a scheme for
 generating a hash value that represents the contents of the group's
 ratchet tree and the members' credentials.
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -581,7 +581,7 @@ in {{tree-hashes}}.
 
 ## Views of a Ratchet Tree {#views}
 
-MLS assumes that each participant maintains a complete and
+We generally assume that each participant maintains a complete and
 up-to-date view of the public state of the group's ratchet tree,
 including the public keys for all nodes and the credentials
 associated with the leaf nodes.
@@ -592,8 +592,8 @@ the nodes.  Instead, each member is assigned to a leaf of the tree,
 which determines the set of secret state known to the member.  The
 credential stored at that leaf is one provided by the member.
 
-In particular, MLS maintains the members view of the tree in such a
-way as to maintain the _tree invariant_:
+In particular, MLS maintains the members' views of the tree in such
+a way as to maintain the _tree invariant_:
 
     The private key for a node in the tree is known to a member of
     the group if and only if that member's leaf is a descendant of
@@ -665,14 +665,15 @@ above:
 
 The members of the group need to keep their views of the tree in
 sync and up to date.  When a client proposes a change to the tree
-(e.g., to add or remove a member), it transmits a set of values for
-intermediate nodes in the direct path of a leaf. Other members of
-the group can use these nodes to update their view of the tree,
-aligning their copy of the tree to the sender's.
+(e.g., to add or remove a member), it transmits a set of public
+values for intermediate nodes in the direct path of a leaf. The
+other members of the group can use these public values to update
+their view of the tree, aligning their copy of the tree to the
+sender's.
 
-To perform an update for a leaf, the sender transmits the following
-information for each node in the direct path of the leaf, as well as
-the root:
+To perform an update for a leaf, the sender broadcasts to the group
+the following information for each node in the direct path of the
+leaf, as well as the root:
 
 * The public key for the node
 * Zero or more encrypted copies of the path secret corresponding to
@@ -714,9 +715,10 @@ values:
 | pk(ns[1])  | E(pk(A), ps[1])                  |
 | pk(ns[0])  |                                  |
 
-Here pk(X) represents the public key corresponding derived from the
-node secret X and E(K, S) to represent public-key encryption to the
-public key K of the path secret S.
+In this table, the value pk(X) represents the public key
+corresponding derived from the node secret X.  The value E(K, S)
+represents the public-key encryption of the path secret S to the
+public key K.
 
 
 # Cryptographic Objects
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -471,10 +471,10 @@ A   B   C   D   E   F   G
 0 1 2 3 4 5 6 7 8 9 0 1 2
 ~~~~~
 
-Each node in the tree is assigned an _index_, starting at zero and
+Each node in the tree is assigned an _node index_, starting at zero and
 running from left to right.  A node is a leaf node if and only if it
-has an even index.  The indices for the nodes in the above tree are
-as follows:
+has an even index.  The node indices for the nodes in the above tree
+are as follows:
 
 * 0 = A
 * 1 = AB
@@ -494,27 +494,101 @@ as follows:
 used for the Merkle trees in the Certificate Transparency protocol
 {{?I-D.ietf-trans-rfc6962-bis}}.)
 
-## Ratchet Tree Nodes and Tree Updates
+The leaves of the tree are indexed separately, using a _leaf index_,
+since the protocol messages only need to refer to leaves in the
+tree.  Like nodes, leaves are numbered left to right.  Note that
+given the above numbering, a node is a leaf node if and only if it
+has an even node index, and a leaf node's leaf index is half its
+node index.  The leaf indices in the above tree are as follows:
 
-Ratchet trees are used for generating shared group secrets. In this
-section, we describe the structure of a ratchet tree.  A particular
-instance of a ratchet tree is based on the following cryptographic
-primitives, defined by the ciphersuite in use:
+* 0 = A
+* 1 = B
+* 2 = C
+* 3 = D
+* 4 = E
+* 5 = F
+* 6 = G
+
+
+## Ratchet Tree Contents
+
+A particular instance of a ratchet tree is based on the following
+cryptographic primitives, defined by the ciphersuite in use:
 
 * A Diffie-Hellman finite-field group or elliptic curve
 * A Key Derivation Function (KDF)
 * A Derive-Key-Pair function that produces an asymmetric keypair
   from a node secret
 
-A ratchet tree is a left-balanced binary tree, in which each node
-contains up to four values:
+Each node in a ratchet tree contains up to four values:
 
 * A secret octet string (optional)
 * An asymmetric private key (optional)
 * An asymmetric public key
-* A credential (optional)
+* A credential (only for leaf nodes)
 
-The contents of the parent are based on the latest-updated child.
+The contents of a given node are set as described below.
+
+A node in the tree may also be _blank_, indicating that no value is
+present at that node.  The _resolution_ of a node is an ordered list
+of non-blank nodes that collectively cover all non-blank descendants
+of the node.  The nodes in a resolution are ordered according to
+their indices.
+
+* The resolution of a non-blank node is a one element list
+  containing the node itself
+* The resolution of a blank leaf node is the empty list
+* The resolution of a blank intermediate node is the result of
+  concatinating the resolution of its left child with the resolution
+  of its right child, in that order
+
+For example, consider the following tree, where the "\_" character
+represents a blank node:
+
+~~~~~
+      _
+    /   \
+   /     \
+  _       CD
+ / \     / \
+A   _   C   D
+
+0 1 2 3 4 5 6
+~~~~~
+
+In this tree, we can see all three of the above rules in play:
+
+* The resolution of node 5 is the list [CD]
+* The resolution of node 2 is the empty list []
+* The resolution of node 3 is the list [A, CD]
+
+
+## Views of a Ratchet Tree
+
+MLS assumes that each participant maintains a complete and
+up-to-date view of the public state of the group's ratchet tree,
+including the public keys for all nodes and the credentials
+associated with the leaf nodes.
+
+No participant in an MLS group has full knowledge of the secret
+state the group, the secret values and private keys associated ot
+the nodes.  Instead, each member is assigned to a leaf of the tree,
+which determines the set of secret state known to the member.  The
+credential stored at that leaf is one provided by the member.
+
+In particular, MLS maintains the members view of the tree in such a
+way as to maintain the _tree invariant_:
+
+    The secret value and private key for a node in the tree are
+    known to a member of the group if and only if that member's leaf
+    is a descendent of the node or equal to it.
+
+In other words, each member holds the secrets for nodes in its
+direct path, and no others.
+
+
+## Ratchet Tree Updates
+
 Nodes in a tree are always updated along the "direct path" from a
 leaf to the root.  The generator of the update chooses a random
 secret value "path_secret[0]", and generates a sequence of "path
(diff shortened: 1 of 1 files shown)


beurdouche:
This is roughly what we do for the TreeKEM paper (I believe we don't re-mix the identities at each op. if I am correct). This seem to cover enough so it's good. I'll try to have a deeper look tomorrow, but this looks reasonable at first glance. It covers both the identities/roster and the group operations, (quid of the welcome message ?) so yes this makes the transcript hash redundant it seems :) I'll get back asap.

draft-ietf-mls-protocol.md
```diff
@@ -786,10 +786,16 @@ struct {
 } Credential;
 ~~~~~
 
-## Group State
+## Commitment to the Group's Cryptographic State {#tree-hash}
 
-Each member of the group maintains a representation of the
-state of the group:
+To allow group members to verify that they agree on the
+cryptographic state of the group, this section define a scheme for
```
rozbb: Typo: "this section define a scheme" should be "this section defines a scheme"

rozbb:
If the goal is to have a new data structure that combined and replaces the ratchet tree and roster, then parent and leaf nodes should be storing a lot more information. Namely, if this to be used as a ratchet tree, all nodes need to have associated public keys and private keys, and parent nodes should retain a hash of their children. No need to store node secrets, though. They're never used after private key derivation.

draft-ietf-mls-protocol.md
```diff
@@ -494,26 +494,101 @@ as follows:
...
+
+## Views of a Ratchet Tree
+
+MLS assumes that each participant maintains a complete and
+up-to-date view of the public state of the group's ratchet tree,
+including the public keys for all nodes and the credentials
+associated with the leaf nodes.
+
+No participant in an MLS group has full knowledge of the secret
+state the group, the secret values and private keys associated ot
```
rozbb: Typo: `ot` -> `to`

rozbb:
I think the definition of Resolution should be in the Tree Computation Terminology section, not the Ratchet Tree Contents section

draft-ietf-mls-protocol.md
```diff
@@ -494,26 +502,105 @@ as follows:
...
+of non-blank nodes that collectively cover all non-blank descendants
+of the node.  The nodes in a resolution are ordered according to
+their indices.
+
+* The resolution of a non-blank node is a one element list
+  containing the node itself
+* The resolution of a blank leaf node is the empty list
+* The resolution of a blank intermediate node is the result of
+  concatinating the resolution of its left child with the resolution
+  of its right child, in that order
```
rozbb: Editorial: "in that order" seems redundant or meaningless here

draft-ietf-mls-protocol.md
```diff
@@ -494,26 +502,105 @@ as follows:
...
+MLS assumes that each participant maintains a complete and
+up-to-date view of the public state of the group's ratchet tree,
+including the public keys for all nodes and the credentials
+associated with the leaf nodes.
 
-* A secret octet string (optional)
-* An asymmetric private key (optional)
-* An asymmetric public key
+No participant in an MLS group has full knowledge of the secret
+state of the tree, i.e., the secret values and private keys associated to
```
rozbb: Remove "the secret values". The only secret a node holds is an optional private key

draft-ietf-mls-protocol.md
```diff
@@ -494,26 +502,105 @@ as follows:
...
+the nodes.  Instead, each member is assigned to a leaf of the tree,
+which determines the set of secret state known to the member.  The
+credential stored at that leaf is one provided by the member.
+
+In particular, MLS maintains the members view of the tree in such a
+way as to maintain the _tree invariant_:
+
+    The secret value and private key for a node in the tree are
```
rozbb: Ditto: remove "secret value and", replace "are known" -> "is known"

draft-ietf-mls-protocol.md
```diff
@@ -494,26 +502,105 @@ as follows:
...
+    known to a member of the group if and only if that member's leaf
+    is a descendent of the node or equal to it.
```
rozbb: Typo (unless you really mean it): "descendent" -> "descendant". The latter is how it's spelled everywhere else in the doc

draft-ietf-mls-protocol.md
```diff
@@ -494,26 +502,105 @@ as follows:
...
+
+In other words, each member holds the secrets for nodes in its
```
rozbb: For specificity "secrets" -> "private keys"

draft-ietf-mls-protocol.md
```diff
@@ -494,26 +502,105 @@ as follows:
...
+direct path, and no others.
+
+
+## Ratchet Tree Updates
 
-The contents of the parent are based on the latest-updated child.
 Nodes in a tree are always updated along the "direct path" from a
```
rozbb: No need to quote "direct path" here. The term has already introduced and used non-quotatively in the previous section

draft-ietf-mls-protocol.md
```diff
@@ -494,26 +502,105 @@ as follows:
...
 leaf to the root.  The generator of the update chooses a random
 secret value "path_secret[0]", and generates a sequence of "path
```
rozbb: Auxiliary point: a path secret is made for the root node, but the root node does not appear in the direct path. Using saying that ps[n] is generated "along the direct path" may not be wrong, but it does seem a bit misleading. Thoughts?

draft-ietf-mls-protocol.md
```diff
@@ -566,107 +653,66 @@ above:
...
+the group can use these nodes to update their view of the tree,
+aligning their copy of the tree to the sender's.
 
 To perform an update for a leaf, the sender transmits the following
 information for each node in the direct path of the leaf:
 
 * The public key for the node
-* Zero or more encrypted copies of the node's parent secret value
+* Zero or more encrypted copies of the path secret corresponding to
+  the node's parent
```
rozbb: Double check this with me: If I am at the leaf, I am _my_ public key, and zero encrypted copies of _my_ path secret (this is `pk(ns[0])   |   [nothing]` in the diagram)

If I am a level up, I am encrypting _my_ public key, and 1 encrypted copy of _my_ path secret (this is `pk(ns[1])   |   E(pk(A), ps[1])` in the diagram).

So this should say:

* The public key for the node
* Zero or more copies of the node's path secret, encrypted for someone in the resolution of the node's sibling

Or, since that last detail is mentioned below, just "Zero or more encrypted copies of the node's path secret"

draft-ietf-mls-protocol.md
psyoptix: To me "in that order" somewhat clarifies that "left.child || right.child" is meant here and not "right.child || left.child". Clear yet though would be something like: 

The resolution of a blank intermediate node is the result of concatenating the resolution of its right child to the end of the resolution of its left child.

draft-ietf-mls-protocol.md
```diff
@@ -814,11 +866,50 @@ struct {
...
+struct {
+  uint8 hash_type = 1;
+  HPKEPublicKey public_key;
+  opaque left_hash<0..255>;
+  opaque right_hash<0..255>;
+} ParentNodeHashInput
+~~~~~
+
+The `left_hash` and `right_hash` fields hold the hashes of the
+node's left and right children, respectively.  The `public_key_hash`
```
psyoptix: The text description doesn't seem to match the structs. 
 * The `public_key_hash` field is missing from the `ParentNodeHashInput` struct. To make it consistent with the text I think the `public_key` should be replaced with `opaque public_key_hash<0..255>;` or something like that. 
 * Also, for completeness, we might want to add something like the following struct as it's mentioned in the text (e.g. to make the `hash_type` field precise) :

`struct {
  uint8 hash_type = 2;
  optional <HPKEPublicKey> node_public_key;
} ParentNodeInfo`

More generally though, I have some comments/questions about this section:
* Why include `LeafNodeInfo` in `LeafNodeHashInput` directly but for parent nodes use *the hash of* `ParentNodeInfo` in `ParentNodeHashInput` instead (at least according to the text)? I don't see why the extra hash of NodeInfo makes sense for parents but not for leaves... Either both or neither no?
* What is the purpose of `hash_type`? If domain separation why is that needed and why is hash_type sufficient?
* On a related note, as defined now all blank leaves will have the same hash. Even internal nodes can have the same hash if they are both roots of identical sub-trees of blank nodes. I'm not sure its a problem... but if so one clean way to avoid this is to including `leaf_index` in the `LeafNodeHashInput` struct as a tie-breaker.

draft-ietf-mls-protocol.md
```diff
@@ -1479,7 +1573,8 @@ to be used for its own sending chain:
            application_secret_[sender]_[0]
 ~~~~~
 
-Note that [sender] represents the index of the member in the roster.
+Note that [sender] represents the index of the member among the
+leaves of the ratchet tree.
```
psyoptix: Make clear if node or leaf index is meant here. E.g. "represents the index of the member among the leaves of the ratchet tree" --> "represents the leaf index of the ratchet tree leaf assigned to the sender"

draft-ietf-mls-protocol.md
bifurcation: I clarified in the next paragraph, "for each node in the direct path of the leaf, as well as
the root".  I wonder if we should just change the definition of dirpath to include the root?

draft-ietf-mls-protocol.md
bifurcation: Correct.  The encrypted copies are for the resolution of the non-updated child, and the leaf node has no children.  And yes, the "parent" in the second bullet shouldn't be there.

draft-ietf-mls-protocol.md
bifurcation: Done.

draft-ietf-mls-protocol.md
```diff
@@ -494,26 +502,105 @@ as follows:
...
+## Ratchet Tree Nodes
 
-Ratchet trees are used for generating shared group secrets. In this
-section, we describe the structure of a ratchet tree.  A particular
-instance of a ratchet tree is based on the following cryptographic
-primitives, defined by the ciphersuite in use:
+A particular instance of a ratchet tree is based on the following
+cryptographic primitives, defined by the ciphersuite in use:
 
 * A Diffie-Hellman finite-field group or elliptic curve
```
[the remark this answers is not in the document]
bifurcation: Good point.  Refactored to say we need:

1. An HPKE ciphersuite (including KEM)
2. A Derive-Key-Pair function that produces a key pair for that KEM from a secret

draft-ietf-mls-protocol.md
bifurcation: My thinking with the pre-hashing of the parent nodes values is that if someone were providing incomplete views of the tree together with proofs, then the proof could just include the hash instead of the value.  But on further reflection, I think this is a mistake since (a) the only contents of the node are an `optional<HPKEPublicKey>` and (b) that's not meaningfully different from its hash.  So I've just eliminated the `_hash` part.

Re `hash_type` - TBH, I'm basically just copy/pasting from [Certificate Transparency](https://tools.ietf.org/html/rfc6962#section-2.1) here; open to arguments either way.  Maybe @beurdouche has some thoughts here?

Re blank nodes - Even more, a blank subtree anywhere in the tree will have the same value.  I had considered just throwing the node index into all the hashes for diversity.  But I don't really see an issue here -- why do we need to distinguish between blanks in different places? Given that in the ultimate root hash they'll end up tied with non-blank stuff and locked into their position that way.

psyoptix: Thought a bit more about hash_type. For Merkle Trees (as in Certificate Transparency) the prefix to the hash inputs are definitely needed to separate leaf and internal node hashing domains. Otherwise MTs trivially aren't binding. E.g. given any target data set d={d(0),...,d(n)} it would be easy to make a new data point {d'} such that MT(d) = MT(d').

For MLS I'm suspect tree hashes may already be binding even without including hash_type. But justifying that would require subtle & messy reasoning including exactly how optional structs, HPKEPublicKey, SignaturesPK, Certificates, etc. are all serialized.

But that discussion can be entirely avoided by including hash_type as you have done. To see binding: hash_type ensures that for 2 tree's to have the same tree hash value they must have the same set of nodes and edges. (Otherwise two hash calls has differing hash_type inputs but results in the same output which is a collision on the underlying hash func). That in turn, means the trees must have the same set of labels for all nodes since otherwise we again have a collision somewhere. Ergo tree hash is binding.

Much cleaner. :-) So I've come round to the inclusion of hash_type even if I'm not convinced its strictly necessary.

bifurcation: Thanks!  Belt and suspenders!

draft-ietf-mls-protocol.md
```diff
@@ -566,107 +654,67 @@ above:
...
+| pk(ns[2])  | E(pk(C), ps[2]), E(pk(D), ps[2]) |
+| pk(ns[1])  | E(pk(A), ps[1])                  |
+| pk(ns[0])  |                                  |
 
-| Public Key | Ciphertext(s)            |
-|:-----------|:-------------------------|
-| pk(G)      | E(pk(C), G), E(pk(D), G) |
-| pk(E)      | E(pk(A), E)              |
-| pk(B)      |                          |
+Here pk(X) represents the public key corresponding derived from the
```
beurdouche: I had trouble to understand the sentence for some reason... :)

draft-ietf-mls-protocol.md
```diff
@@ -494,27 +502,107 @@ as follows:
...
+a corresponding _hash_ that summarizes the contents of the subtree
+below that node.  The rules for computing these hashes are described
+in {{tree-hashes}}.
+
+## Views of a Ratchet Tree {#views}
 
-* A secret octet string (optional)
-* An asymmetric private key (optional)
-* An asymmetric public key
+MLS assumes that each participant maintains a complete and
```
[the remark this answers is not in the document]
bifurcation: Softened to "we generally assume".

draft-ietf-mls-protocol.md
```diff
@@ -566,107 +654,67 @@ above:
...
+sync and up to date.  When a client proposes a change to the tree
+(e.g., to add or remove a member), it transmits a set of values for
+intermediate nodes in the direct path of a leaf. Other members of
+the group can use these nodes to update their view of the tree,
+aligning their copy of the tree to the sender's.
 
 To perform an update for a leaf, the sender transmits the following
-information for each node in the direct path of the leaf:
+information for each node in the direct path of the leaf, as well as
+the root:
```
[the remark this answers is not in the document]
bifurcation: That "generalized sibling" is the resolution of the sibling.

draft-ietf-mls-protocol.md
```diff
@@ -566,107 +654,67 @@ above:
...
-* Zero or more encrypted copies of the node's parent secret value
+* Zero or more encrypted copies of the path secret corresponding to
+  the node
 
-The secret value is encrypted for the subtree corresponding to the
+The path secret value for a given node is encrypted for the subtree corresponding to the
 parent's non-updated child, i.e., the child on the copath of the leaf node.
-There is one encrypted secret for each public key in the resolution
+There is one encrypted path secret for each public key in the resolution
 of the non-updated child.  In particular, for the leaf node, there
```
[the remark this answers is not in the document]
bifurcation: The resolution stage accounts for blanks.

draft-ietf-mls-protocol.md
```diff
@@ -566,107 +654,67 @@ above:
...
 are no encrypted secrets, since a leaf node has no children.
 
 The recipient of an update processes it with the following steps:
 
-1. Compute the updated secret values
+1. Compute the updated path secrets
   * Identify a node in the direct path for which the local member
```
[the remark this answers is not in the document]
bifurcation: Let's discuss in a follow-on.

draft-ietf-mls-protocol.md
bifurcation: Probably because it's poorly written ;)  I rewrote it to hopefully be clearer.

PR #135, opened 2019-03-06
 Address difference between roster and tree index 

kkohbrok:
Just stumbled across the difference again. I being a little more explicit about this a good way of preventing implementation error regarding mix-up of indices.


bifurcation:
I agree that this should not be merged.  There is a factor-of-2 issue here (which I have run into in my implementation), but I think the right answer is to remove the distinction between leaves and roster slots, as suggested in #134.

PR #136, opened 2019-03-06
Fixed example in 'Ratchet Tree Update' section

kkohbrok:
Previously, it was KEMing to the root key, which is not in the resolution of the coPath of B.


bifurcation:
This PR now seems like a noop, so closing.  Feel free to reopen if there's actually something to do here.

Issue #129, opened 2019-02-25
Fix issues introduced in #114


bifurcation:
The PR from @kkohbrok introducing a KDF was merged before I had a chance to review it.  There are a number of problems with it:

* Instead of calling for an abstract KDF, it should define one, as we have with the use of HKDF elsewhere
* We can't just say `private_key = KDF(...)`, since KDFs produce octet strings, not private keys.  We need `Derive-Key-Pair()`, not just `Derive-Public-Key()`.
* This doesn't match what I recall discussing on the list.  

I think what you want is actually as follows:

```
path_secret[n+1]
     ^
     |
path_secret[n] --> node_secret --> node_key_pair
```

Or in prose:

```
node_secret = HKDF-Expand-Label(path_secret[n], "node", "", Hash.Length)
node_private, node_public = Derive-Key-Pair(node_secret)
path_secret[n+1] = HKDF-Expand-Label(path_secret[n], "path", "", Hash.Length)
```

kkohbrok:
That was exactly the question I was asking on the list, i.e., if we need an explicit "derive-keypair" function. The idea with deriving the private key directly was to avoid a redundant derive-key operation. I'm totally fine re-doing the PR to represent either approach.

beurdouche:
Merging that PR too quickly is on me.
For the parent derivation, we can just use what the formal model for TreeKEM does, which is roughly the `HKDF_Expand_Label` Richard points out. The part I don't like about this, is to define `Derive-Key-Pair` in the MLS specification as it depends on the HPKE internal algorithm.

The TreeKEM formal specification, it is done the following way (because we know that the HPKE interface we use can take random bytes as a secret key):
```
(** Key Derivation: compute a parent secret from the group secret *)
val derive_parent_secret: alg:algorithm_hash -> context:bytes -> secret_s -> Tot secret_s
let derive_parent_secret alg context s =
  Spec.HKDF.hkdf_expand_label alg s "parent" context size_secret

(** Key Derivation: compute a KEM encryption key from the group secret *)
val derive_encryption_secret: alg:algorithm_hash -> context:bytes -> secret_s -> Tot key_secret_s
let derive_encryption_secret alg context s =
  Spec.HKDF.hkdf_expand_label alg s "kem enc" context size_key_secret
```

In the MLS specification, I suggest we avoid defining anything new but point to the HPKE spec instead.
[...]

PR #139, opened 2019-03-11
Initial definition of the KDF and the Derive-Key-Pair function

beurdouche:
Initial attempt to solve #129. This defines the KDF as discussed in #129 and relocates the DH computation obligation and verification as part of the Derive-Key-Pair function definition.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -494,7 +494,7 @@ as follows:
 used for the Merkle trees in the Certificate Transparency protocol
 {{?I-D.ietf-trans-rfc6962-bis}}.)
 
-## Ratchet Tree Nodes
+## Ratchet Tree Nodes and Tree Updates
 
 Ratchet trees are used for generating shared group secrets. In this
 section, we describe the structure of a ratchet tree.  A particular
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -549,6 +537,62 @@ X, then the tree will have the following structure.
...
+path_secret[n] --> node_secret --> node_key_pair
+~~~~~
+
+~~~~~
+node_secret = HKDF-Expand-Label(path_secret[n], "node", "", Hash.Length)
+node_private, node_public = Derive-Key-Pair(node_secret)
+path_secret[n+1] = HKDF-Expand-Label(path_secret[n], "path", "", Hash.Length)
+~~~~~
+
+where Derive-Key-Pair can be defined as a function where given
```
bifurcation: Everything from here down should be deleted, and replaced with the existing definitions that have been deleted in this PR.

draft-ietf-mls-protocol.md
```diff
@@ -503,28 +503,16 @@ primitives, defined by the ciphersuite in use:
...
 * A Key Derivation Function (KDF)
-* A Derive-Public-Key function that produces a public key from a private key
+* A Derive-Key-Pair function that produces an asymmetric keypair
+  from a node secret
 
 A ratchet tree is a left-balanced binary tree, in which each node
 contains up to three values:
 
-* A secret octet string (optional)
+* A path secret octet string (optional)
```
bifurcation: Nope, this is the node secret, which I don't think needs the extra word.

draft-ietf-mls-protocol.md
```diff
@@ -503,28 +503,16 @@ primitives, defined by the ciphersuite in use:
...
 * An asymmetric private key (optional)
 * An asymmetric public key
 
-The private key for a node is derived from its secret value using the KDF. The
```
bifurcation: There's no reason to delete this para.  It's still true that the key pair from the node derives from the node secret.  You just need to edit it to revert back to what it said before #114, namely that the key pair comes from the node secret via Derive-Key-Pair.

draft-ietf-mls-protocol.md
```diff
@@ -549,6 +537,62 @@ X, then the tree will have the following structure.
 X    B          C    D
 ~~~~~
 
+A node secret is derived from its path secret. The path secret
```
bifurcation: I would move this back up to where you deleted a chunk. Instead of "The contents of a parent node...", you should say something like "The tree is always updated along a direct path from a leaf to the root... [path secrets, node secrets]"

bifurcation:
Fixing my own review comments here, since @beurdouche appears to be offline and we need to publish -04.

PR #141, opened 2019-03-27
Added public_key_index to Add message

rozbb:
The original text says "Set the leaf node in the tree at position index to a new node containing the public key from the UserInitKey in the Add corresponding to the ciphersuite in use". Although it's never explicitly stated, the implication is that a UserInitKey can have at most 1 DhPublicKey per ciphersuite.

I think a `public_key_index` could make errors somewhat nicer. I would rather have the issue be that the Adder picked incomensurate ciphersuite, as opposed to saying that the option to Add exists, but there's no consistent way to pick the public key index.

Including an index could also open up the option of publishing large, long-lived UserInitKeys that can get downloaded and used over time (I'm not sure if this fits our model as-is, though).

Thoughts?

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1182,19 +1182,28 @@ member:
 
 ~~~~~
 struct {
-    uint32 index;
+    uint32 tree_index;
     UserInitKey init_key;
+    uint16 public_key_index;
     opaque welcome_info_hash<0..255>;
 } Add;
 ~~~~~
 
-The `index` field indicates where in the tree the new member should
+The `tree_index` field indicates where in the tree the new member should
 be added.  The new member can be added at an existing, blank leaf
-node, or at the right edge of the tree.  In any case, the `index`
-value MUST satisfy `0 <= index <= n`, where `n` is the size of the
-group. The case `index = n` indicates an add at the right edge of
-the tree).  If `index < n` and the leaf node at position `index` is
+node, or at the right edge of the tree.  In any case, the `tree_index`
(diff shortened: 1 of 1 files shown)


rozbb:
Ah, ok. Nevermind that last point then. This is just a small semantic matter. I think the adder should explicitly pick a pubkey, rather than letting it be implicitly chosen, subject to an assumed constraint. 

One concern here is also misuse. While writing my implementation, I was tempted to search for the first index of `cipher_suites` that matched the current ciphersuite, and then use that. But that would be a subtle bug, and the lack of a check there would put GroupStates out of sync. I think specifying an index and making sure the ciphersuite matches is less error-prone.

Also, the spec does not currently specify any constraints on `cipher_suites`, other than the one relating to its length. If this PR is rejected, I'll happily disambiguate this in a separate PR.

beurdouche:
The way you pick the ciphersuite is left to you and if you pick ciphersuite J in the list, you pick key J. Do you mean that the one-to-one key/ciphersuite mapping is not explicit enough in the sentence I pointed earlier ?

rozbb:
I thought there's no choice involved. The only `init_key` you _can_ pick is the one that corresponds to the current ciphersuite. I just think that that method, while perfectly well-defined and doable, is more prone to error than having the adder explicitly choose the `init_key` index and including it in the Add message.

And yes, I believe that the one-to-one key/ciphersuite correspondence is not explicit enough. I didn't know that that was a condition until I read the Add subsection of the Handshake section in the spec.

rozbb:
Closing because the `index -> tree_index` change is wrong and doesn't belong here. Also I shouldn't have based this on my `master`. I'll redo this change in a sec.

PR #143, opened 2019-04-03
Fixes to tree manipulation in Remove

rozbb:
This clears up the wording around how to prune the ratchet tree on a Remove operation. Previously, it could be interpreted as "remove nodes from the tree until the rightmost node is not null", but this is not what is intended by the authors, and it also has the ability to produce invalid trees (i.e., trees with an even number of elements).

The second change is to do the truncation _after_ doing the Blank propogation up the direct path of the removed leaf. Note that the order of these steps doesn't matter in that they produce the same resulting tree. However, if we blank after pruning, an implementor would have to take care to not to try to blank out previously-removed nodes. Prune-then-blank is a little easier to reason about.

The third change is a semantic fix. It is clear from the last paragraph in this section that the root node should be blanked, but the text only said to blank the direct path of the leaf, which does not include the root node. I just added that part in.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1139,14 +1139,6 @@ for the new member using HPKE.  The recipient key pair for the
 HPKE encryption is the one included in the indicated UserInitKey,
 corresponding to the indicated ciphersuite.
 
-We say a "pre-Add" GroupState is a GroupState that was created from a
-a WelcomeInfo object and has not yet received the corresponding Add
-message. A pre-Add GroupState does not have a defined signer key and
-its roster does not have an entry with the public key corresponding to
-the member's identity key. Thus, a pre-Add GroupState is incapable of
-processing Update and Remove messages. A member in a pre-Add state MAY
-therefore ignore all Handshake messages that are not an Add operation.
-
 ~~~~~
 struct {
   ProtocolVersion version;
@@ -1306,13 +1298,11 @@ state as follows:
 * Truncate the tree such that the rightmost non-blank leaf is the
   last node of the tree
 
-Note that there must be at least one non-null element in the roster.
-For if this GroupState is pre-Add, then it is incapable of processing
-a Remove operation at all. And if this GroupState is not pre-Add, then
-it must have the current member in the roster. Since self-removal is
-prohibited, at least 1 non-null element must remain. The same
-reasoning justifies the existence of a non-blank leaf in the ratchet
-tree.
+Note that there must be at least one non-null element in the roster,
+since any valid GroupState must have the current member in the roster
+and self-removal is prohibited. Thus, at least 1 non-null element must
+remain. The same reasoning justifies the existence of a non-blank leaf
+in the ratchet tree.
 
 The update secret resulting from this change is the secret for the
 root node of the ratchet tree after the second step (after the third
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1271,9 +1271,11 @@ root node of the ratchet tree.
 
 ## Remove
 
-A Remove message is sent by a group member to remove one or more
+A Remove message is sent by a group member to remove one or more other
 members from the group. A member MUST NOT use a Remove message to
-remove themselves from the group.
+remove themselves from the group. If a member of a group receives a
+Remove message where the removed index is equal to the signer index,
+the recipient MUST reject the message as malformed.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1287,10 +1287,13 @@ state as follows:
   the null optional value
 * Update the ratchet tree by replacing nodes in the direct
   path from the removed leaf using the information in the Remove message
-* Reduce the size of the roster and the tree until the rightmost
-  element roster element and leaf node are non-null
 * Update the ratchet tree by setting to blank all nodes in the
-  direct path of the removed leaf
+  direct path of the removed leaf, and also setting the root node
+  to blank
```
bifurcation: This is redundant, given that the root path is on the direct path of every node.  Unless there's some definition that comes up one short?

draft-ietf-mls-protocol.md
```diff
@@ -1287,10 +1287,13 @@ state as follows:
...
+* Truncate the roster such that the last roster element is
+  non-null (clearing the roster if no such element exists)
+* Truncate the tree such that the rightmost non-blank leaf is the
+  last node of the tree (clearing the tree if no such leaf exists)
```
bifurcation: I would prefer not to have the "(clearing..." parts.  There's no such thing as a zero-member group, so it would be better to simply disallow Remove on a one-member group.

bifurcation: Though now that I say that, I realize that requirement is not totally trivial to enforce, since the last remaining member could have any index.  But it still seems like the most semantically clear approach.

rozbb: What's the issue exactly with a zero-member group? You can't semantically do anything with it, so you just get a useless object at the end of the `Remove`, and you have no choice but to throw it out.

draft-ietf-mls-protocol.md
rozbb: From section 5.1: "The direct path of a root is the empty list, and of any other node is the concatenation of that node with the direct path of its parent."

Direct paths do not contain the root node.

draft-ietf-mls-protocol.md
bifurcation: Well, that's the problem -- you can't do anything with it.  So it seems like it shouldn't be a valid state.

rozbb: Ah, I was going to say "bad states are states that are semantically incorrect, and there's nothing incorrect about an empty group" but I found a reason why it's not correct (at least for me): `signer_index` is not `optional` in my GroupState object. If the roster is empty, then there is no valid value for `signer_index`. Since this is necessarily an invalid state, I agree that zero-member groups should be disallowed. I'll make an update to reflect this prohibition for Removes.

draft-ietf-mls-protocol.md
```diff
@@ -1264,7 +1270,8 @@ root node of the ratchet tree.
 ## Remove
 
 A Remove message is sent by a group member to remove one or more
-members from the group.
+members from the group. A member MUST NOT use a Remove message to
+remove themselves from the group.
```
bifurcation: The receiver POV here seems more useful, i.e., "If a member of a group receives a Remove message where the `removed` index is equal to the `sender` index, then the recipient MUST reject the message as malformed."  That's not to say we can't have both, though.

Might tee this up better if you said "remove one or more **other** members from the group" (emphasis only for here, not for the doc).

draft-ietf-mls-protocol.md
```diff
@@ -1287,10 +1294,16 @@ state as follows:
...
-  direct path of the removed leaf
+  direct path of the removed leaf, and also setting the root node
+  to blank
+* Truncate the roster such that the last roster element is
+  non-null (there must be at least one such an element, since a
+  pre-Add GroupState cannot process a Remove, a non-pre-Add GroupState
+  must have the current member in the roster, and self-removal is
+  prohibited)
+* Truncate the tree such that the rightmost non-blank leaf is the
+  last node of the tree (clearing the tree if no such leaf exists)
```
bifurcation: We're coming down to kind of a matter of taste here, since I don't really like discussing things that are impossible :)  How about we compromise with the following:

* Leave these clarifications out of the instruction bullets
* Add a para afterward of the form "Note that there is always at least one non-blank..."

rozbb: agreed. That will also give room to mention that the same argument holds for tree truncation

PR #147, opened 2019-04-17
Added public_key_index to Add message

rozbb:
This is a more correct version of #141. See there for prior conversation.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1136,8 +1136,8 @@ The Welcome message contains the information that the new member
 needs to initialize a GroupState object that can be updated to the
 current state using the Add message.  This information is encrypted
 for the new member using HPKE.  The recipient key pair for the
-HPKE encryption is the one included in the indicated UserInitKey,
-corresponding to the indicated ciphersuite.
+HPKE encryption is the one included in the indicated UserInitKey at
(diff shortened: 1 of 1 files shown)


bifurcation:
I would prefer that we not take this approach, but instead tighten up the requirements for uniqueness in the cipher_suites array, in which case there is no ambiguity in the current approach.

If we're going to go this explicit-index route, then you would at least need to require that the recipient verify that `cipher_suites[i] == group.cipher_suite`.  But that seems more fragile than indexing by cipher suite, since the recipient can omit this check and end up trying to re-use key shares across groups.

rozbb:
I argue that `cipher_suites[i] == group.cipher_suite` is actually less fragile than indexing by ciphersuite. I have ciphersuite information encoded in all my DH and AEAD key types, so even if I were to forget that check, I would get a runtime error very quickly, which is a good thing.

But if I were to forget to check for uniqueness of ciphersuites, it could result in two members deriving disagreeing yet valid GroupStates. I think that this situation is harder to diagnose than the above.

Still, I won't push too hard on this if everyone else is in agreement that it's not necessary.

bifurcation:
I am inclined to close this with no action.  @beurdouche @raphaelrobert wdyt?

beurdouche:
@bifurcation Yes, I agree with closing this.

PR #148, opened 2019-04-17
Added that users SHOULD verify pubkeys in an Update

rozbb:
Adding this check would be the first step in recognizing that an Update operation may have maliciously excluded a person from the group.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -637,13 +637,13 @@ The recipient of an update processes it with the following steps:
     the private key from the resolution node
   * Derive secret values for ancestors of that node using the KDF keyed with the
     decrypted secret
+  * Members MUST verify that the received public keys agree with the
+    public keys derived from the new node_secret values
 2. Merge the updated secrets into the tree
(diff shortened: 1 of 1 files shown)


bifurcation:
This would go better in the "Ratchet Tree Updates" section, right after "Derive secret values for ancestors of that node using the KDF keyed with the decrypted secret".  That would align better with where in the code you want the checks to be.  Other than that, LGTM.

rozbb:
Fixed

draft-ietf-mls-protocol.md
```diff
@@ -637,6 +637,8 @@ The recipient of an update processes it with the following steps:
     the private key from the resolution node
   * Derive secret values for ancestors of that node using the KDF keyed with the
     decrypted secret
+  * Members MUST verify that the received public keys agree with the
```
bifurcation: Editorial: "Members" -> "The recipient"

rozbb: Done

beurdouche:
Thanks Michel

beurdouche:
Thanks Michael !

PR #149, opened 2019-04-17
Did a bunch of clarification around path secrets

rozbb:
A few things:

1. Restructured the sections "Ratchet Tree Nodes and Tree Updates" and "Ratchet Tree Updates" into "Ratchet Tree Nodes" and "Ratchet Tree Updates". I also put them next to each other, whereas before the "Blank Nodes and Resolution" section came between them.
2. Removed node secrets from the set of things that a node must contain. There is no need for this. The node secret is used precisely once—to derive the node's keypair. It needn't be stored after that.
3. Clarified things and made the terminology more consistent. For example, the diagram in the previous "Ratchet Tree Updates" section only referred to "secrets" without specification (because there used to only be node secrets). There were also statements that referred to node secrets when they meant path secrets. These should be fixed now.
4. Renamed `RatchetNode::node_secrets` to `RatchetNode::path_secrets`

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -494,81 +494,6 @@ as follows:
 used for the Merkle trees in the Certificate Transparency protocol
 {{?I-D.ietf-trans-rfc6962-bis}}.)
 
-## Ratchet Tree Nodes and Tree Updates
-
-Ratchet trees are used for generating shared group secrets. In this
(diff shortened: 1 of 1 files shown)


bifurcation:
@rozbb Could you please verify that #134 has the required changes in it?

rozbb:
Yup, looks like everything made it in. See #134 for further comments.

PR #152, opened 2019-04-23
Make epochs unpredictable

bifurcation:
As discussed on the mailing list, further reduces the ability of the Delivery Service to infer what's going on in the group.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -137,6 +137,10 @@ shared keys with costs that scale as the log of the group size.
 
 RFC EDITOR PLEASE DELETE THIS SECTION.
 
+draft-05
+
+- Make epoch numbers unpredictable
+
 draft-04
 
 - Updating the language to be similar to the Architecture document
@@ -956,6 +960,9 @@ update_secret -> HKDF-Extract = epoch_secret
                      |
                      +--> Derive-Secret(., "app", GroupState_[n])
                      |    = application_secret
+                     |
+                     +--> Derive-Secret(., "epoch", GroupState_[n])
+                     |    = epoch_change_secret
                      |
                      +--> Derive-Secret(., "confirm", GroupState_[n])
                      |    = confirmation_key
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -849,18 +849,20 @@ its GroupState.
 
 Different group operations will have different effects on the group
 state.  These effects are described in their respective subsections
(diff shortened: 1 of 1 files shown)


psyoptix:
I see we use HMAC to derive epoch_[n] which forces explicit truncation of HMACs output (to 4 bytes). Why not use HMAC-Expand instead? It's already used in another place by MLS so should already be available and it already takes the desired output length as a third parameter.

psyoptix:
I'm sure this is a silly question so please humor me... :-)

In the GroupOperation struct there is a select command which, I think, results in one of 4 struct types (e.g. "Init", "Add", etc) being chosen depending on the value of msg_type. But I dont see the actual *name* for that struct member being defined there (say, "msg"). Is it missing...  or am I just missing something? :-)

In case there's no bug here and I'm just not familiar with that syntax could someone point me to a resource where I can read up on it?

bifurcation:
Thanks for the comments, @psyoptix 

Re: `HKDF-Expand` -- Yeah, this is a good idea.  I'll refactor to use that.  It's still just one HKDF instantiation, after all.

Re: `GroupOperation` -- According to [the syntax specification](https://tools.ietf.org/html/rfc8446#section-3.8), the field names are optional, but I would be fine including them.  Want to send a small PR for that?

psyoptix:
Yup, so as expected the issue lies with my knowledge of the correct syntax not with the doc so nevermind. :-) thanks for the reference by the way.

(and as long as we're never referring to that member there's no need in giving it a name so the definition is fine as is.)

bifurcation:
@beurdouche based on your comments in #155 do you want to push this to draft-06?

beurdouche:
> @beurdouche based on your comments in #155 do you want to push this to draft-06?

Yes, that is not a difficult issue but I think handling unpredicatable epoch and unpredictable group_id makes sense. That might be a short discussion topic at the interim ?

beurdouche:
> > @beurdouche based on your comments in #155 do you want to push this to draft-06?
> 
> Yes, that is not a difficult issue but I think handling unpredicatable epoch and unpredictable group_id makes sense. That might be a short discussion topic at the interim ?

at the same time*

raphaelrobert:
I fully support the intention of this, but it is not clear to me how the DS can enforce total order on HS messages if the epochs are no longer comparable.

bifurcation:
Closing this until we have a better theory for the overall environment (e.g., how server assist would work).

beurdouche:
This can be discussed as part of "Server assist".

PR #153, opened 2019-04-23
Common framing, consolidated

bifurcation:
I started off trying to just edit #131 to be landable, but as I looked at it, it seemed like we had accreted enough complexity that a refactor was in order.  So this PR mostly rewrites the sections related to encryption keys and message framing, hopefully in a way that's clearer to the reader.  For the most part, though, the technical bits are the same as in #131.  There are a couple of innovations here:

* Using "masking" instead of AEAD for sender data encryption (cf. [QUIC header protection](https://tools.ietf.org/html/draft-ietf-quic-tls-17#section-5.4)). This saves 28 bytes of overhead, but there's no longer authentication except in the content AEAD.  This change could be done as a follow-on PR, but it seemed brief enough to include here.

* I eliminated the `group_id` field in the messages, since it seems plausible that we could rely on applications to provide that.    The benefit of this is that the wire image that MLS itself presents (in `MLSCiphertext`) is entirely opaque to the delivery service (even if the DS might add some context that indicates the group ID).


bifurcation:
Draft implementation here: https://github.com/cisco/mlspp/pull/48

PR #155, opened 2019-04-30
Common framing, consolidated with modifications

beurdouche:
This PR is updates Richard's changes with a few adjustments:

- Merge master;
- Fix missing definition to derive key/nonce used for encryption of Handshake messages;
- Delete the "remove_confirmation" function for simplicity. It has a very low cost anyway;
- Revert to having the GID and the epoch in the clear, otherwise there is a need for a magic channel to communicate it to the recipient. We can eventually do it but this should be discussed seriously in Berlin;
- Revert the masking of the content type, (keeps masking for sender and generation) otherwise it seem to force plaintext OR the Delivery Service to have access to the sender_data_secret. It seem impossible to only order Handshake messages otherwise.


draft-ietf-mls-protocol.md
```diff
@@ -967,6 +977,100 @@ update_secret -> HKDF-Extract = epoch_secret
...
+
+As described in {{message-framing}}, MLS encrypts two different
+types of information:
+
+* Handshake messages
+* Application messages
+
+Each handshake message is encrypted using a key and a nonce derived
+from the handshake_secret for a specific sender to prevent two senders
+to perform in the following way:
```
bifurcation: This text assumes that there's only a need for one handshake message to be sent per epoch.  I realize that should be true in most cases, but I'm concerned about things like retries leading to nonce reuse.  This is why in #153, I had handshake keys follow the same path as application keys.  This also allows code reuse.

draft-ietf-mls-protocol.md
```diff
@@ -967,6 +977,100 @@ update_secret -> HKDF-Extract = epoch_secret
...
+For application messages, a chain of keys is derived for each sender
+in a similar fashion. This allows forward secrecy at the level of
+application messages within and out of an epoch.
+A step in this chain (the second subscript) is called a "generation".
+
+~~~~~
+           application_secret
+                     |
+                     V
+           HKDF-Expand-Label(., "sender", [sender], Hash.length)
```
bifurcation: s/sender/ap sender/ ?

draft-ietf-mls-protocol.md
bifurcation: Oh, I see you have conveniently not deleted that text :)  So I would just delete the text you've added up here.

draft-ietf-mls-protocol.md
```diff
@@ -610,7 +614,7 @@ In this tree, we can see all three of the above rules in play:
 ## Ratchet Tree Updates
 
 In order to update the state of the group such as adding and
-removing clients, MLS messages are used to make changes to the
+removing clients, Handshake messages are used to make changes to the
```
bifurcation: In general, I've been using lower-case "handshake message", since we don't have a Handshake struct any more.

draft-ietf-mls-protocol.md
```diff
@@ -1008,14 +1112,178 @@ struct {
...
+    (255)
+} ContentType;
+
+struct {
+    opaque group_id<0..255>;
+    uint32 epoch;
+    uint32 sender;
+    ContentType content_type;
+
+    select (MLSPlaintext.type) {
```
bifurcation: s/type/content_type/

draft-ietf-mls-protocol.md
```diff
@@ -1008,14 +1112,178 @@ struct {
...
+    opaque masked_sender_data[8];
+    opaque ciphertext<0..2^32-1>;
+} MLSCiphertext;
+~~~~~
+
+The remainder of this section describe how to compute the signature of
+an MLSPlaintext object and how to convert it to an MLSCiphertext object.
+The overall process is as follows:
+
+* Compute the metadata by concatenating the following information:
```
bifurcation: "Concatenating" is not quite right, since it's a bit more complex than that.  Maybe "Gather the required metadata:"

draft-ietf-mls-protocol.md
```diff
@@ -1008,14 +1112,178 @@ struct {
...
+
+* Sign the protected content and metadata
+
+* Encrypt the content using a content encryption key identified by
+  the metadata
+
+* Mask the sender_data used to identify the content encryption key
+  using a per-epoch key
+
+The group identifier and epoch fields are copied from the MLSPlaintext object directly.
```
bifurcation: Also content_type now, right?

draft-ietf-mls-protocol.md
```diff
@@ -1008,14 +1112,178 @@ struct {
...
+
+The ciphertext field of the MLSCiphertext object is produced by
+supplying the inputs described below to the AEAD function specified
+by the ciphersuite in use.  The plaintext input contains content and
+signature of the MLSPlaintext, plus optional padding.  These values
+are encoded in the following form:
+
+~~~~~
+struct {
+    opaque content[MLSPlaintext.length];
```
bifurcation: It occurs to me that there's not actually a `length` field in `MLSPlaintext`.  Let's just use `length_of_content`, in parallel to `length_of_padding`.

draft-ietf-mls-protocol.md
```diff
@@ -1008,14 +1112,178 @@ struct {
...
+    uint8 signature[MLSInnerPlaintext.sig_len];
+    uint16 sig_len;
+    uint8  marker = 1;
+    uint8  zero\_padding[length\_of\_padding];
+} MLSCiphertextContent;
+~~~~~
+
+The key and nonce used for this encryption depend on the content
+type of the message.  The sender chooses an handshake_key for an
```
bifurcation: "a handshake_key".  But really you need to change this to say "an unused generation from it's application or handshake key chain"

draft-ietf-mls-protocol.md
```diff
@@ -1008,14 +1112,178 @@ struct {
...
+The Additional Authenticated Data (AAD) input to the encryption
+contains an object of the following form, with the values used to
+identify the key and nonce:
+
+~~~~~
+struct {
+    opaque group_id<0..255>;
+    uint32 epoch;
+    ContentType content_type;
+    opaque masked_sender_data[8];
```
bifurcation: This is impossible.  The masking is based on the ciphertext, so you can't use the masked values as an input to the encryption.  Just use the plain values.

draft-ietf-mls-protocol.md
```diff
@@ -1100,14 +1202,181 @@ struct {
...
+    opaque content[length\_of\_content];
+    uint8 signature[MLSInnerPlaintext.sig_len];
+    uint16 sig_len;
+    uint8  marker = 1;
+    uint8  zero\_padding[length\_of\_padding];
+} MLSCiphertextContent;
+~~~~~
+
+The key and nonce used for this encryption depend on the content
+type of the message.  The sender chooses an handshake_key for an
```
bifurcation: "a handshake key"

draft-ietf-mls-protocol.md
```diff
@@ -1100,14 +1202,181 @@ struct {
...
+handshake message or an ununsed generation from its (per-sender)
```
bifurcation: "a handshake message"

draft-ietf-mls-protocol.md
bifurcation: I can live with that for now.  Please remove the para starting "The same rules are used to generate a sequence of keys and nonces for handshake encryption"

I see what you're saying, I don't feel 100% confident that we won't get wedged by this in practice.  And I dislike the extra code paths.  But we can merge with this flavor and ask the WG.

beurdouche: I agree, there might be a better solution but I don't immediately have it.
Should I add an open issue or something ?

draft-ietf-mls-protocol.md
beurdouche: Fixed, thanks !

draft-ietf-mls-protocol.md
beurdouche: Fixed, thanks !

bifurcation:
@raphaelrobert Last call on this one.  If I don't hear from you by the morning ET, I'm going to merge this.

bifurcation:
Note that this PR will be pretty immediately updated by #156, which reverts the "masking" parts to use AEAD, in order to make analysis simpler for this revision.

PR #163, opened 2019-05-17
Reorder blanking and update in the Remove operation

### LINKED ISSUE #162: Reorder blanking and update in the Remove operation.

beurdouche:
Fixes #162

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1625,7 +1625,10 @@ struct {
 
 The sender of a Remove message generates it as as follows:
 
-* Blank the path from the removed leaf to the root node
+* Blank the path from the removed leaf to the root node for
+  the time of the computation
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1617,27 +1621,26 @@ struct {
 
 The sender of a Remove message generates it as as follows:
 
+* Blank the path from the removed leaf to the root node
```
bifurcation: We should note here that the blanking should be temporary / undone after this computation.  The generation of messages should not change the sender's state.

bifurcation: Also, you should go ahead and truncate here, which will save you unnecessary derivations.  So:

* Before: Encrypt / blank
* After: Blank / Truncate / Encrypt

Same change in the processing logic below.

PR #164, opened 2019-05-17
Rename UserInitKey to ClientInitKey

### LINKED ISSUE #137: Rename UserInitKey to ClientInitKey
These keys are Client one-time use keys, so "UserInitKey" seems like a strange name.

beurdouche:
Fixes #137


bifurcation:
Overall, this seems OK.  Just using "InitKey" would be clear and less wordy.

beurdouche:
I tried this and that led people to confuse `init key` and `init secret`

Issue #159, opened 2019-05-11
Rename the GroupState structure as it is not a group state.


beurdouche:
We should change the name of this struct as it doesn't even contain the tree.

bifurcation:
Fixed by #166

PR #166, opened 2019-05-17
GroupState is not a group state, renaming to GroupContext

beurdouche:
Fix for #159


draft-ietf-mls-protocol.md
```diff
@@ -914,16 +918,16 @@ and only if the node is blank.
 
 ## Group State
 
-Each member of the group maintains a representation of the
-state of the group:
+Each member of the group maintains a partial representation of the
```
bifurcation: It's not partial, it's a summary / commitment.

"Each member of the group maintains a GroupContext object that summarizes the state of the group:"

PR #170, opened 2019-05-23
Confirmation and transcript improvements

### LINKED ISSUE #157: Resolve circular dependency between GroupOperation.confirmation and transcript_hash
```
GroupOperation.confirmation =
    HMAC(confirmation_key, GroupState.transcript\_hash)
```

bifurcation:
This PR addresses two issues that were noted in draft-05, one that is definitely a bug, and one that should improve security analysis.

* In draft-05, a circular dependency was introduced, wherein the confirmation MAC was calculated from the transcript, which includes the GroupOperation, which includes the confirmation MAC.  Here we move the confirmation MAC up to the MLSPlaintext object to avoid this circularity.

* In all prior versions, the transcript covered only the GroupOperation objects, not the sender metadata or signatures.  Here we cover the whole MLSPlaintext object, but in "staggered" form -- the confirmation MAC and signature aren't entered into the transcript until the next epoch.

Fixes #157

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -954,11 +954,25 @@ operations:
   is processed
 * The `tree_hash` is updated to represent the current tree and
   credentials
-* The `transcript_hash` is updated by a GroupOperation message
-  `operation` in the following way:
+* The `transcript_hash` is updated with the data for an MLSPlaintext
+  message encoding a group operation in two parts:
 
 ~~~~~
-transcript_hash_[n] = Hash(transcript_hash_[n-1] || operation)
+struct {
+  opaque group_id<0..255>;
+  uint32 epoch;
+  uint32 sender;
+  ContentType content_type;
+  GroupOperation operation;
+} MLSPlaintextOpContent;
+
+struct {
+  opaque confirmation<0..255>;
+  opaque signature<0..2^16-1>;
+} MLSPlaintextOpAuthData;
+
+intermediate_hash_[n] = Hash(transcript_hash_[n-1] || MLSPlaintextOpAuthData_[n-1]);
+transcript_hash_[n] = Hash(intermediate_hash_[n] || MLSPlaintextOpContent_[n]);
 ~~~~~
 
 When a new one-member group is created (which requires no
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -954,11 +961,25 @@ operations:
...
+  ContentType content_type;
+  GroupOperation operation;
+} MLSPlaintextOpContent;
+
+struct {
+  opaque confirmation<0..255>;
+  opaque signature<0..2^16-1>;
+} MLSPlaintextOpAuthData;
+
+intermediate_hash_[n] = Hash(transcript_hash_[n-1] || MLSPlaintextOpAuthData_[n-1]);
```
raphaelrobert: I'm not sure we have to introduce the notion of an intermediate hash, since it will never be used independently. We could just define the transcript hash as:

```
transcript_hash_[n] = Hash(transcript_hash_[n-1] || MLSPlaintextOpAuthData_[n-1] || MLSPlaintextOpContent_[n]);
```

bifurcation: The idea of the intermediate hash is that an implementation could go ahead and update the transcript hash to this state and be done with MLSPlaintext_[n-1], so it wouldn't need to keep that around..  I'll add some text to clarify.

draft-ietf-mls-protocol.md
```diff
@@ -954,11 +961,25 @@ operations:
...
+* The `transcript_hash` is updated with the data for an MLSPlaintext
+  message encoding a group operation in two parts:
 
 ~~~~~
-transcript_hash_[n] = Hash(transcript_hash_[n-1] || operation)
+struct {
+  opaque group_id<0..255>;
+  uint32 epoch;
+  uint32 sender;
+  ContentType content_type;
```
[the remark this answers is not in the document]
bifurcation: I agree that in practice this field will be constant, but I like keeping the alignment with the MLSPlaintext struct.  I'll add an annotation to clarify.

draft-ietf-mls-protocol.md
raphaelrobert: That makes sense

PR #171, opened 2019-05-24
Specify an Init message

bifurcation:
This is something we've been meaning to do for a while, and just haven't gotten around to.

The Init message directly creates a group with a specified member list.  The cost of group creation is O(N) DH operations for the sender, and O(1) for receivers.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1501,7 +1501,7 @@ struct {
   opaque group_id<0..255>;
   ProtocolVersion version;
   CipherSuite cipher_suite;
-  UserInitKey members<0..2^32-1>;
+  ClientInitKey members<0..2^32-1>;
   DirectPath path;
 } Init;
 ~~~~~
@@ -1524,10 +1524,16 @@ the Init message as follows:
 * Construct a ratchet tree as above
 * Update the cached ratchet tree by replacing nodes in the direct
   path from the first leaf using the direct path
+* Update the cached ratchet tree by replacing nodes in the direct
+  path from the first leaf using the information contained in the
+  "path" attribute
 
 The update secret for this interaction, used with an all-zero init
 secret to generate the first epoch secret, is the `path_secret[i+1]`
-derived from the `path_secret[i]` associated to the root node.
+derived from the `path_secret[i]` associated to the root node.  The
+members learn the relevant path secrets by decrypting one of the
+encrypted path secrets in the DirectPath and working back to the
+root (as in normal DirectPath processing).
 
 ## Add
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1535,6 +1535,14 @@ members learn the relevant path secrets by decrypting one of the
 encrypted path secrets in the DirectPath and working back to the
 root (as in normal DirectPath processing).
 
+[[ OPEN ISSUE: This approach leaks the initial contents of the tree
+to the Delivery Service, unlike the sequential-Add case. ]]
+
+[[ OPEN ISSUE: It might be desireable for the group creator to be
+able to "pre-warm" the tree, by providing values for some nodes not
+on its direct path.  This would violate the tree invariant, so we
+would need to figure out what mitigations would be necessary. ]]
+
 ## Add
 
 In order to add a new member to the group, an existing member of the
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1452,11 +1452,43 @@ message with invalid values for public keys in the ratchet tree. ]]
...
+the Init message as follows:
+
+* Note the group ID, protocol version, and cipher suite in use
+* Construct a ratchet tree as above
+* Update the cached ratchet tree by replacing nodes in the direct
+  path from the first leaf using the direct path
+
+The update secret for this interaction, used with an all-zero init
+secret to generate the first epoch secret, is the `path_secret[i+1]`
+derived from the `path_secret[i]` associated to the root node.
```
raphaelrobert: It's not clear to me how the members learn about the path secrets. The creator would have to KEm secrets to each leaf node individually I suppose.

draft-ietf-mls-protocol.md
```diff
@@ -1452,11 +1452,43 @@ message with invalid values for public keys in the ratchet tree. ]]
...
+using adding members individually.  For cases where the initial list
+of members is known, the Init message allows a group to be created
+more efficiently.
+
+~~~~~
+struct {
+  opaque group_id<0..255>;
+  ProtocolVersion version;
+  CipherSuite cipher_suite;
+  UserInitKey members<0..2^32-1>;
```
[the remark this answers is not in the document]
bifurcation: Yes, we have.  And in general, this needed a merge from master, which I have done.

draft-ietf-mls-protocol.md
```diff
@@ -1452,11 +1452,43 @@ message with invalid values for public keys in the ratchet tree. ]]
...
+  DirectPath path;
```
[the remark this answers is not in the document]
bifurcation: You would actually want a giant vector of `DirectPathNode` values, where each private key was encrypted to the `2^level` receipients below it.  But I agree we should punt this to a future version.

draft-ietf-mls-protocol.md
bifurcation: The members get path secrets from the DirectPath, just like in an Update.  Note that the DirectPath `path` attribute here has linear size: Each path secret is encrypted directly to the leaves under it.  I've added a clarification.

draft-ietf-mls-protocol.md
```diff
@@ -1491,11 +1491,49 @@ message with invalid values for public keys in the ratchet tree. ]]
...
+  path from the first leaf using the direct path
+* Update the cached ratchet tree by replacing nodes in the direct
+  path from the first leaf using the information contained in the
+  "path" attribute
+
+The update secret for this interaction, used with an all-zero init
+secret to generate the first epoch secret, is the `path_secret[i+1]`
+derived from the `path_secret[i]` associated to the root node.  The
+members learn the relevant path secrets by decrypting one of the
+encrypted path secrets in the DirectPath and working back to the
```
beurdouche: nit. s/working back to/derive secrets up-to/ ?

draft-ietf-mls-protocol.md
```diff
@@ -1491,11 +1491,49 @@ message with invalid values for public keys in the ratchet tree. ]]
...
+Each member of the newly-created group initializes its state from
+the Init message as follows:
+
+* Note the group ID, protocol version, and cipher suite in use
+* Construct a ratchet tree as above
+* Update the cached ratchet tree by replacing nodes in the direct
+  path from the first leaf using the direct path
+* Update the cached ratchet tree by replacing nodes in the direct
+  path from the first leaf using the information contained in the
+  "path" attribute
```
beurdouche: nit. `adequate` seems inadequate; replace by `field` ?

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
beurdouche: I think the current scenario looks fine. In the scenario were we would break the secrecy invariant and allow warmup and pre-populating nodes, there is an performance optimization: we can just send the public and the secret part of the tree independently to avoid large numbers of PKE of a large payload. In that scenario I would suggest encrypting the pre-warmed public tree encrypted under a fresh AEAD key/nonce and PKE encrypt the key and the nonce to the recipient along with the new path secret of its parent.

beurdouche:
This now has an OPEN ISSUE regarding the fact that we'll need to add PKE encryption for that message, as agreed out of band, let's merge this and add encryption in the next iteration.

PR #172, opened 2019-05-24
DirectPathNode contains encrypted node secrets, not path secrets

rozbb:
The paragraph below the changed field refers to `node_secrets`, not `encrypted_path_secrets`.


bifurcation:
This is incorrect.  The sender needs to transmit the path secret, so that the receivers can compute the hashes up the tree.  

You're correct that there's a mismatch between the struct definition and the prose below.  I've fixed this in #176.

PR #184, opened 2019-06-26
Clarify how HPKE ciphertexts are computed

rozbb:


draft-ietf-mls-protocol.md
```diff
@@ -1037,8 +1037,16 @@ the length of the resolution of the corresponding copath node.  Each
 ciphertext in the list is the encryption to the corresponding node
 in the resolution.
 
-The HPKECiphertext values are computed according to the Encrypt
-function defined in {{!I-D.barnes-cfrg-hpke}}.
+The HPKECiphertext values are computed as
+
+~~~~~
+context = SetupBaseI(node_public_key, "")
```
bifurcation: This should be `ephemeral_key, context`.  The lack of the `ephemeral_key` output is a bug in the current spec, and will be fixed in the [upcoming version](https://github.com/cfrg/draft-irtf-cfrg-hpke/blame/master/draft-irtf-cfrg-hpke.md#L323).

PR #187, opened 2019-07-08
Simplify AS tree/ratchet computations

bifurcation:
This PR implements a collection of comments on #146.  It aims to achieve the goal stated by @psyoptix of having every derivation have a different relationship, but with a simpler construction.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1073,7 +1073,7 @@ struct {
 } HkdfLabel;
 
 Derive-Secret(Secret, Label) =
-    Derive-In-Group(Secret, Label, "", Hash.length)
+    HKDF-Expand-Label(Secret, Label, "", Hash.length)
 ~~~~~
 
 The Hash function used by HKDF is the ciphersuite hash algorithm.
@@ -1905,8 +1905,8 @@ tree with the same set of nodes and edges as the epoch's ratchet tree. Each
 leaf in the AS Tree is associated with the same group member as the
 corresponding leaf in the ratchet tree. Nodes are also assigned an index
 according to their position in the array representation of the tree (described
-in {{tree-math}}). If V is a node index in the AS Tree then left(V)
-and right(V) denote the children of V (if they exist).
+in {{tree-math}}). If N is a node index in the AS Tree then left(N)
+and right(N) denote the children of N (if they exist).
 
 Each node in the tree is assigned a secret. The root's secret is simply the
 application_secret of that epoch. (See {{key-schedule}} for the definition of
@@ -1921,7 +1921,7 @@ using a call to Derive-App-Secret.
 
 ~~~~
 Derive-App-Secret(Secret, Label, Node, Generation, Length) =
-    Derive-In-Group(Secret, Label, ApplicationContext, Length)
+    HKDF-Expand-Label(Secret, Label, ApplicationContext, Length)
 
 Where ApplicationContext is specified as:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1066,13 +1066,14 @@ HKDF-Expand-Label(Secret, Label, Context, Length) =
...
+  opaque group_context<0..255> = Hash(GroupContext_[n]);
+  uint16 length = Length;
+  opaque label<7..255> = "mls10 " + Label;
+  opaque context<0..2^32-1> = Context;
 } HkdfLabel;
 
-Derive-Secret(Secret, Label, Context) =
-    HKDF-Expand-Label(Secret, Label, Hash(Context), Hash.length)
+Derive-Secret(Secret, Label) =
+    Derive-In-Group(Secret, Label, "", Hash.length)
```
beurdouche: This is undefined.

draft-ietf-mls-protocol.md
```diff
@@ -1904,48 +1905,44 @@ tree with the same set of nodes and edges as the epoch's ratchet tree. Each
 leaf in the AS Tree is associated with the same group member as the
 corresponding leaf in the ratchet tree. Nodes are also assigned an index
 according to their position in the array representation of the tree (described
-in {{tree-math}}). If V is a node in the AS Tree then IndexOf(V) denotes it's
-index while V.leftChild and V.rightChild denote the children of V (if they
-exist).
+in {{tree-math}}). If V is a node index in the AS Tree then left(V)
```
beurdouche: `V` is just horrible, can you just call it `n`,`ni` or `i` for node, node index or index instead ?

draft-ietf-mls-protocol.md
```diff
@@ -1904,48 +1905,44 @@ tree with the same set of nodes and edges as the epoch's ratchet tree. Each
...
+and right(V) denote the children of V (if they exist).
 
 Each node in the tree is assigned a secret. The root's secret is simply the
 application_secret of that epoch. (See {{key-schedule}} for the definition of
 application_secret.)
 
 ~~~~
-astree_node_[IndexOf(root)]_secret = application_secret
+astree_node_[root]_secret = application_secret
```
beurdouche: This is useless, it is described just above.

draft-ietf-mls-protocol.md
```diff
@@ -1959,54 +1956,36 @@ leaf uses the j-th key/nonce pair in the sequence to encrypt (using the AEAD)
...
+member assigned to the leaf with node index V is simply the secret of
+that leaf.
 
 ~~~~
-struct {
-  opaque gshash<0..2^32-1> = Hash(GroupState_[n]);
-  uint32 leaf_index;
-  uint32 ratchet_position;
-} RatchetContext
+application_[V]_[0]_secret = astree_node_[V]_secret
```
beurdouche: Useless

draft-ietf-mls-protocol.md
```diff
@@ -1959,54 +1956,36 @@ leaf uses the j-th key/nonce pair in the sequence to encrypt (using the AEAD)
...
+ratchet tree is the same as the index of the leaf in the AS Tree
+used to initialize the sender's ratchet.
 
 ~~~~
-application_[i]_[j]_secret
+application_[V]_[j]_secret
       |
-      +--> HKDF-Expand-Label(., "app-nonce", HashRatCont[i,j+1], AEAD.nonceLen)
-      |    = application_[i]_[j+1]_nonce
+      +--> Derive-App-Secret(., "app-nonce", V, j, AEAD.nonce_length)
```
beurdouche: This is not an App-Secret derivation here, I would prefer to stick to a standard HKDF-Expand-Label for these.

draft-ietf-mls-protocol.md
```diff
@@ -2059,13 +2039,15 @@ A0  B0  C0  D0 -+- KD0
             |   +- ND1
             |
             D2 -+- KD2
```
beurdouche: If KD2 and ND2 are on the diagram, I would prefer D3 to be there as well. The reader should see those as derived at the same time by consuming D2.

draft-ietf-mls-protocol.md
beurdouche: But I agree that it make sense for the app-secret to app-secret derivation below.

draft-ietf-mls-protocol.md
bifurcation: Having it called out as an equation, in the notation used below, is useful.

draft-ietf-mls-protocol.md
bifurcation: As above, disagree :)

draft-ietf-mls-protocol.md
bifurcation: Do you think this is harmful, or just un-aesthetic?  The goal here is to preserve Joël's invariant that every derivation has different context.

draft-ietf-mls-protocol.md
bifurcation: As I said in the discussion on #146, this doesn't follow.  Just because you've generated KD2 and ND2 doesn't mean you have to derive forward to D3.  That's only required when you consume one of KD2/ND2.

PR #196, opened 2019-08-08
Cleaned up beginning of Cryptographic Objects section

rozbb:
One of the issues was that the ciphersuites were subsubsections in a section, so it got rendered as
```
6.1) Curve25519, SHA-256, and AES-128-GCM
  6.1.1) P-256, SHA-256, and AES-128-GCM
```
which is odd.

Also, previously, the beginning of the Cryptographic Objects section talked about ciphersuites and public keys. I moved their descriptions to where they are more relevant.

Finally, I replaced "cipher suite" with "ciphersuite", since the latter had 25 occurrences while the former had 2.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -767,11 +767,15 @@ following primitives to be used in group key computations:
 * A Diffie-Hellman finite-field group or elliptic curve
 * An AEAD encryption algorithm {{!RFC5116}}
 
-The ciphersuite must also specify an algorithm `Derive-Key-Pair`
-that maps octet strings with the same length as the output of the
-hash function to key pairs for the asymmetric encryption scheme.
+The ciphersuite's Diffie-Hellman group is used to instantiate an HPKE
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -769,23 +771,9 @@ The ciphersuite must also specify an algorithm `Derive-Key-Pair`
 that maps octet strings with the same length as the output of the
 hash function to key pairs for the asymmetric encryption scheme.
 
-Public keys used in the protocol are opaque values
-in a format defined by the ciphersuite, using the following types:
-
-~~~~~
-opaque HPKEPublicKey<1..2^16-1>;
```
bifurcation: We still need this definition. Maybe put it under the `CipherSuite` enum, as with `SignaturePublicKey` and the `SignatureScheme` enum.

bifurcation: Also, some descriptive text like you added for `SignaturePublicKey` would be helpful.

PR #200, opened 2019-08-19
Non-destructive add

bifurcation:
This PR attempts to implement the "non-destructive add" proposal discussed at IETF 105.  Instead of blanking nodes on add, we keep track of "unmerged leaves" below each intermediate node.  This involves a bit more complexity in the tree state that is passed around, but avoids unnecessary blank nodes.


bifurcation:
As discussed in the interim: Remove processing should be updated such that for each node in the removed node's direct path:

* If the removed leaf is in `unmerged_leaves` for the node, then just remove it from that list
* Otherwise, blank the node

PR #204, opened 2019-08-30
Update ClientInitKey to contain only one ciphersuite.

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -1192,40 +1192,36 @@ The details of application key derivation are described in the
...
-and used in the HPKE construction for TreeKEM.
-
-The whole structure is signed using the client's identity key.
-A ClientInitKey object with an invalid signature field MUST be
-considered malformed.  The input to the signature computation
-comprises all of the fields except for the signature field.
+In order to facilitate asynchronous addition of clients to a group, it is
+possible to pre-publish initialization keys that provide some public information
+about a user.  ClientInitKey messages provide information about a client that
+any existing member can use to add this client to the group asynchronously.
```
bifurcation: Please don't re-wrap text that hasn't been modified.  It makes the history unnecessarily complicated.

draft-ietf-mls-protocol.md
```diff
@@ -1192,40 +1192,36 @@ The details of application key derivation are described in the
...
 
 ~~~~~
 uint8 ProtocolVersion;
 
 struct {
     opaque client_init_key_id<0..255>;
-    ProtocolVersion supported_versions<0..255>;
-    CipherSuite cipher_suites<0..255>;
-    HPKEPublicKey init_keys<1..2^16-1>;
+    ProtocolVersion supported_version;
```
bifurcation: Please move this to the first slot in the struct.  That way we can have a very simple protocol invariant -- the first two octets of the InitKey tell you the format of the rest.  (We should write down these invariants, but that's for a different PR.)

draft-ietf-mls-protocol.md
```diff
@@ -1499,7 +1495,7 @@ struct {
 
 The creator of the group constructs an Init message as follows:
 
-* Fetch a ClientInitKey for each member (including the creator)
+* Fetch ClientInitKeys for each member (including the creator)
```
bifurcation: Maybe say "one or more ClientInitKeys"

draft-ietf-mls-protocol.md
Bren2010: Protocol version is only one octet, is there something else you want here?

bifurcation: Misstatement on my part, sorry.

PR #208, opened 2019-09-16
Add authenticated_data to the mls message.

psla:
[As previously discussed](https://mailarchive.ietf.org/arch/msg/mls/jB5ygAJs3P8TLkduj6Q9vFWmSJI), this is a proposal to add AAD to the application messages.

There are many motivations to do this:
• It's a 'cheap' feature to add (although it could potentially be misused)
• It avoids duplicate content: if there is a content that needs to be authenticated, but also needs to be visible to the server, the only solution today is to repeat it in the header (to the server) and then in the encrypted body.
• Modern ciphers already provide support for AAD, and MLS takes advantage of this. In fact most(all?) of the fields in the header are authenticated already.
• This field is optional(as in: can be empty), which means that it doesn't have to be used in the implementation.

The primary benefit is for the server to have access to the fields that are otherwise authenticated, but are not part of MLS message. Typically, the server has another encryption mechanism with the client (e.g. TLS) and as such client-server communication is already secure. As a matter of fact, handshake messages can already be transported in plaintext (in case server needs to examine their content), but application messages are not allowed to have any plaintext content, even though server may need to examine some metadata as well.

A couple of thoughts that may be worth discussing: 
* what should be the max size of authenticated_data (I assumed 32KB, just like application message)
* whether *authenticated data* should be offered only as part of *application*, or both *handshake message* and *application message*. It is fair to consider it only for an *application message*, though for simplicity I added it to both message types. Happy to change it based on general preferences.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1351,6 +1351,7 @@ struct {
     uint32 epoch;
     ContentType content_type;
     opaque sender_data_nonce<0..255>;
+    opaque authenticated_data<0..2^32-1>;
 } MLSCiphertextSenderDataAAD;
 ~~~~~
 
@@ -1405,8 +1406,8 @@ struct {
     uint32 epoch;
     ContentType content_type;
     opaque sender_data_nonce<0..255>;
-    opaque encrypted_sender_data<0..255>;
     opaque authenticated_data<0..2^32-1>;
+    opaque encrypted_sender_data<0..255>;
 } MLSCiphertextContentAAD;
 ~~~~~
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1291,9 +1291,9 @@ struct {
     opaque group_id<0..255>;
     uint32 epoch;
     ContentType content_type;
+    opaque authenticated_data<0..2^32-1>;
     opaque sender_data_nonce<0..255>;
     opaque encrypted_sender_data<0..255>;
-    opaque authenticated_data<0..2^32-1>;
     opaque ciphertext<0..2^32-1>;
 } MLSCiphertext;
 ~~~~~
@@ -1350,8 +1350,8 @@ struct {
     opaque group_id<0..255>;
     uint32 epoch;
     ContentType content_type;
-    opaque sender_data_nonce<0..255>;
     opaque authenticated_data<0..2^32-1>;
+    opaque sender_data_nonce<0..255>;
 } MLSCiphertextSenderDataAAD;
 ~~~~~
 
@@ -1405,8 +1405,8 @@ struct {
     opaque group_id<0..255>;
     uint32 epoch;
     ContentType content_type;
-    opaque sender_data_nonce<0..255>;
     opaque authenticated_data<0..2^32-1>;
+    opaque sender_data_nonce<0..255>;
     opaque encrypted_sender_data<0..255>;
 } MLSCiphertextContentAAD;
 ~~~~~
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1263,6 +1263,7 @@ struct {
     uint32 epoch;
     uint32 sender;
     ContentType content_type;
+    opaque authenticated_data<0..2^32-1>;
 
     select (MLSPlaintext.content_type) {
         case handshake:
@@ -1273,7 +1274,6 @@ struct {
             opaque application_data<0..2^32-1>;
     }
 
-    opaque authenticated_data<0..2^32-1>;
     opaque signature<0..2^16-1>;
 } MLSPlaintext;
 
@@ -1390,7 +1390,7 @@ struct {
     ContentType content_type;
     opaque sender_data_nonce<0..255>;
     opaque encrypted_sender_data<0..255>;
-    opaque authenticated_content[length_of_authenticated_content];
+    opaque authenticated_data<0..2^32-1>;
 } MLSCiphertextContentAAD;
 ~~~~~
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1273,6 +1273,7 @@ struct {
             opaque application_data<0..2^32-1>;
     }
 
+    opaque authenticated_data<0..2^32-1>;
```
bifurcation: Nit: I would put this just below the content_type field, just to preserve parallelism (`MLSCiphertext.ciphertext -> MLSPlaintext.operation/application_data`)

eomara:
@bifurcation per our discussion in London, we agreed to accept this change. you mentioned other places need to be updated,  which one?

draft-ietf-mls-protocol.md
```diff
@@ -1388,6 +1390,7 @@ struct {
     ContentType content_type;
     opaque sender_data_nonce<0..255>;
     opaque encrypted_sender_data<0..255>;
+    opaque authenticated_content[length_of_authenticated_content];
```
beurdouche: `opaque` already encompass the length, I believe.

draft-ietf-mls-protocol.md
beurdouche: Also it would avoid interleaving plaintexts and ciphertexts...

draft-ietf-mls-protocol.md
psla: I was inspired by
```
    opaque content[length_of_content];
```
above (in MLSCiphertextContent). Is that one also incorrect?

psla:
> this AAD field is AEAD authenticated but does not seem covered by the signature, while probably should.
I don't quite understand how to address this comment. I believe I added it to the signature, maybe I missed something

> ** Sign the plaintext metadata -- the group ID, epoch, sender index, and  content type **as well as the authenticated data** and message content

Can you clarify which part is missing the AAD?

draft-ietf-mls-protocol.md
bifurcation: This can just be the same definition as in the `MLSCiphertext` object.  The only reason for the `[]` notation w.r.t. content is the weird encoding of `MLSCiphertextContent`, which IIRC is about to get reverted.

bifurcation:
@beurdouche i think you're wrong here.  In {{content-signing-and-encryption}}, we have the following:

>  The signature covers the plaintext metadata and message content, i.e., all fields of MLSPlaintext except for the signature field.

I think this is ready to go as soon as @psla fixes the two minor comments we have (order of fields and `[]` vs `<>` notation).

draft-ietf-mls-protocol.md
psla: I agree. Done!

draft-ietf-mls-protocol.md
psla: Done.
I also renamed it to authenticated_data instead of authenticated_content, in order to keep the naming the same. (I was in the pickle here, since the data struct is called 'MLSCiphertextConterntAAD' and then the field inside is called authenticated_data. I initially thought that calling it authenticated_content will be better, but I believe it's better to keep the same name all over the place).

psla:
Are you saying that `MLSCiphertextSenderDataAAD` needs it as well? That would lead to two AAD blobs, right? One of which would be encrypted? (I am a bit fuzzy about the relation between `MLSCiphertextContentAAD` and encrypted_sender_data (which is encrypted `MLSCiphertextSenderDataAAD`, right?)

I added it there, but I don't fully understand it. Please take another look.

bifurcation:
@beurdouche  - I agree with @psla that it doesn't seem necessary to include the `authenticated_data` in the `MLSCiphertextSenderDataAAD`.  Do you have a specific issue in mind?

beurdouche:
Ok, I am quite exhausted this week, so my clarity is low sorry…
I think I confused the positionning of the new field in the MLSPlaintext vs the MLSCiphertext

It all depends where you place the new field in the MLSCiphertext, right ?!
In the current design for the AEADs the thing you authenticate is the prefix of
the ciphertext you are decrypting...

So if the new field is just after the content type in the MLSCiphertext
it is part of the prefix of the encrypted sender data and has to be included in the SenderDataAAD but
it is also in the prefix of the content so it has to also be in the ContentAAD.

If the new field is placed between the encrypted SenderData and the encrypted Content
it is just a prefix of the second one and should appear only in the ContentAAD.

Do I make sense ? :)
B.

bifurcation:
If we espouse that theory (not clear that it necessarily holds, but let's go with it for now), the question is where you draw the line between metadata and content.  Compare the current state with two cases (a) and (b):

```
NOW              A.               B.

group_id         group_id         group_id
epoch            epoch            epoch
sender           sender           sender
content_type     content_type     content_type
============     ============     aad
content          aad              ============
signature        content          content
                 signature        signature
```

Basically, the `=====` turns into the sender_data in the MLSCiphertext.  In case (B), you would want the AAD in the `SenderDataAAD`; in case (A) you would not.  I don't really see a reason why we need to assume case (B), so since case (A) is a bit simpler, I'm inclined to go that way.  Simpler in the sense that the AAD is used as AAD exactly once, together with the protected content.

beurdouche:
Those were my proposals…
My preference goes to Fig 8 but I don’t have a strong opinion.

B.

bifurcation:
Ah, even better illustration than mine!  OK, when you put it that way, I can see the appeal of including the AAD in the sender data.  Though I admit this is mostly an aesthetic point, not one that I have any security analysis to back up.

@psla would you mind moving the AAD to above the `sender_data_nonce` in all cases?  Then I think this is ready to merge.

psla:
It makes sense to me too now. Thanks. I think I addressed it.

PR #209, opened 2019-09-18
Proposals and Commits

bifurcation:
This PR implements the proposal discussed in Prague and on the mailing list, to separate GroupOperations into Proposals and Commits.  This change leaves the underlying ratchet tree operations untouched, but changes how these operations are folded into the overall key schedule.

This is also a fairly initial effort, with several OPEN ISSUEs noted.  The most important IMO is that we should allow a single MLSPlaintext to pack one or more Proposals along with a Commit, which makes it simpler to emulate the current behavior, and might allow us to get rid of the Init message.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -339,7 +339,7 @@ There are three major operations in the lifecycle of a group:
 Each of these operations is "proposed" by sending a message of the corresponding
 type (Add / Update / Remove).  The state of the group is not changed, however,
 until a Commit message is sent to provide the group with fresh entropy.  In this
-section, we show each proposal being committed immediately, but in more advance
+section, we show each proposal being committed immediately, but in more advanced
 deployment cases, an application might gather several proposals before
 committing them all at once.
 
@@ -808,8 +808,6 @@ enum {
 opaque HPKEPublicKey<1..2^16-1>;
 ~~~~~
 
-## Ciphersuites
-
 ### Curve25519, SHA-256, and AES-128-GCM
 
 This ciphersuite uses the following primitives:
@@ -1045,20 +1043,20 @@ struct {
 
 confirmed_transcript_hash_[n] =
     Hash(interim_transcript_hash_[n-1] ||
-         MLSPlaintextOpContent_[n]);
+         MLSPlaintextCommitContent_[n]);
 
 interim_transcript_hash_[n] =
     Hash(confirmed_transcript_hash_[n] ||
-         MLSPlaintextOpAuthData_[n]);
+         MLSPlaintextCommitAuthData_[n]);
 ~~~~~
 
 Thus the `confirmed_transcript_hash` field in a GroupContext object represents a
-transcript over the whole transcript of MLSPlaintext Commit messages, up to the
+transcript over the whole history of MLSPlaintext Commit messages, up to the
 confirmation field in the current MLSPlaintext message.  The confirmation and
 signature fields are then included in the transcript for the next epoch.  The
 interim transcript hash is passed to new members in the WelcomeInfo struct, and
 enables existing members to incorporate a handshake message into the transcript
-without having to store the whole MLSPlaintext structure.
+without having to store the whole MLSPlaintextCommitAuthData structure.
 
 When a new group is created, the `interim_transcript_hash` field is set to the
 zero-length octet string.
@@ -1443,8 +1441,6 @@ an Init message and sending it to those members.
 ~~~~~
 struct {
   opaque group_id<0..255>;
-  ProtocolVersion version;
-  CipherSuite cipher_suite;
   ClientInitKey members<0..2^32-1>;
   DirectPath path;
 } Init;
@@ -1464,7 +1460,9 @@ The creator of the group constructs an Init message as follows:
 Each member of the newly-created group initializes its state from
 the Init message as follows:
 
-* Note the group ID, protocol version, and ciphersuite in use
+* Verify that all of the ClientInitKeys listed are for the same protocol and
+  ciphersuite; if not, reject the Init as malformed
+* Note the group ID, protocol version, and ciphersuite in use.
 * Construct a ratchet tree as above
 * Update the cached ratchet tree by replacing nodes in the direct
   path from the first leaf using the direct path
@@ -1542,6 +1540,11 @@ struct {
 } Proposal;
 ~~~~~
 
+On receiving a MLSPlaintext containing a Proposal, a client MUST verify the
+signature on the enclosing MLSPlaintext.  If the signature verifies
+successfully, then the Proposal should be cached in such a way that it can be
+retrieved using a ProposalID in a later Commit message.
+
 ### Add
 
 An Add proposal requests that a client with a specified ClientInitKey be added
@@ -1565,10 +1568,9 @@ in the Commit message by taking the following steps:
 
 * Blank the path from the leaf at position `index` to the root
 
-* Set the leaf node in the tree at position `index` to a new node
-  containing the public key from the ClientInitKey in the Add
-  corresponding to the ciphersuite in use, as well as the
-  credential under which the ClientInitKey was signed
+* Set the leaf node in the tree at position `index` to a new node containing the
+  public key from the ClientInitKey in the Add, as well as the credential under
+  which the ClientInitKey was signed
 
 ### Update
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1522,218 +1464,274 @@ the Init message as follows:
...
+    ClientInitKey init_key;
+} Add;
 ~~~~~
 
-In the description of the tree as a list of nodes, the `credential`
-field for a node MUST be populated if and only if that node is a
-leaf in the tree.
+The proposer of the Add does not control where in the group's ratchet tree the
+new member is added.  Instead, the sender of the Commit message chooses a
+location for each added member and states it in the Commit message.
```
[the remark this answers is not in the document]
bifurcation: That is the desire, but it runs into the problem that "first" isn't well-defined in a context where members might see two Add messages in different orders.  So the idea here is that the Committer acts as the arbiter and decides the order.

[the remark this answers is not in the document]
Bren2010: Doesn't the Commit message define an order over the Adds? You could
```diff
- AddCommit adds<0..2^16-1>;
+ ProposalID adds<0..2^16-1>;
```
because the index is implicit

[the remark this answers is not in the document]
bifurcation: Oh, I see. Good point.  If I try hard, I can imagine scenarios where you would not want that ("Each user gets 8 devices, so we're going to assign each user a slice of the tree"), but for this first cut, I'm happy to have it simpler.  We can add the flexibility back if people ask for it.

draft-ietf-mls-protocol.md
```diff
@@ -1190,6 +1206,10 @@ A step in this chain (the second subscript) is called a "generation".
 The details of application key derivation are described in the
 {{astree}} section below.
 
+[[ OPEN ISSUE: With the addition of Proposals, handshake encryption is now
+broken.  We need a framework where each sender has multiple keys, so we should
+probably just generate another AStree for handshake encryption. ]]
```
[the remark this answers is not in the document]
bifurcation: It's a question of where you separate the handshake keys from the application keys.  You want them to be separate to simplify the analysis, so that the implication always goes Handshake => Application and not vice versa.

In the current draft, the nearest common ancestor is the epoch secret.  I think you're on the right track here, and a better design would be to use the AStree to fan out secrets per participant, then split out different keys for different uses at the leaves.

@beurdouche should probably opine here, but in any case, it seems like a follow-on PR, not something to address here.

draft-ietf-mls-protocol.md
```diff
@@ -1524,223 +1460,285 @@ The creator of the group constructs an Init message as follows:
...
 
-An Update message is sent by a group member to update its leaf
-secret and key pair.  This operation provides post-compromise security
-with regard to the member's prior leaf private key.
+A Commit message initiates a new epoch for the group, based on a collection of
+proposals.  It instructs group members to update their representation of the
+state of the group by applying the proposals and advancing the key schedule.
+
+A group member that has observed one or more Proposal messages within an epoch
+MUST send a Commit message before sending application data.  This ensures, for
```
Bren2010: Should implementations be required to incorporate all valid proposals that they know of? If not, meaning proposals have a chance of being dropped, how are they meant to be communicated to an end-user?

draft-ietf-mls-protocol.md
```diff
@@ -1524,223 +1460,285 @@ The creator of the group constructs an Init message as follows:
...
+example, that any members whose removal was proposed during the epoch are
+actually removed before any application information is transmitted.
```
Bren2010: Because proposals can't build on top of each other, you have weird cases like:

1. Alice adds Bob to the group (without committing).
2. Carol removes Bob from the group.

It implicitly requires Carol to Commit->Remove because she can't form her Remove proposal without first committing. Whatever API you expose to an application would have to understand this

draft-ietf-mls-protocol.md
```diff
@@ -791,8 +808,6 @@ enum {
 opaque HPKEPublicKey<1..2^16-1>;
 ~~~~~
 
-## Ciphersuites
-
 ### Curve25519, SHA-256, and AES-128-GCM
```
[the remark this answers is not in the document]
bifurcation: It's still there, it just got moved up in #196.

draft-ietf-mls-protocol.md
```diff
@@ -1524,223 +1460,285 @@ The creator of the group constructs an Init message as follows:
...
-are revealed to the new member.
+* Blank the path from the leaf at position `index` to the root
 
-Since the new member is expected to process the Add message for
-itself, the Welcome message should reflect the state of the group
-before the new user is added. The sender of the Welcome message can
-simply copy all fields from their GroupContext object.
+* Set the leaf node in the tree at position `index` to a new node containing the
+  public key from the ClientInitKey in the Add, as well as the credential under
+  which the ClientInitKey was signed
```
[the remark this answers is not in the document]
bifurcation: Could you say more?  I already updated this to remove the reference to multiple ciphersuites in the CIK.

draft-ietf-mls-protocol.md
bifurcation: This being a distributed system, there's always a risk of proposals being dropped.  One of the drawbacks of this proposal is that it increases the complexity of the retry logic.  Nonetheless, we can require that the Commit contain all proposals **known to the committer**.

draft-ietf-mls-protocol.md
bifurcation: This seems workable to me.  I would expect that the API just wouldn't mark Bob as added until the Add was committed, so Carol wouldn't even be able to form the Remove message until then.

draft-ietf-mls-protocol.md
```diff
@@ -1524,223 +1460,294 @@ The creator of the group constructs an Init message as follows:
...
+} Welcome;
+~~~~~
+
+In the description of the tree as a list of nodes, the `credential`
+field for a node MUST be populated if and only if that node is a
+leaf in the tree.
 
+[[ OPEN ISSUE: If multiple members are being added, it would be nice to be able
+to encrypt the group state once and just encrypt the key to the members.  In
+that case, you would also need to factor out the "index" field. ]]
```
Bren2010: There's a section below here, "Merging Updates", that I think you can remove. Merging multiple Commits doesn't make sense

draft-ietf-mls-protocol.md
```diff
@@ -1524,223 +1460,294 @@ The creator of the group constructs an Init message as follows:
...
+   public key from the credential stored at the leaf in the tree indicated by
+   the `sender` field.
+
+3. Generate a provisional GroupContext object by applying the proposals
+   referenced in the commit object in the order provided, as described in
+   {{proposals}}.  Add proposals are applied left to right: Each Add proposal is
+   applied at the leftmost unoccupied leaf, or appended to the right edge of the
+   tree if all leaves are occupied.
+
+4. Process the `update_path` value to update the ratchet tree referenced by the
```
rozbb: Typo: `path`

Also, if there's only a single DirectPath per epoch, won't this lead to trees becoming sparser (i.e., having fewer filled nodes) over time? It's fine as long as the root node is filled, but DirectPath messages might have to get bigger to accommodate the sparseness.

Bren2010: The sparseness of the tree depends on how often all the members Commit. For example, if you have a group where only one person ever Commits, then the direct path above that person is the only one that will ever be populated, and MLS will degrade to "encrypt every message to every person individually" scheme.

If you have an application-level understanding of who needs to Commit and who's allowed to be lazy, it could be fine(?). It gives you a performance trade-off.

Bren2010:
Note: Need to stop truncating hash of proposal

draft-ietf-mls-protocol.md
```diff
@@ -1547,223 +1483,296 @@ The creator of the group constructs an Init message as follows:
...
 would need to figure out what mitigations would be necessary. ]]
 
-## Add
+# Group Evolution
+
+Over the lifetime of a group, its membership can change, and existing members
+might want to change their keys in order to achieve post-compromise security.
+In MLS, each such change is accomplished by a two-step process:
+
+1. A proposal to make the change is broadcast to the group in a Proposal message
```
beurdouche: s/broadcast/broadcasted

draft-ietf-mls-protocol.md
beurdouche: Yes. As we need to indicate all the proposal accepted and rejected by the committer so that other members know what proposals have been considered.

draft-ietf-mls-protocol.md
beurdouche: Richard is correct here.

beurdouche:
This is good. Thanks Richard !

PR #213, opened 2019-09-30
Change ciphertext content to be standard TLS format.

Bren2010:
I don't like the pattern of having a marker byte followed by zero-padding. It has to be parsed manually, backwards, unlike every other object in the protocol which all follow a common format.


bifurcation:
Agreement at the interim was that people don't care about the overhead (for now), so we're going to go with the simpler, TLS-syntax-based structure.

PR #218, opened 2019-10-01
Favor not reusing ClientInitKey reuse (SHOULD)

beurdouche:
This has been reported unclear multiple times. I strongly prefer forbidding reuse altogether.
A validation mechanism is needed for the formal security proof, it can be implemented according to an expiration mechanism (see Architecture - 4.2.  Delivery Service Compromise)

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1213,8 +1213,8 @@ A ClientInitKey object specifies a ciphersuite that the client
 supports, as well as providing a public key that others can use
 for key agreement. The client's identity key is intended to be
 stable throughout the lifetime of the group; there is no mechanism to
-change it.  Init keys are intended to be used only once and SHOULD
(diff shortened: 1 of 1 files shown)


Bren2010:
We need to allow Last Resort keys. Not sure what you mean by "replay attacks"

beurdouche:
> We need to allow Last Resort keys. Not sure what you mean by "replay attacks"

I rewrote it in favor of using "SHOULD NOT reuse", which is true, except in case of last resort to avoid DOS, which we are willing to allow.

Normal meaning... If I remember correctly, there is a cross channel forwarding attack (replay) possible on the first message in Signal when reusing prekeys (the last resort key) because it lacks some channel binding but I would have to check. I am working to make sure we don't suffer the same defect but I am documenting it in the mean time.

bifurcation:
I can live with this for now; we can refine later.

Issue #160, opened 2019-05-11
Advertize a global app generation for a sender


beurdouche:
We could use the sender_generation field of the MLS message to signal what was the global app generation of the sender.  Functionally, this allows receivers to know if they received all previous application messages from the sender. In the current design, if S sent M0, M1, M2 you can know if you missed M1 when receiving M2, but you cannot know if you missed M2. There are many ways of doing this: global sender counter, counter between two operations of the sender...

beurdouche:
We seemed in agreement at the interim because it prevents the application message suppression attacks

PR #219, opened 2019-10-01
Fix application message truncation attack

beurdouche:
2020/03/04 - Will be resolved by the next interim

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1357,6 +1357,27 @@ recipient MUST verify that the sender field represents an occupied
 leaf in the ratchet tree.  In particular, the sender index value
 MUST be less than the number of leaves in the tree.
 
+The sender's generation counter not only allows the recipient
+to choose the correct decryption key, it also prevents application
+message truncation attacks. In the case of an handshake message
+the generation MUST be increased by one as if it were an application
+message, before reset to zero for the next epoch.
+
+Upon receipt of a Handshake message, members SHOULD check that they
+received and decrypted all application messages advertized by the
+sender and SHOULD delete keys associated with a generation equal or
+higher to that number to spare storage space.
+
+This mechanism allows the recipients to know if application
+messages have been lost in transit or suppressed by an active
+adversary via an application message truncation attack without
+the need for an explicit acknowledgment.
+
+Note that the generation is reset to zero only for the sender
+updating its contribution to the group secrets via a handshake
+message.
+
+
 ## Content Signing and Encryption
 
 The signature field in an MLSPlaintext object is computed using the


Bren2010:
I don't agree with this change. It's not MLS's job to enforce an order on messages, or enforce that messages don't get lost. If you want to ensure things, you can build your own mechanism and store it in the AAD of the messages. This also lets people build their own thing that's as strong/weak as they need:

- Ensure per-peer message order, no messages lost: Each peer puts a counter in their AAD, "This is my 7th message." and others can complain if they haven't seen message 6.
- Ensure total message order: Each message has the hash of the last message in its AAD.
- Maybe you have no mechanism because you're fine with messages getting lost.

katrielalex:
I'd be happy to include in the AAD the length of the previous chain ("PN" in Signal's language), but I don't have strong feelings.

beurdouche:
> I don't agree with this change. It's not MLS's job to enforce an order on messages,
> 
Nobody is talking about this...
> or enforce that messages don't get lost.
> 
Modern protocols gives you this property. I believe Signal does and minimally, you should know if I drop all your messages.
> If you want to ensure things, you can build your own mechanism and store it in the AAD of the messages. This also lets people build their own thing that's as strong/weak as they need:
> 
> Ensure per-peer message order, no messages lost: Each peer puts a counter in their AAD, "This is my 7th message." and others can complain if they haven't seen message 6.
> Ensure total message order: Each message has the hash of the last message in its AAD.
> Maybe you have no mechanism because you're fine with messages getting lost
This reasonning is completly the reverse of the MLS goals, we actually want to enforce as much
as possible in the protocol and relax at the application level but definitely not let people design complex
things at the application level without any information callback from the protocol layer

Bren2010:
> This reasonning is completly the reverse of the MLS goals, we actually want to enforce as much
as possible in the protocol

We leave a lot of things out of the protocol that are better handled at another layer, or where there are multiple acceptable solutions. "Application message truncation attack" is not an attack, it's just another way to say "packet loss" which is better handled at the TCP / application layer, because those are the layers that know how things are going to get lost.

TLS, for example, doesn't know how to recover lost packets -- that's TCP's job. TLS also doesn't know how to do application-level retries -- that's the application's job.

raphaelrobert:
I think that including the counter is a meaningful mechanism to ensure the DS behaves correctly. If possible at all, this should not be solved at the application layer. Naturally applications can ignore lost/suppressed messages if they so wish.
Transcript consistency was one the properties discussed early on and this goes in the right direction at a very low cost.

bifurcation:
This PR has gotten quite stale and out of sync with the draft.  Closing for now.  @beurdouche holds the token to propose an alternate resolution to #160.

PR #221, opened 2019-10-01
Precisions on authentication and IdentityKeys

beurdouche:
I couldn't find this written anywhere else and this is also related to a discussion about PCS with Cas and al.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -916,6 +916,19 @@ enum {
 opaque SignaturePublicKey<1..2^16-1>;
 ~~~~~
 
+Note that each new credential that has not already been validated
+by the application SHOULD be validated against the Authentication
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -916,6 +916,19 @@ enum {
 opaque SignaturePublicKey<1..2^16-1>;
 ~~~~~
 
+Note that each new credential that has not already been validated
+by the application SHOULD be validated against the Authentication
+Service.
```
Bren2010: maybe just "MUST be validated."

bifurcation: I don't think you can make this a MUST.  You can imagine applications that don't really care about authentication, but want all the other properties.  Also, there is no comparable requirement in TLS.

PR #233, opened 2019-10-09
Initialize new members directly into the current epoch

### LINKED ISSUE #223: Welcome should not leak previous epoch encryption keys
Instead of initializing new members with information from the previous epoch and requiring them to process an Add, we should initialize new members directly with current epoch information.  This is covered in #209, but should be factored out.

bifurcation:
Fixes #223 

Discussion of security analysis in the issue.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1142,20 +1142,20 @@ proceeds as shown in the following diagram:
                      V
 update_secret -> HKDF-Extract = epoch_secret
                      |
-                     +--> Derive-Secret(., "sender data", GroupContext_[n])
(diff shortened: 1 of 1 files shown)


raphaelrobert:
Shouldn't we rebase this on the Proposals proposal? Otherwise we might just throw away a good part of this PR soon.

bifurcation:
@raphaelrobert - Since there wasn't clear agreement on the Proposals PR, I went ahead and pulled this out.  If we are going to do Proposals, it won't be a bad rebase.  And in any case, doing this on its own made it a bit easier to do the supporting analysis.

bifurcation:
Closing this, as it looks like #209 is going to land soon.

PR #234, opened 2019-10-09
Add downgrade protection to ClientInitKeys

### LINKED ISSUE #224: Add extensions for ClientInitKeys
To address the risk of version / ciphersuite downgrade by the CIK server, CIKs need to express the lists of versions and ciphersuites that a client supports.

* Add a general extensions field to ClientInitKey, in the same form as TLS ClientHello/ServerHello
* Define a `supported_versions` extension
* Define a `supported_ciphersuites` extension
* Require both extensions to be present

This allows the creator of a group to verify that CIKs it gets do not omit versions that the other client actually supports.  On Init, the participants of the group can examine the CIKs in the group to verify that the proper version / ciphersuite have been selected.

bifurcation:
Fixes #224

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -147,6 +147,8 @@ draft-08
 
 - Replace Init messages with multi-recipient Welcome message (\*)
 
+- Add extensions to ClientInitKeys for expiration and downgrade resistance (\*)
+
 draft-07
 
 - Initial version of the Tree based Application Key Schedule (\*)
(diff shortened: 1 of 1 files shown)


raphaelrobert:
I think @bifurcation added what we discussed at the interim, but I agree with @beurdouche that we shouldn't label it as "protection" because of the lack of guarantees.
It is indeed just an information mechanism, but a number of questions are left open:
 - what should the client creating a group do if better ciphersuites /higher versions are advertised but the DS doesn't deliver the corresponding keys?
 - what should an added client do when it realises it was added with a sub-optimal ClientInitKey?
 - ...

bifurcation:
This now has a MUST for using the extensions to prevent downgrade.

PR #192, opened 2019-07-19
Made it clear that Init messages MUST be sent unencrypted

rozbb:
It took me a few reads to gather that that was the purpose of Init messages. The OPEN ISSUE actually cleared it up for me.

PR #239, opened 2019-11-11
Welcome* == Init

bifurcation:
This PR makes two major changes that are enabled by Proposals:

* Multiple participants added in the same epoch receive the same Welcome message
* Init is always a special case of Add

Prior to Proposals, treating Init as a special case of Add led to a performance degradation (`n -> n log n`).  With Proposals, any number of Adds can be done with a single Commit (at `O(n)` cost), so there is no more degradation.  The result seems (IMO) much more elegant than before.

The main cost is that encryption of Welcome messages is now in two stages (three if you include the internal HPKE stage!), even if there is only one new member.

Depends on #209 
Obsoletes #192 and #194

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1510,16 +1510,17 @@ In principle, the above process could be streamlined by having the creator
 directly create a tree and choose a random value for first epoch's epoch secret.
 We follow the steps above because it removes unnecessary choices, by which, for
 example, bad randomness could be introduced.  The only choices the creator makes
-here are its own HPKE key and credential, and the leaf secret from which the
-Commit is built.
+here are its own HPKE key and credential, the leaf secret from which the
+Commit is built, and the intermediate key pairs along the direct path to the
+root.
 
 A new member receiving a Welcome message can recognize group creation if the
-signer of the Welcome message is at index 0, and the number of entries in the
-`members` array is equal to the number of leaves in the tree minus one.  A
-client receiving a Welcome message SHOULD verify whether it is a newly created
-group, and if so, SHOULD verify that the above process was followed by
-reconstructing the Add and Commit messages and verifying that the resulting
-transcript hashes and epoch secret match those found in the Welcome message.
+number of entries in the `members` array is equal to the number of leaves in the
+tree minus one.  A client receiving a Welcome message SHOULD verify whether it
+is a newly created group, and if so, SHOULD verify that the above process was
+followed by reconstructing the Add and Commit messages and verifying that the
+resulting transcript hashes and epoch secret match those found in the Welcome
+message.
 
 # Group Evolution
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1465,69 +1467,59 @@ The ciphertext field of the MLSCiphertext object is produced by
...
+
+* Transmit the Welcome message to the other new members
+
+The recipient of a Welcome message processes it as described in
+{{welcoming-new-members}}.
+
+In principle, the above process could be streamlined by having the creator
+directly create a tree and choose a random value for first epoch's epoch secret.
+We follow the steps above because it removes unnecessary choices, by which, for
+example, bad randomness could be introduced.  The only choices the creator makes
```
beurdouche: The creator chooses all CIKs and all secrets from its leaf to the root.

draft-ietf-mls-protocol.md
```diff
@@ -1465,69 +1467,59 @@ The ciphertext field of the MLSCiphertext object is produced by
...
+here are its own HPKE key and credential, and the leaf secret from which the
+Commit is built.
+
+A new member receiving a Welcome message can recognize group creation if the
+signer of the Welcome message is at index 0, and the number of entries in the
```
beurdouche: I would prefer avoiding the requirement that the creator is at index 0 because it breaks privacy of its identity.
When newcomers are included later on, it might be nice to not know immediately who is the creator.

draft-ietf-mls-protocol.md
```diff
@@ -1751,16 +1743,17 @@ tree. ]]
...
+members with the current state of the group, after the application of the Commit
+message.  The new members will not be able to decrypt or verify the Commit
+message, but will have the secrets they need to participate in the epoch
 initiated by the Commit message.
 
-The information in a Welcome message is encrypted for the new member using HPKE.
-The recipient key pair for the HPKE encryption is the one included in the
-indicated ClientInitKey.
+In order to allow the same Welcome message to be sent to all new members,
+information describing the group is encrypted with a symmetric key and nonce
```
beurdouche: randomly chosen key and nonce

draft-ietf-mls-protocol.md
```diff
@@ -1769,28 +1762,69 @@ struct {
...
+  uint32 signer_index;
+  opaque signature<0..255>;
+} GroupInfo;
 
 struct {
-    opaque client_init_key_id<0..255>;
-    HPKECiphertext encrypted_welcome_info;
+  opaque group_info_key<1..255>;
+  opaque group_info_nonce<1..255>;
+} MemberInfo;
```
beurdouche: MemberInfo is a bad name, welcome keypackage or something ?

draft-ietf-mls-protocol.md
```diff
@@ -1769,28 +1762,69 @@ struct {
...
+
+struct {
+  opaque client_init_key_hash<1..255>;
+  HPKECiphertext encrypted_member_info;
+} EncryptedMemberInfo;
+
+struct {
+  ProtocolVersion version = mls10;
+  CipherSuite cipher_suite;
+  EncryptedMemberInfo members<1..V>;
```
beurdouche: Same thing, for the name: `KeyPackage packages` or something ?

draft-ietf-mls-protocol.md
```diff
@@ -1769,28 +1762,69 @@ struct {
...
-leaf in the tree.
+leaf in the tree (i.e., a node with an even index).
+
+On receiving a Welcome message, a client processes it using the following steps:
+
+* Identify an entry in the `members` array where the `client_init_key_hash`
+  value corresponds to one of this client's ClientInitKeys, using the hash
+  indicated by the `cipher_suite` field.  If no such field exists, or if the
+  ciphersuite indicated in the  ClientInitKey does not match the one in the
+  Welcome message, return an error.
```
beurdouche: What kind of error ?

PR #242, opened 2019-11-12
Pack multiple Proposals and a Commit together

bifurcation:
This is a light optimization over #209, removing the overhead of extra encryptions/signatures in the case where proposal(s) and a commit are sent at the same time.

Depends on #209


beurdouche:
Unfortunately, using a single content type enforces the DS to order everything while there is no reason to force ordering of proposals made for the same epoch.

bifurcation:
On conversation with @beurdouche - Will keep content types separate, but allow Commit to have proposals.

PR #245, opened 2019-11-14
Unpredictable epochs

bifurcation:
This allows for forks in the group history.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1059,8 +1059,9 @@ These effects are described in their respective subsections of {{proposals}}.
 The following general rules apply:
 
 * The `group_id` field is constant
-* The `epoch` field increments by one for each Commit message that
-  is processed
+* The `epoch` field is set to the first four octets of the hash of  the
+  MLSPlaintext carrying the relevant commit, interpreted as an unsigned integer
+  in network byte order
 * The `tree_hash` is updated to represent the current tree and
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1059,8 +1059,9 @@ These effects are described in their respective subsections of {{proposals}}.
 The following general rules apply:
 
 * The `group_id` field is constant
-* The `epoch` field increments by one for each Commit message that
-  is processed
+* The `epoch` field is set to the first four octets of the hash of  the
```
Bren2010: - Stop truncating hashes!
- What about the first epoch?

beurdouche: I am more concern about loosing lexicographic ordering here. The DS wouldn't know how to reorder.

raphaelrobert:
I'm not sure I understand how the DS can do its job with this. To re-iterate what that is:

 - The DS needs to be the arbitrator in a race condition when two clients send a HS message based on the same previous epoch
 - In order to do so, it needs to be able to compare the new with the old epoch
 - It should additionally be able to detect when a HS message is not based on the previous and reject it

The above obviously works nicely with an incremental counter for epochs. I'd be happy if we can further obfuscate the epoch number, I'm just not sure what options we have that satisfy the above requirements.

bifurcation:
Obsoleted by #281

PR #246, opened 2019-11-16
Bugfixes in ClientInitKey, Commit, and Welcome

bifurcation:
1. The `client_init_key_id` is no longer needed, since CIKs are referenced by hash.
2. ProposalIDs were no longer usable to refer to proposals since we enabled multiple proposals per MLSPlaintext
3. The syntax description for Welcome was invalid

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1314,10 +1314,9 @@ struct {
 } Extension;
 
 struct {
-    ProtocolVersion supported_version;
-    opaque client_init_key_id<0..255>;
+    ProtocolVersion version;
     CipherSuite cipher_suite;
-    HPKEPublicKey init_key;
+    HPKEPublicKey leaf_key;
     Credential credential;
     Extension extensions<0..2^16-1>;
     opaque signature<0..2^16-1>;
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1072,7 +1072,6 @@ struct {
   uint32 epoch;
   uint32 sender;
   ContentType content_type = commit;
-  Proposal proposals<0..2^32-1>;
   Commit commit;
 } MLSPlaintextCommitContent;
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1314,10 +1314,9 @@ struct {
 } Extension;
 
 struct {
-    ProtocolVersion supported_version;
-    opaque client_init_key_id<0..255>;
```
beurdouche: I am quite sure that it is needed for lookup, the emitter should have the ability to validate it and you want to include it in the signature.

draft-ietf-mls-protocol.md
```diff
@@ -1314,10 +1314,9 @@ struct {
...
+    ProtocolVersion version;
     CipherSuite cipher_suite;
-    HPKEPublicKey init_key;
+    HPKEPublicKey leaf_key;
```
beurdouche: I prefer `init_key` which conveys the fact that it is the very first KEM public key

draft-ietf-mls-protocol.md
bifurcation: Why do you need it for lookup?  The only place it's referenced is Welcome, which references it by hash.

draft-ietf-mls-protocol.md
bifurcation: OK, sure, whatever :)

draft-ietf-mls-protocol.md
bifurcation: Why do you need it for lookup?  The only place it's referenced is Welcome, which references it by hash.

beurdouche: It might be fine to use the hash as an implicit identifier for outside the protocol for now but I suspect that in some applications we'll want the application to pick the identifier.

bifurcation:
Merge with master obsoleted some changes.  I re-implemented the "supported_version"->"version" change, and added a "key_id" extension to address Benjamin's comment.

PR #247, opened 2019-11-16
Welcome confirmation and key derivation

bifurcation:
1. Once a new member instantiates the group state, they can verify the confirmation MAC that was sent in the Commit that created the state.  This lets new members verify that they've joined correctly.  (It does not assure that they're seeing the same confirmation as anyone else.)

2. Instead of generating a fresh key, we can simply provide the epoch key in the key package and derive the GroupInfo encryption key from it.  This avoids relying on fresh entropy, and makes it more difficult for the sender to provide different keys that will all decrypt and authenticate the GroupInfo.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1195,6 +1195,9 @@ proceeds as shown in the following diagram:
                      V
 update_secret -> HKDF-Extract = epoch_secret
                      |
+                     +--> HKDF-Expand(., "mls 1.0 welcome", Hash.length)
+                     |    = welcome_secret
+                     |
                      +--> Derive-Secret(., "sender data", GroupContext_[n])
                      |    = sender_data_secret
                      |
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1903,18 +1906,15 @@ struct {
   uint32 epoch;
   optional<RatchetNode> tree<1..2^32-1>;
   opaque confirmed_transcript_hash<0..255>;
-
-  // Inputs to the next round of the key schedule
   opaque interim_transcript_hash<0..255>;
-  opaque epoch_secret<0..255>;
 
+  opaque confirmation<0..255>;
```
beurdouche: I am trusting you that this doesn't introduce another cyclic dependency... :)

bifurcation:
This appears to have been rendered obsolete by #265

beurdouche:
Actually this is the nice one providing the nice properties, unlike #265 which breaks the cryptographic properties we expect from the secrecy invariant (see #277)

bifurcation:
Ah, thank you for reminding me of #277! I will rebase this and merge.

PR #251, opened 2019-11-18
Initial proposal for PSK inputs

beurdouche:
This is for injection of out-of-band shared entropy.


beurdouche:
Note that an alternative to this is to make PSK a group operation containing a PSK identifier.

PR #252, opened 2019-11-18
Bugfixes: fix a few typos in text and structs

beurdouche:


draft-ietf-mls-protocol.md
```diff
@@ -1007,7 +1007,7 @@ of a `ParentNodeHashInput` struct:
 ~~~~~
 struct {
     HPKEPublicKey public_key;
-    uint32_t unmerged_leaves<0..2^32-1>;
+    opaque unmerged_leaves<0..2^32-1>;
```
Bren2010: I don't think this is meant to be opaque. It's an array of leave indices

bifurcation:
Yeah, I noticed that too as I was implementing. I will fix it in a
“reconciliation with implementation” PR I’m working on.

PR #253, opened 2019-11-18
Initial version of the Tree of Signature

beurdouche:


bifurcation:
It seems like there's a major thing missing here, namely how the sender of a Commit transmits the new node signatures to the other members of the group.  We need something like that before this lands.

The most obvious approach would be to add a `signature` field to DirectPathNode, as you have with RatchetNode.

bifurcation:
It would also be handy to include in the intermediate node an indication of which leaf signed it.  Otherwise you have to do "trial verifications" (cf. trial decryption), which is expensive.

draft-ietf-mls-protocol.md
```diff
@@ -1960,16 +1983,11 @@ keys for nodes in its direct path that are being reset by the corresponding
...
-    HPKEPublicKey public_key;
-    opaque unmerged_leaves<0..2^32-1>;
-    optional<ClientInitKey> client_init_key;
-} RatchetNode;
-
 struct {
   // GroupContext inputs
   opaque group_id<0..255>;
   uint64 epoch;
+  opaque root_node_hash<0..255>;
```
raphaelrobert: This is a good starting point, we should see if we need something fancier in the future, like a Merkle tree (or reuse the node hashes).

draft-ietf-mls-protocol.md
```diff
@@ -1089,40 +1100,52 @@ struct {
...
+~~~~~
 struct {
     uint32 node_index;
-    optional<ParentNodeInfo> info;
+    optional<ParentNode> parent_node;
     opaque left_hash<0..255>;
     opaque right_hash<0..255>;
-} ParentNodeHashInput;
+    opaque signature<0..2^16-1>;
+} ParentNodeHash;
```
raphaelrobert: Isn't this where we discussed that the signer's index is missing?

kelrit:
> It seems like there's a major thing missing here, namely how the sender of a Commit transmits the new node signatures to the other members of the group. We need something like that before this lands.
> 
> The most obvious approach would be to add a `signature` field to DirectPathNode, as you have with RatchetNode.

Was this addressed? it's mentioned that the signature should be transmitted in "Synchronizing Views of the Tree" but there's no signature in the transmitted DirectPathNode

Bren2010:
@kelrit #287 changes the strategy for signing the tree, everything should be sorted out there.

PR #254, opened 2019-11-18
Initial work to use CIK in leaves

beurdouche:


Bren2010:
I went back and watched the recording of the last interim to get some context for this PR, and the motivation given there isn't entirely clear to me. The reasoning given then was:
- We want to allow rotation of identity key.
- This helps authenticate the leaf key by signing it.

But if we want to allow rotating the identity key, what we need is an operation that changes the `credential` field of the leaf. Putting the entire CIK in the leaf doesn't help that. And I imagine that the leaf key would already be signed by the changes required for #253.

bifurcation:
Note that this PR will need to change Commit as well -- the DirectPath in the commit sets the HPKEPublicKey in the sender's leaf node.  Suggested fix:

1. Update DirectPath so that the first node in the path is the one above the leaf (not the leaf itself)
2. Add a new ClientInitKey field to Commit that replaces the leaf

That's also forward-compatible with hash-based solutions to tree signing (cf. #253), since we can just define a ClientInitKey extension that contains the hash to be signed.

Bren2010:
To answer my own question, putting the entire CIK in the leaf (instead of just, say, a single signature) means the ratchet tree will contain information about node capabilities. This makes it possible for members being added to a group to detect downgrade attacks.

draft-ietf-mls-protocol.md
```diff
@@ -680,16 +680,22 @@ they receive the private keys for nodes, as described in
...
-path_secret[0] is used for the leaf, path_secret[1] for its parent,
-and so on.  At each step, the path secret is used to derive a new
-secret value for the corresponding node, from which the node's key
-pair is derived.
+When performing a Commit, the leaf ClientInitKey of the commiter and
+its direct path to the root are updated with new secret values.
+
+The generator of the Commit starts by using the HPKE secret key
+"leaf_hpke_secret_key" associated to the "init_key" of its new leaf
+CIK (See {{initialization-keys}}) to compute the "path_secret[0]"
```
Bren2010: - we've never abbreviated this before, CIK -> ClientInitKey
- lowercase, See -> see
- remove "the" in front of "path_secret[0]"

draft-ietf-mls-protocol.md
```diff
@@ -680,16 +680,22 @@ they receive the private keys for nodes, as described in
...
-secrets", one for each node from the leaf to the root.  That is,
-path_secret[0] is used for the leaf, path_secret[1] for its parent,
-and so on.  At each step, the path secret is used to derive a new
-secret value for the corresponding node, from which the node's key
-pair is derived.
+When performing a Commit, the leaf ClientInitKey of the commiter and
+its direct path to the root are updated with new secret values.
+
+The generator of the Commit starts by using the HPKE secret key
+"leaf_hpke_secret_key" associated to the "init_key" of its new leaf
```
Bren2010: associated to -> associated with

draft-ietf-mls-protocol.md
```diff
@@ -680,16 +680,22 @@ they receive the private keys for nodes, as described in
...
+CIK (See {{initialization-keys}}) to compute the "path_secret[0]"
+value and generates a sequence of "path secrets", one for each
```
Bren2010: - remove "value"
- generates -> generate

draft-ietf-mls-protocol.md
```diff
@@ -680,16 +680,22 @@ they receive the private keys for nodes, as described in
...
+intermediate nodes between the leaf and the root.  That is,
```
Bren2010: "intermediate nodes between the leaf and the root" -> "ancestor of its leaf"

draft-ietf-mls-protocol.md
```diff
@@ -987,23 +1103,19 @@ struct {
...
-respectively.  The `info` field is equal to the null optional value
-when the leaf is blank (i.e., no member occupies that leaf).
+The content within the leaf of a ratchet tree is composed of
+a `ClientInitKey` when the leaf is populated. The `info` field is
+equal to the null optional value when the leaf is blank (i.e., no
+member occupies that leaf).
 
-Likewise, the hash of a parent node (including the root) is the hash
-of a `ParentNodeHashInput` struct:
+The intermediate nodes contain less information, the hash of a parent
```
Bren2010: I would leave this sentence how it was, "Likewise, ..."

draft-ietf-mls-protocol.md
```diff
@@ -425,7 +425,7 @@ their state and the new client can use to initialize its state.
 To enforce forward secrecy and post-compromise security of messages,
 each member periodically updates its leaf secret which represents
 its contribution to the group secret.  Any member of the
-group can send an Update at any time by generating a fresh leaf secret
+group can send an Update at any time by using a fresh ClientInitKey
```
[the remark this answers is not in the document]
beurdouche: Work in progress... There is an editorial discussion going on between the editors to see what we should keep in the PR. That's likely gonna be discussed at the interim.

draft-ietf-mls-protocol.md
beurdouche: Thanks !

PR #255, opened 2019-11-18
Move epoch numbers to UInt64

beurdouche:


raphaelrobert:
#245 might invalidate this, but in the meantime we should go ahead with this!

PR #263, opened 2019-12-10
Change special values to structed types for external senders

kelrit:
Action item from Singapore: it seems better to specify directly in the message if we have an external sender instead of using reserved ranges.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1446,8 +1446,8 @@ enum {
 enum {
     invalid(0),
     member(1),
-    external_authority(2),
-    external_self(3),
+    preconfigured(2),
+    new_member(3),
     (255)
 } SenderType;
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1384,10 +1384,23 @@ enum {
     (255)
 } ContentType;
 
+enum {
+    invalid(0),
+    member(1),
+    external_authority(2),
```
bifurcation: I'm not sure "authority" is quite the semantic here.  Maybe just "external" or "preconfigured"?

draft-ietf-mls-protocol.md
```diff
@@ -1384,10 +1384,23 @@ enum {
...
+    external_self(3),
```
bifurcation: I think something like `new_member` might be clearer here.

draft-ietf-mls-protocol.md
```diff
@@ -1384,10 +1384,23 @@ enum {
...
+struct {
+    SenderType sender_type;
+    uint32 sender;
+} Sender;
+
 struct {
     opaque group_id<0..255>;
     uint32 epoch;
-    uint32 sender;
+    Sender sender;
```
bifurcation: There's no need to use the full `Sender` struct here, since you have to be a member to send an `MLSCiphertext`; just use uint32.  In other words, encrypting an MLSPlaintext with `Sender.sender_type != member` is an error, and an MLSPlaintext decrypted from an MLSCiphertext always has `Sender.sender_type != member`.

bifurcation: As a bonus, this makes it syntactically impossible for an external sender to send an MLSCiphertext.

draft-ietf-mls-protocol.md
```diff
@@ -1732,29 +1748,29 @@ A member of the group applies a Remove message by taking the following steps:
 
 * Blank the intermediate nodes along the path from the removed leaf to the root
 
-### External Proposals
+### External Proposals {#external-proposals}
```
bifurcation: Don't need the tag; it's auto-generated from the title itself.

draft-ietf-mls-protocol.md
```diff
@@ -1732,29 +1748,29 @@ A member of the group applies a Remove message by taking the following steps:
...
-MUST verify that the MLSPlaintext carrying the Proposal message is validly
-signed with this key.
-
-The remaining values 0xFFFFFF00 - 0xFFFFFFFE are reserved for signer that are
+a newly-hired staff member to a group representing a real-world team.  Proposals
+originating outside the group are identified by an `external` SenderType in
+MLSPlaintext.
+
+The `external_self` SenderType is used for clients proposing that they
+themselves be added.  For this ID type the sender value is ignored.  Proposals
```
bifurcation: Rather than ignoring, I would suggest that it MUST be zero.

draft-ietf-mls-protocol.md
```diff
@@ -1732,29 +1748,29 @@ A member of the group applies a Remove message by taking the following steps:
...
-The specific value 0xFFFFFFFF is reserved for clients proposing that they
-themselves be added.  Proposals with types other than Add MUST NOT be sent with
-this sender index.  In such cases, the MLSPlaintext MUST be signed with the
-private key corresponding to the ClientInitKey in the Add message.  Recipients
-MUST verify that the MLSPlaintext carrying the Proposal message is validly
-signed with this key.
-
-The remaining values 0xFFFFFF00 - 0xFFFFFFFE are reserved for signer that are
+a newly-hired staff member to a group representing a real-world team.  Proposals
+originating outside the group are identified by an `external` SenderType in
```
bifurcation: Note that `external` is not defined above, but probably should be :)

Also, this should be "`external` or `new_member`" (or whatever words we decide on).

draft-ietf-mls-protocol.md
kelrit: iiuc - the proposal is to treat all MLSPlaintexts as `member` sender types? That would mean external senders would need some other mechanism for sender? How can we differentiate between self add and a preconfigured sender?

I'm confused by this comment - were you thinking this was in MLSSenderData?

bifurcation: There's a difference between MLSPlaintext and MLSCiphertext here.  An MLSPlaintext can be constructed and sent by anyone; all you need is a signing key that is known to the group.  In order to construct an MLSCiphertext, you have to be a member of the group, since otherwise you won't have the right symmetric keys.  In other words, external senders can only send *unencrypted* messages to the group right now.  (I think Raphael is working on a PR to enable encryption to the group from outside.)

draft-ietf-mls-protocol.md
```diff
@@ -1477,9 +1490,12 @@ struct {
...
 ~~~~~
 
+External sender types are sent as MLSPlaintext, see {{external-proposals}}
+for their use.
+
 The remainder of this section describes how to compute the signature of
-an MLSPlaintext object and how to convert it to an MLSCiphertext object.
-The overall process is as follows:
+an MLSPlaintext object and how to convert it to an MLSCiphertext object
+for `member` sender types. The steps are:
```
bifurcation: In light of the above discussion, I would suggest adding a step to the encryption where the sender verifies that the `MLSPlaintext.sender.type == member`, then copies `MLSPlaintext.sender.sender` to `MLSCiphertext.sender`.

draft-ietf-mls-protocol.md
kelrit: Yea - I think we are on the same page. MLSPlaintext can be sent by anyone (in or outside the group) so it needs a full Sender object to disambiguate. However MLSCiphertext needs only the sender index, which is done by keeping MLSSenderData as is.

draft-ietf-mls-protocol.md
kelrit: It wasn't quite clear where that'd go; I re-organized the steps to be a little more clear and put the directive for constructing MLSSenderData objects (which is the only place in MLSCiphertext where the sender is identified)

draft-ietf-mls-protocol.md
bifurcation: Oh I see.  I was mistaken about what line we were looking at.  LGTM.

PR #265, opened 2019-12-11
Align spec text with implementation

bifurcation:
I have managed to implement Proposal / Commit in mlspp, but it required some modifications to the Welcome and Commit structs and processes.  Two major things to point out:

* This PR proposes to send the same DirectPath to both new members and current members.  This avoids the need to have two code paths and special tree math to implant path secrets, at the cost of some bloat in the Welcome and Commit messages.

* As a result, the sequencing of GroupContexts is a bit tricky.  The GroupContext used for encrypting/decrypting DirectPaths is an intermediate context -- the new group context, but *before the transcript is updated*.  

All that said, this version is [verified implementable](https://github.com/cisco/mlspp/pull/64).  The implementation exercise highlighted that the net effect of Proposal/Commit was to isolate basically all of the state management complexity to three methods in the client stack:

* Processing of Welcome messages
* Generation of Commit+Welcome messages
* Processing of Commit messages

This PR reflects the algorithms for those three things.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -644,7 +644,7 @@ In this tree, we can see all of the above rules in play:
 Every node, regardless of whether the node is blank or populated, has
 a corresponding _hash_ that summarizes the contents of the subtree
 below that node.  The rules for computing these hashes are described
-in {{tree-hashes}}.
+in {{tree-hashes-and-signatures}}.
 
 ## Views of a Ratchet Tree {#views}
 
@@ -1171,7 +1171,7 @@ The fields in this state have the following semantics:
 * The `epoch` field represents the current version of the group key.
 * The `tree_hash` field contains a commitment to the contents of the
   group's ratchet tree and the credentials for the members of the
-  group, as described in {{tree-hashes}}.
+  group, as described in {{tree-hashes-and-signatures}}.
 * The `confirmed_transcript_hash` field contains a running hash over
   the handshake messages that led to this state.
 
@@ -1228,7 +1228,7 @@ zero-length octet string.
 
 ## Direct Paths
 
-As described in {{ratchet-tree-commits}}, each MLS message needs to
+As described in {{commit}}, each MLS Commit message needs to
 transmit a ClientInitKey leaf and node values along its direct path.
 The path contains a public key and encrypted secret value for all
 intermediate nodes in the path above the leaf.  The path is ordered
@@ -1487,7 +1487,7 @@ struct {
           opaque application_data<0..2^32-1>;
 
         case proposal:
-          Proposal proposals;
+          Proposal proposal;
 
         case commit:
           Commit commit;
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1455,10 +1457,9 @@ struct {
           opaque application_data<0..2^32-1>;
 
         case proposal:
-          Proposal proposals<1..2^32-1>;
+          Proposal proposals;
```
Bren2010: `Proposal proposal` (singular)

draft-ietf-mls-protocol.md
```diff
@@ -1851,10 +1852,7 @@ contains the hash of the MLSPlaintext in which the Proposal was sent, using the
 hash function for the group's ciphersuite.
 
 ~~~~~
-struct {
-    uint32 sender;
-    opaque hash<0..255>;
-} ProposalID;
+opaque ProposalID<0..255>;
```
Bren2010: Update line 1849 to reflect this change

draft-ietf-mls-protocol.md
```diff
@@ -1879,40 +1877,91 @@ them to the group.  It might be we need to re-introduce this assumption, though
...
+* Construct an initial Commit object with `updates`, `removes`, `adds`, and
+  `ignored` fields populated from Proposals received during the current epoch,
+  and an empty `path` field.
 
-2. Verify that the signature on the MLSPlaintext message verifies using the
-   public key from the credential stored at the leaf in the tree indicated by
-   the `sender` field.
+* Generate a provisional GroupContext object by applying the proposals
+  referenced in the initial Commit object in the order provided, as described in
+  {{proposals}}.  Add proposals are applied left to right: Each Add proposal is
```
Bren2010: Nit: one space between sentences

draft-ietf-mls-protocol.md
Bren2010: Please also update MLSPlaintextSignatureInput

Bren2010: Please also change to the singular in MLSPlaintext

bifurcation:
@beurdouche I added one minor change (order of fields) and rebased.  I think this is ready to merge.

beurdouche:
Ack, I’ll take a look in tonight or tomorrow morning and merge it.
B.

beurdouche:
I am merging this now because it includes fixes we definitely need even though I note that I have one philosophical concern and one security concern...

1. Multiplying the numbers of transcript hashes in the state is an implementation concern and not a protocol concern, I would much prefer keep only the real (confirmed) transcript hash. (The implementation should keep the current state and one provisional updated state but the struct shouldn't have to keep both)
2. Sending the init_secret and the update_secret at once *seem* equivalent to sending the epoch secret but I don't think it is actually the case since it could be rejected by the DS, breaking the encryption key of that message would hence still have consequences even if the message is rejected. I'll open an issue for that.

Issue #264, opened 2019-12-10
Fix inconsistent definitions of Derive-Secret


bifurcation:
On the one hand...
```
Derive-Secret(Secret, Label) =
    HKDF-Expand-Label(Secret, Label, "", Hash.length)
```

On the other hand...
```
...
 |
 +--> Derive-Secret(., "sender data", GroupContext_[n])
 |    = sender_data_secret
```

We probably want Derive-Secret to have a third argument for the GroupContext, or define that Derive-Secret always takes the current group context.

beurdouche:
I would much prefer to go back to the TLS definition, we don't need the `GroupContext` thing at all, this is an implementation artifact...

Definition we use for TLS and the symbolic proof in the formal model:
```Derive-Secret(Secret, Label, Messages) =
            HKDF-Expand-Label(Secret, Label,
                              Transcript-Hash(Messages), Hash.length)```

beurdouche:
Ps: because `Messages` is quite long in MLS, we directly pass `Transcript-Hash(Messages)` to `Derive-Secret` instead.

PR #274, opened 2019-12-20
Fix contexts and simplify key schedule

beurdouche:
This is a fix for #264 as well as for missing context that was previously partially added in the messages but not in the Tree intermediate derivations.

The changes are centered around putting back the `transcript_hash` (aka the full history) to be what should be included as default context.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -697,14 +697,16 @@ node, from which the node's key pair is derived.
 
 ~~~~~
 path_secret[0] = HKDF-Expand-Label(leaf_hpke_secret_key,
-                                   "path", "", Hash.Length)
+                                   "path", thx, Hash.Length)
 path_secret[n] = HKDF-Expand-Label(path_secret[n-1],
-                                   "path", "", Hash.Length)
+                                   "path", thx, Hash.Length)
 node_secret[n] = HKDF-Expand-Label(path_secret[n],
-                                   "node", "", Hash.Length)
+                                   "node", thx, Hash.Length)
 node_priv[n], node_pub[n] = Derive-Key-Pair(node_secret[n])
 ~~~~~
 
+In this setting, `thx` represents the value of the `transcript_hash`
+for the current epoch of the tree.
 For example, suppose there is a group with four members:
 
 ~~~~~
@@ -1134,7 +1136,7 @@ struct {
     opaque group_id<0..255>;
     uint64 epoch;
     opaque tree_hash<0..255>;
-    opaque confirmed_transcript_hash<0..255>;
+    opaque transcript_hash<0..255>;
 } GroupContext;
 ~~~~~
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -697,14 +697,16 @@ node, from which the node's key pair is derived.
 
 ~~~~~
 path_secret[0] = HKDF-Expand-Label(leaf_hpke_secret_key,
-                                   "path", "", Hash.Length)
+                                   "path", txh, Hash.Length)
```
bifurcation: I would prefer to spell this out, if we have horizontal space

draft-ietf-mls-protocol.md
```diff
@@ -1181,24 +1183,20 @@ struct {
   opaque signature<0..2^16-1>;
 } MLSPlaintextCommitAuthData;
 
-confirmed_transcript_hash_[n] =
-    Hash(interim_transcript_hash_[n-1] ||
+transcript_hash_[n] =
+    Hash(transcript_hash_[n-1] ||
```
bifurcation: I don't understand your proposal here.  The transcript hash you define has an odd structure -- it only covers the CommitContent, not the AuthData.  So the confirmation hashes and signatures are not covered.  It's as if the TLS transcript hash didn't include CertificateVerify or Finished.  Is that your intent?

bifurcation: In any case, if you're going to remove the interim transcript hash, you need to do more radical surgery.   For example, MLSPlaintextCommitAuthData is no longer needed; likewise much of the prose below.

draft-ietf-mls-protocol.md
```diff
@@ -697,14 +697,16 @@ node, from which the node's key pair is derived.
...
-                                   "path", "", Hash.Length)
+                                   "path", txh, Hash.Length)
 node_secret[n] = HKDF-Expand-Label(path_secret[n],
-                                   "node", "", Hash.Length)
+                                   "node", txh, Hash.Length)
 node_priv[n], node_pub[n] = Derive-Key-Pair(node_secret[n])
 ~~~~~
 
+In this setting, `txh` represents the value of the `transcript_hash`
+for the current epoch of the tree.
```
bifurcation: Here and below, I would prefer if we expanded this.

draft-ietf-mls-protocol.md
```diff
@@ -1293,33 +1291,38 @@ proceeds as shown in the following diagram:
...
                      |
                      V
                init_secret_[n]
 ~~~~~
 
+Here we denote txh[n] the `transcript_hash` for epoch N, which result
+from hashing the previous transcript hash (N-1) with the MLSCiphertext
+that lead to the current epoch_secret[n].
+
+
```
bifurcation: Nit: Extra blank line.

beurdouche:
@bifurcation Yes, this is still a "Work in Progress" and is labeled as such... :) 
It will probably not be updated it before RWC... I'll ping when it is in a reasonable state.

bifurcation:
Closing for now.  Happy to reconsider on refresh/rebase.

PR #275, opened 2019-12-20
Mandate expiration of ClientInitKeys and presence of some extensions

beurdouche:


draft-ietf-mls-protocol.md
```diff
@@ -1050,6 +1050,8 @@ ProtocolVersion supported_versions<0..255>;
 CipherSuite supported_ciphersuites<0..255>;
 ~~~~~
 
+This extension MUST be always present in an MLS ClientInitKey.
```
bifurcation: There are two extensions here.  So if you mean both of them, you should say so.

draft-ietf-mls-protocol.md
```diff
@@ -1063,11 +1065,13 @@ it for any further processing.
...
-ClientInitKeys to ensure continued reachability may choose to omit the
-expiration extension from these keys, or give them much longer lifetimes than
-other ClientInitKeys.
+Applications that rely on "last resort" ClientInitKeys MAY set the
+expiration to its maximum value even though this is NOT RECOMMENDED.
+It is RECOMMENDED to rotate last resort keys at a pace chosen by the
+application even though they can have much longer lifetimes than other
+ClientInitKeys.
+
+This extension MUST be always present in an MLS ClientInitKey.
```
bifurcation: The need for this is not clear to me.  Let's discuss this at the interim before merging this PR.

draft-ietf-mls-protocol.md
```diff
@@ -1050,6 +1050,8 @@ ProtocolVersion supported_versions<0..255>;
 CipherSuite supported_ciphersuites<0..255>;
 ~~~~~
 
+These extensions MUST be always present in an MLS ClientInitKey.
```
Bren2010: I believe this is only one extension

Bren2010: but "MUST always be" would be nicer

draft-ietf-mls-protocol.md
bifurcation: This was discussed at the interim, and people were comfortable with this requirement.

draft-ietf-mls-protocol.md
Bren2010:
```suggestion
These extensions MUST be always present in a ClientInitKey.
```

draft-ietf-mls-protocol.md
Bren2010:
```suggestion
This extension MUST always be present in a ClientInitKey.
```

draft-ietf-mls-protocol.md
```diff
@@ -1063,11 +1065,13 @@ it for any further processing.
...
-Note that as an extension, it is not required that any given ClientInitKey have
-an expiration time.  In particular, applications that rely on "last resort"
-ClientInitKeys to ensure continued reachability may choose to omit the
-expiration extension from these keys, or give them much longer lifetimes than
-other ClientInitKeys.
+Applications that rely on "last resort" ClientInitKeys MAY set the
+expiration to its maximum value even though this is NOT RECOMMENDED.
+It is RECOMMENDED to rotate last resort keys at a pace chosen by the
+application even though they can have much longer lifetimes than other
+ClientInitKeys.
```
Bren2010: The expiration of ClientInitKeys is important for PCS, so setting to the maximum value shouldn't be encouraged. If you have keys that never expire and you're compromised, those keys can be used at any time in the future to "add" you to a group when they're actually giving the adversary access.

Last Resort keys are a defense against DoS, when somebody sends you a lot of messages in a short period of time. As such, any ClientInitKey can be a Last Resort key, regardless of how long it's valid for, because nobody is checking how many times a specific key gets used.

Last Resort keys are not a suitable defense against the user being offline for an arbitrary amount of time. Handling this securely probably requires more consideration.

We should recommend an application-specified maximum duration, and switch to a `(notBefore, notAfter)` system so that everyone know's the key's lifetime.

PR #279, opened 2019-12-28
Bring back more ciphersuites

raphaelrobert:


draft-ietf-mls-protocol.md
```diff
@@ -818,35 +818,21 @@ strings with length Hash.length to HPKE key pairs.
...
 protocol (see the Ciphersuites section of the HPKE specification for more
-information):
+information).
 
 ~~~~~
-enum {
-    P256_SHA256_AES128GCM(0x0000),
-    X25519_SHA256_AES128GCM(0x0001),
-    (0xFFFF)
-} CipherSuite;
```
Bren2010: Now the CipherSuite type is never defined

draft-ietf-mls-protocol.md
```diff
@@ -2447,26 +2436,48 @@ and administered under a Specification Required policy {{!RFC8126}}.
...
-[[ Note to RFC Editor: Please replace "XXXX" above with the number assigned to
-this RFC. ]]
+| Component | Contents |
+|:----------|:---------|
+| MLS       | The string "MLS" followed by the major and minor version, e.g. "MLS10" |
+| LVL       | The security level |
+| KEM       | The KEM algorithm used for HPKE in TreeKEM group operations |
+| AEAD      | The AEAD algorithm used for HPKE and message protection |
+| HASH      | The hash algorithm used for HPKE and the MLS KDF |
+| SIG       | The Signature algorithm used for message authentication |
```
Bren2010: Please update line 810 to say that ciphersuites are defined by: KEM, AEAD, HASH, SIG.

draft-ietf-mls-protocol.md
```diff
@@ -2447,26 +2436,48 @@ and administered under a Specification Required policy {{!RFC8126}}.
...
+| MLS10_256_HPKEP521_AES256GCM_SHA384_P521            | {0x00,0x05} |
+| MLS10_256_HPKEX448_CHACHA20POLY1305_SHA256_Ed448    | {0x00,0x06} |
+
+The KEM/DEM constructions used for HPKE are defined by {{HPKE}}.
+The corresponding AEAD algorithms AEAD_AES_128_GCM and AEAD_AES_256_GCM, are
+defined in {{RFC5116}}. AEAD_CHACHA20_POLY1305 is defined
+in {{RFC7539}}. The corresponding hash algorithms are defined in {{!SHS}}.
+
+The mandatory-to-implement ciphersuite for MLS 1.0 is
+`MLS10\_128\_HPKE25519\_AES128GCM\_SHA256\_Ed25519` which is using
```
Bren2010: Shouldn't the mandatory suite be P-256? X25519 has never been mandatory

draft-ietf-mls-protocol.md
```diff
@@ -818,35 +818,21 @@ strings with length Hash.length to HPKE key pairs.
...
-
-This ciphersuite uses the following primitives:
-
-* Hash function: SHA-256
-* AEAD: AES-128-GCM
+The ciphersuites are defined in section {{MLS Ciphersuites}}.
 
-When HPKE is used with this ciphersuite, it uses the following
-algorithms:
+### Notes on Diffie-Hellman groups
```
Bren2010: I would make this section at the same level as "Ciphersuites" (so "## Diffie-Hellman Groups") and add explainer text that DH groups are used to build KEMs with HPKE

draft-ietf-mls-protocol.md
```diff
@@ -2447,26 +2436,48 @@ and administered under a Specification Required policy {{!RFC8126}}.
 
 ## MLS Ciphersuites
 
-The "MLS Ciphersuites" registry lists identifiers for suites of cryptographic
-algorithms defined for use with MLS.  These are two-byte values, so the maximum
-possible value is 0xFFFF = 65535.  Values in the range 0xF000 - 0xFFFF are
-reserved for vendor-internal usage.
+A cipher suite value defines a combinaison of a protocol version
+and of the set of cryptographic algorithms used for this version MLS.
```
Bren2010: A ciphersuite is a combination of a protocol version and the set of cryptographic algorithms that should be used.

draft-ietf-mls-protocol.md
raphaelrobert: It is defined in section "MLS Ciphersuites" as a two byte value now.

draft-ietf-mls-protocol.md
raphaelrobert: The missing part would be the signature, but the context here is "primitives to be used in group key computations".
I guess we could reference HPKE here.

draft-ietf-mls-protocol.md
raphaelrobert: > Shouldn't the mandatory suite be P-256?

Why?

> X25519 has never been mandatory

It has always been so far

draft-ietf-mls-protocol.md
raphaelrobert: Thanks, will address this in another commit.

draft-ietf-mls-protocol.md
raphaelrobert: I'll add the explainer, but I think it makes sense to keep it at this level, since it is about ciphersuites after all.

raphaelrobert: (added)

draft-ietf-mls-protocol.md
```diff
@@ -818,35 +818,21 @@ strings with length Hash.length to HPKE key pairs.
...
 opaque HPKEPublicKey<1..2^16-1>;
 ~~~~~
 
-### Curve25519, SHA-256, and AES-128-GCM
-
-This ciphersuite uses the following primitives:
-
-* Hash function: SHA-256
-* AEAD: AES-128-GCM
+The ciphersuites are defined in section {{MLS Ciphersuites}}.
```
bifurcation: I think you need a bit more description here, just to say what a ciphersuite indicates, namely:

* Parameters for an instance of HPKE
* A KDF to be used 
* As a consequence, a Derive-Key-Pair function, as defined below.

bifurcation: Also, does `{{MLS Ciphersuites}}` actually work?  Might need to be `{{mls-ciphersuites}}`.

draft-ietf-mls-protocol.md
bifurcation: This extra section heading doesn't seem necessary.

draft-ietf-mls-protocol.md
```diff
@@ -857,19 +843,22 @@ Section 6 of {{RFC7748}}.  If implementers use an alternative
...
 
-### P-256, SHA-256, and AES-128-GCM
+#### X448
 
-This ciphersuite uses the following primitives:
-
-* Hash function: SHA-256
-* AEAD: AES-128-GCM
+Given an octet string X, the private key produced by the
+Derive-Key-Pair operation is SHA-512(X) truncated to 448 bits (Recall that any 56-octet
```
bifurcation: Should we use HKDF instead of SHA directly?  In other words, make node secrets variable-length, where the length is determined by the KEM in use.

* P256, X25519: 32 bytes
* X448: 56 bytes
* P521: 66 bytes

As a bonus, that would save a couple of hash invocations.

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
raphaelrobert: Have we decided on that? In theory there could be different signature algorithms for credentials and message signing. Since credentials are completely defined by the AS, it might be tricky to get consensus when picking the CIKs.

draft-ietf-mls-protocol.md
```diff
@@ -2447,26 +2436,48 @@ and administered under a Specification Required policy {{!RFC8126}}.
...
-| 0x0001 | X25519_SHA256_AES128GCM | RFC XXXX  |
+~~~
+   CipherSuite MLS_LVL_KEM_AEAD_HASH_SIG = VALUE;
+~~~
 
-[[ Note to RFC Editor: Please replace "XXXX" above with the number assigned to
-this RFC. ]]
+| Component | Contents |
+|:----------|:---------|
+| MLS       | The string "MLS" followed by the major and minor version, e.g. "MLS10" |
```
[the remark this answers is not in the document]
raphaelrobert: I wanted to avoid exactly that scenario, it could be dangerous to re-use ciphersuites between versions is the primitives are used for a different purpose in the new version. Being more explicit might be the safer choice here.

draft-ietf-mls-protocol.md
```diff
@@ -2447,26 +2436,48 @@ and administered under a Specification Required policy {{!RFC8126}}.
...
+| LVL       | The security level |
```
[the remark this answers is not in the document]
raphaelrobert: The idea is to give you a rough idea of the overall security level for informational purposes. Maybe @beurdouche can elaborate more on that.

draft-ietf-mls-protocol.md
```diff
@@ -2447,26 +2436,48 @@ and administered under a Specification Required policy {{!RFC8126}}.
...
+| KEM       | The KEM algorithm used for HPKE in TreeKEM group operations |
+| AEAD      | The AEAD algorithm used for HPKE and message protection |
+| HASH      | The hash algorithm used for HPKE and the MLS KDF |
+| SIG       | The Signature algorithm used for message authentication |
+
+This specification defines the following cipher suites for use with MLS 1.0.
+
+|          Description                                |    Value    |
+|:----------------------------------------------------|:------------|
+| MLS10_128_HPKEX25519_AES128GCM_SHA256_Ed25519       | {0x00,0x01} |
```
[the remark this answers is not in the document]
raphaelrobert: The idea was to explicitly base this on what HPKE defines. This is just to emphasize that we want to use HPKE rather than some other KEM mechanism.

draft-ietf-mls-protocol.md
```diff
@@ -2447,26 +2436,48 @@ and administered under a Specification Required policy {{!RFC8126}}.
...
+| MLS10_256_HPKEX448_AES256GCM_SHA384_Ed448           | {0x00,0x04} |
+| MLS10_256_HPKEP521_AES256GCM_SHA384_P521            | {0x00,0x05} |
+| MLS10_256_HPKEX448_CHACHA20POLY1305_SHA256_Ed448    | {0x00,0x06} |
+
+The KEM/DEM constructions used for HPKE are defined by {{HPKE}}.
+The corresponding AEAD algorithms AEAD_AES_128_GCM and AEAD_AES_256_GCM, are
+defined in {{RFC5116}}. AEAD_CHACHA20_POLY1305 is defined
+in {{RFC7539}}. The corresponding hash algorithms are defined in {{!SHS}}.
+
+The mandatory-to-implement ciphersuite for MLS 1.0 is
```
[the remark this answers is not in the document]
raphaelrobert: Why?

draft-ietf-mls-protocol.md
raphaelrobert: 1) That's why I wanted to leave "HPKE" in the name of the ciphersuite
2) I guess that should then go in line 809
3) That's what we have in the "Diffie-Hellmann groups section"

Regarding the notation for hyperlinks, I'll take your word for it

draft-ietf-mls-protocol.md
raphaelrobert: Unless there is a specific reasons not to do it (cc @beurdouche), I think it would be great to harmonise this. Having different hash algorithms in here makes no sense to me if it can be avoided.

bifurcation:
It seems like we have a few separable issues here:
1. Whether we require a single signature scheme for the whole group
2. Whether signature schemes are included in the ciphersuite
3. Whether we have an MTI ciphersuite / signature algorithm
4. What new ciphersuites should be defined, and what algorithms should go in them

Let's discuss this at the interim and maybe break this PR into a few bits

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
raphaelrobert: In the discussion at the last interim the consensus was to keep the signature algorithms as part of the ciphersuite.

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
raphaelrobert: As discussed at the last interim, we want ciphersuites to be specific for a certain MLS protocol version.

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
raphaelrobert: As discussed at the interim, we want to include the MTI.

draft-ietf-mls-protocol.md
raphaelrobert: As  discussed at the last interim, we keep this MTI.

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
raphaelrobert: HPKE changed to DHKEM

draft-ietf-mls-protocol.md
raphaelrobert: done

seanturner:
Note that specification required registries imply that there is a designated expert pool. I will add some text to create one. The text is very similar to the TLS DE pool text from [RFC8447](https://datatracker.ietf.org/doc/rfc8447/) with some tweaks that, hopefully, address the misunderstandings uncovered during non-standards track registry requests.

Note the process is: (assuming that the IESG eventually approves this draft) When the IESG approves this draft, the responsible AD assigns some DEs (like 2-3 suggested by the WG chairs). These DEs are in charge of all registrations regardless of whether the draft comes from the WG or elsewhere as all requests go to the MLS DEs mailing list.

draft-ietf-mls-protocol.md
seanturner: I fixed this in mine. The anchor is {#anchor} and then you link in text with {{anchor}}.

draft-ietf-mls-protocol.md
seanturner: I blew it away in mine.

draft-ietf-mls-protocol.md
```diff
@@ -818,35 +819,25 @@ strings with length Hash.length to HPKE key pairs.
...
-information):
+information).
 
-~~~~~
-enum {
-    P256_SHA256_AES128GCM(0x0000),
-    X25519_SHA256_AES128GCM(0x0001),
-    (0xFFFF)
-} CipherSuite;
+The signature algorithm specified in the ciphersuite is the mandatory algorithm to be used for the signatutes in MLSPlaintext and the tree signatures. It can be different from the signature algorithm specified in credential fielf of ClientInitKeys.
```
Bren2010: - Please wrap this at 80-ish chars
- "field"
- Can it really be different?

draft-ietf-mls-protocol.md
```diff
@@ -818,35 +819,25 @@ strings with length Hash.length to HPKE key pairs.
...
-* KEM: 0x0002 = DHKEM(Curve25519)
-* KDF: 0x0001 = HKDF-SHA256
-* AEAD: 0x0001 = AES-GCM-128
+#### X25519
 
 Given an octet string X, the private key produced by the
-Derive-Key-Pair operation is SHA-256(X).  (Recall that any 32-octet
-string is a valid Curve25519 private key.)  The corresponding public
+Derive-Key-Pair operation is SHA-256(X) (Recall that any 32-octet
+string is a valid X25519 private key). The corresponding public
```
Bren2010: The punctuation was correct before
Alternatively, lower-case r in "Recall"

draft-ietf-mls-protocol.md
```diff
@@ -2447,26 +2441,56 @@ and administered under a Specification Required policy {{!RFC8126}}.
...
-reserved for vendor-internal usage.
+A ciphersuite is a combination of a protocol version and the set of cryptographic algorithms that should be used.
+
+Cipher suite names follow the naming convention:
+
+~~~
+   CipherSuite MLS_LVL_KEM_AEAD_HASH_SIG = VALUE;
+~~~
+
+Where VALUE is a is represented as two 8bit octets:
```
Bren2010: VALUE is represented
8-bit

draft-ietf-mls-protocol.md
```diff
@@ -2447,26 +2441,56 @@ and administered under a Specification Required policy {{!RFC8126}}.
...
+uint8 CipherSuite[2];
+
+| Component | Contents |
+|:----------|:---------|
+| MLS       | The string "MLS" followed by the major and minor version, e.g. "MLS10" |
+| LVL       | The security level |
+| KEM       | The KEM algorithm used for HPKE in TreeKEM group operations |
+| AEAD      | The AEAD algorithm used for HPKE and message protection |
+| HASH      | The hash algorithm used for HPKE and the MLS KDF |
+| SIG       | The Signature algorithm used for message authentication |
```
Bren2010: lower-case s in "Signature"

draft-ietf-mls-protocol.md
```diff
@@ -2447,26 +2441,56 @@ and administered under a Specification Required policy {{!RFC8126}}.
...
-* Name: The name of the ciphersuite
-* Reference: Where this algorithm is defined
+It is advisable to keep the number of ciphersuites low to increase the chances clients can interoperate in a federated environment, therefore the ciphersuites only inlcude modern, yet well-established algorithms.
+Depending on their requirements, clients can choose between two security levels (roughly 128 bit and 256bit). Within the security levels clients can choose between faster X25519/X448 curves and FIPS 140-2 compliant curves for Diffie-Hellman key negotiations. Additionally clients that run predominantly on mobile processors can choose ChaCha20Poly1305 over AES-GCM for performance reasons. Since ChaCha20Poly1305 is not listed by FIPS 140-2 it is not paired with FIPS 140-2 compliant curves. The security level of symmetric encryption algorithms and hash functions is paired with the security level of the curves.
 
-The initial contents for this registry are as follows:
+The mandatory-to-implement ciphersuite for MLS 1.0 is
+`MLS10\_128\_HPKE25519\_AES128GCM\_SHA256\_Ed25519` which is using
+Curve25519, HKDF over SHA2-256 and AES-128-GCM for HPKE,
```
Bren2010: HKDF over SHA-256, AES-128-GCM for HPKE,

draft-ietf-mls-protocol.md
```diff
@@ -2447,26 +2441,56 @@ and administered under a Specification Required policy {{!RFC8126}}.
...
+and AES-128-GCM with Ed25519 for symmetric encryption and
+signatures.
```
Bren2010: "message encryption and authentication"

PR #281, opened 2020-01-03
Extend the epoch with a commit hash

bifurcation:
As discussed on the mailing list, this allows for fork detection.


bifurcation:
I am withdrawing this proposal.  Discussion at the interim indicated that consideration of fork/merge cases is more complex than it might have seemed, and it's not clear that this extension is needed outside of those cases.

And in any case, if something like this is needed, it seems like it can be done in an extension.

PR #282, opened 2020-01-03
Specify an Ack message

bifurcation:


bifurcation:
After discussion at Jan 2020 interim, the current feeling is that Acks are not necessary at the moment.  We may re-add them later, once we have some more thorough analysis of recovery scenarios.

PR #283, opened 2020-01-03
Use the same ratchet for Handshake and Application keys

bifurcation:
The current draft uses a hash ratchet to derive single-use keys for application messages, but for handshake messages, re-uses the same key with a sequence of different nonces.  This results in a lot of unnecessary extra code.

This PR uses the same hash ratchet for handshake and application messages.  The only difference is in the base key for the hash ratchet.  In the application case, we use the AStree.  In the HS case, we derive the base keys directly from the handshake secret.  The lack of a tree in the HS case means that either the client will do a linear number of derivations on the first send/receive, or it will keep the handshake secret around and not get FS for handshake messages within the epoch.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1379,7 +1379,7 @@ As described in {{message-framing}}, MLS encrypts three different
 types of information:
 
 * Metadata (sender information)
-* Handshake messages (e.g., Proposal and Commit)
+* Handshake messages (Proposal and Commit)
 * Application messages
 
 The sender information used to look up the key for the content encryption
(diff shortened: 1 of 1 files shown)


bifurcation:
@psyoptix - would appreciate your thoughts here

bifurcation:
As discussed on call:
* There's no point to FS for Proposals because clients have to cache the plaintext anyway
* Given that, the "flat derivation" approach should be fine
* We should have separate keys per sender to it easier to avoid nonce collisions
* Unclear whether we should derive nonces on a hash ratchet or use a counter

draft-ietf-mls-protocol.md
```diff
@@ -1379,52 +1379,64 @@ As described in {{message-framing}}, MLS encrypts three different
...
-in a similar fashion. This allows forward secrecy at the level of
-application messages within and out of an epoch.
-A step in this chain (the second subscript) is called a "generation".
-The details of application key derivation are described in the
-{{astree}} section below.
+The base secret of for each sender is used to initiate a symmetric hash ratchet
+which generates a sequence of keys and nonces. The sender uses the j-th
+key/nonce pair in the sequence to encrypt (using the AEAD) the j-th message they
+send during that epoch.  In particular, each key/nonce pair MUST NOT be used to
+encrypt more than one message.
```
raphaelrobert: There is a fundamental difference between application message sender ratchets and handshake message ratchets: the requirement for forward secrecy.
For application messages clients MUST delete old key material in order to obtain FS.
For handshake messages FS is not required. Moreover since there is no ordering requirement for Proposals clients should keep older key material around in order to be able to decrypt out-of-order handshake messages within one epoch.
I think it's important to make that more clear.

PR #286, opened 2020-01-06
Editorial: Unclear that Commits always include an Update/refreshes the CIK for the committer.

Bren2010:


beurdouche:
You can now rebase this on master for clarity.

draft-ietf-mls-protocol.md
```diff
@@ -1926,15 +1922,16 @@ the next epoch change anyway. ]]
...
+  fields populated from Proposals received during the current epoch, an Update
+  proposal for the member's own leaf in `committer_update`, and an empty `path`
+  field.
 
 * Generate a provisional GroupContext object by applying the proposals
   referenced in the initial Commit object in the order provided, as described in
   {{proposals}}. Add proposals are applied left to right: Each Add proposal is
   applied at the leftmost unoccupied leaf, or appended to the right edge of the
-  tree if all leaves are occupied.
+  tree if all leaves are occupied. The committer's own Update is applied last.
```
bifurcation: It would be good to make clear that the Update in question is the `committer_update`.  Also, could you please check to make sure that DirectPath no longer touches the leaf?  Since we have split out Update and `committer_update`, DirectPath should start at the parent of the Committer.

Bren2010: https://github.com/mlswg/mls-protocol/pull/294

draft-ietf-mls-protocol.md
```diff
@@ -1914,18 +1897,31 @@ struct {
     ProposalID updates<0..2^16-1>;
     ProposalID removes<0..2^16-1>;
     ProposalID adds<0..2^16-1>;
-    ProposalID ignored<0..2^16-1>;
+
+    Update committer_update;
```
raphaelrobert: Since an Update only contains a CIK, we should use the CIK right away here.

PR #287, opened 2020-01-06
Switch to signing strategy using one signature per leaf.

Bren2010:


raphaelrobert:
After the discussion at the interim I took a look again and I would like to bring up the following points:

 - The PR contains functional changes as well as editorial changes. I think it makes no sense to combine the two and the two kind of changes should be split in at least two PRs.
 - We discussed that enforcing the tree invariant is the minimum we need, there is still an ongoing discussion on what we want to guarantee regarding group agreement.
 - This PR covers more than just the direct path needed for the tree invariant, because the fields `left_hash` and `right_hash` are present in `ParentNodeHashInput`. Is this on done on purpose? If so, why?
 - I think it is not actually possible to verify the hash chain in a direct path other than in the direct path of the last Commit. The reason is that the hash chain breaks on all other direct paths because at some level the direct path gets overwritten by the direct path of the committer and the hashes are therefore no longer correct. The way forward would be to use a Merckle tree instead of a hash in the CIK extension (I think @bifurcation proposed that at some point).

beurdouche:
You can now rebase this on master for clarity.
Please avoid unrelated editorial as much as possible.

bifurcation:
As discussed on PR call:
* People are generally OK going with this approach for now, pending further proofs
* We should get #285 and #286 merged first

raphaelrobert:
I think the last commits address my concerns.

bifurcation:
One more note here: It might be good to add an OPEN ISSUE to acknowledge the still-outstanding concerns about group agreement vs. deniability, , something like:

> [[ OPEN ISSUE: This scheme, in which the tree hash covers the parent hash, is designed to allow for more deniable deployments, since a signature by a member covers only its direct path.  The other possible scheme, in which the parent hash covers the tree hash, provides better group agreement properties, since a member's signature covers the entire membership of the trees it is in.  Further discussion is needed to determine whether the benefits to deniability justify the harm to group agreement properties, or whether there are alternative approaches to deniability that could be compatible with the other approach. ]]

PR #294, opened 2020-01-30
Re-define direct path to not include the leaf.

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -522,14 +522,15 @@ used for the Merkle trees in the Certificate Transparency protocol
...
+is the concatenation of that node's parent along with the parent's direct path.
+The _copath_ of a node is the node's sibling concatenated with the list of
+siblings of all the nodes in its
 direct path. The _frontier_ of a tree is the list of heads of the maximal
 full subtrees of the tree, ordered from left to right.
 
 For example, in the below tree:
 
-* The direct path of C is (C, CD, ABCD)
+* The direct path of C is (CD, ABCD)
```
bifurcation: Add root here, so it lines up with the copath

bifurcation:
You can also strike this sentence further down: " In particular, for the leaf node, there are no encrypted secrets, since a leaf node has no children."

bifurcation:
Note that this severs the connection between the HPKE key in the leaf and the parent of the leaf.  Whereas before, the leaf HPKE key was derived from a path secret, and the parent HPKE key was derived from the next path secret.  Now the parent path secret is set directly, and the leaf HPKE key is set completely independently, with no dependency on a path secret.

So we're losing a bit of internal structure in the tree.  But this seems OK because that structure was never observable to the group -- only the holder of the leaf knows the leaf path secret, so nobody else could verify that the leaf was related to its parent.

PR #295, opened 2020-02-01
Use path secret instead of full DirectPath

bifurcation:
This PR adapts the Welcome logic so that the new joiners get their path secrets directly, rather than from a DirectPath.  This cleans things up in a few ways.  The new joiners no longer have to have the context for the previous epoch (which was used to encrypt the DirectPath).  And we can get away with one HPKE operation per new joiner instead of two (one for the EncryptedKeyPackage and one for the DirectPath).

The major cost is that there's a bit more complexity in the TreeKEM.  Instead of just producing and consuming DirectPaths, you now need to know which path secrets go with which nodes.  You also need tree math to identify the common ancestor of two leaves, but there's a simple closed-form formula for that, expressed in the Python code here.

[Corresponding PR on the Go implementation.](https://github.com/bifurcation/mls/pull/37)

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1949,13 +1949,6 @@ message at the same time, by taking the following steps:
   applied at the leftmost unoccupied leaf, or appended to the right edge of the
   tree if all leaves are occupied.
 
-* Create an initial, partial GroupInfo object reflecting the following values:
-  * Group ID: The group ID for the group
-  * Epoch: The epoch ID for the next epoch
-  * Tree: The group's ratchet tree after the commit has been applied
-  * Prior confirmed transcript hash: The confirmed transcript hash for the
-    current state of the group (not the provisional state)
-
 * Create a DirectPath using the new tree (which includes any new members).  The
   GroupContext for this operation uses the `group_id`, `epoch`, `tree`, and
   `prior_confirmed_transcript_hash` values in the initial GroupInfo object.
@@ -1970,19 +1963,25 @@ message at the same time, by taking the following steps:
   value in the MLSPlaintext.  Sign the MLSPlaintext using the current epoch's
   GroupContext as context.
 
-* Complete the GroupInfo by populating the following fields:
-  * Confirmed transcript hash: The confirmed transcript hash including the
-    current Commit object
-  * Interim transcript hash: The interim transcript hash including the current
-    Commit object
-  * Confirmation: The confirmation from the MLSPlaintext
-  * Sign the GroupInfo using the member's private signing key
-  * Encrypt the GroupInfo using the key and nonce derived from the `init_secret`
-    for the current epoch (see {{welcoming-new-members}})
+* Update the tree in the provisional state by applying the direct path
 
-* For each new member in the group, compute an EncryptedKeyPackage object that
-  encapsulates the `init_secret` for the current epoch.  Construct a Welcome
-  message from the encrypted GroupInfo object and the encrypted key packages.
+* Construct a GroupInfo reflecting the new state:
+  * Group ID, epoch, tree, confirmed transcript hash, and interim transcript
+    hash from the new state
+  * The confirmation from the MLSPlaintext object
+  * Sign the GroupInfo using the member's private signing key
+  * Encrypt the GroupInfo using the key and nonce derived from the `epoch_secret`
+    for the new epoch (see {{welcoming-new-members}})
+
+* For each new member in the group:
+  * Identify the lowest common ancestor in the tree of the new member's
+    leaf node and the member sending the Commit
+  * Compute the path secret corresponding to the commonn ancestor node
+  * Compute an EncryptedKeyPackage object that encapsulates the `init_secret`
+    for the current epoch and the path secret for the common ancestor.
+
+* Construct a Welcome message from the encrypted GroupInfo object and the
+  encrypted key packages.
 
 A member of the group applies a Commit message by taking the following steps:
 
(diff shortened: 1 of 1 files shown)


bifurcation:
@raphaelrobert - Curious if this seems implementable to you.  Clearly it's possible, since I did it in Go :)  But it also needs to make sense to other people.

draft-ietf-mls-protocol.md
```diff
@@ -1970,19 +1963,25 @@ message at the same time, by taking the following steps:
...
-    for the current epoch (see {{welcoming-new-members}})
+  * Encrypt the GroupInfo using the key and nonce derived from the `epoch_secret`
+    for the new epoch (see {{welcoming-new-members}})
 
-* For each new member in the group, compute an EncryptedKeyPackage object that
-  encapsulates the `init_secret` for the current epoch.  Construct a Welcome
-  message from the encrypted GroupInfo object and the encrypted key packages.
+* For each new member in the group:
+  * Identify the lowest common ancestor in the tree of the new member's
+    leaf node and the member sending the Commit
```
raphaelrobert: I think there might be more math involved here. Unless it's a group of two, the direct path of the new member should be blank until it intersects with the direct path of the adder (who populated their path with the Commit). In other words, the common ancestor node has to be resolved, leading to an array of resolved nodes. The new member should be one of these resolved nodes.
To me this looks like you end up KEMing to the new member's leaf node directly.

draft-ietf-mls-protocol.md
```diff
@@ -2767,37 +2765,13 @@ def copath(x, n):
...
-
-    if level(x) == 0:
-        return []
-
-    L = resolve(tree, left(x), n)
-    R = resolve(tree, right(x, n), n)
-    return L + R
+# The common ancestor of two leaves is the lowest node that is in the
+# lowest-level node that is in the direct paths of both leaves.
+def common_ancestor(x, y):
```
raphaelrobert: For simplicity I would just iterate over `parent()` until the value is equal.

bifurcation: That actually doesn't work, because the distance to the parent might not be equal.  Consider, for example, the three-member tree:

```
    ABC
   /   \
 AB     |
 / \    |
A   B   C
```

The common ancestor of A and B is `ABC = parent(parent(A)) = parent(C)`.

You would have to first compute the direct path of one node, then iterate `parent()` on the other until you end up in the direct path.  The formula here is a little bit of index magic, but I think it can be proven correct, and it's simpler/faster.

draft-ietf-mls-protocol.md
bifurcation: You're kind of making a point I should have made, namely that sending the path secret of the common ancestor as we are here is equivalent to sending the direct path -- in the direct path encryption, you would encrypt the path secret to the corresponding copath node, which would ultimately resolve to the leaf node (and possibly some others).

What this PR is doing is taking the KEMing to the new member's leaf and taking it out of the context of the DirectPath.  Where before you were doing an HPKE to the new member once for EncryptedKeyPackage and again in the DirectPath, now you just do it for the EncryptedKeyPackage.

raphaelrobert: Sure, I think that makes sense. 

But there might be another issue: I forgot that we don't blank the direct path of new members anymore and instead use `unmerged_leaves`. This list should now be shorter and only include nodes from leaf to common ancestor. This is however something every member in the group processing the `Add` has to calculate as it is non explicit.

draft-ietf-mls-protocol.md
raphaelrobert: True, my idea only works reliably on the left part of the tree. On the right part you'd have to also use `level()` to skip some nodes. If you're confident your math works, it might just be simpler to use that.

bifurcation: The algorithm passes tests on a non-trivial tree, so I have pretty good confidence that it works.  Proof sketch:

* Each subtree of the tree is identified by a common prefix
    * The root of the subtree is `<prefix>0<1 * level>`
    * The nodes within the subtree are `<prefix>0...0` to <prefix>1...10
* Therefore every common ancestor of A and B will have a prefix that is common to A and B, since a common ancestor is the root of a subtree containing both nodes
* This algorithm finds the longest prefix shared by A and B, then computes the root of the corresponding subtree

Of course, all of this is just a recommended implementation.  Implementations are free to use a more explicit algorithm like yours if they prefer.

draft-ietf-mls-protocol.md
bifurcation: > This list should now be shorter and only include nodes from leaf to common ancestor

This is already true.  Even before this PR, the DirectPath processing would reset the `unnmerged_leaves` arrays to `[]` for nodes in the direct path of the committer.

bifurcation: In any case, we should be clear that you reset `unmerged_leaves` whenever you write to a node.  I don't think that needs to be blocking on this PR.

draft-ietf-mls-protocol.md
```diff
@@ -1961,13 +1961,6 @@ message at the same time, by taking the following steps:
...
-* Create an initial, partial GroupInfo object reflecting the following values:
-  * Group ID: The group ID for the group
-  * Epoch: The epoch ID for the next epoch
-  * Tree: The group's ratchet tree after the commit has been applied
-  * Prior confirmed transcript hash: The confirmed transcript hash for the
-    current state of the group (not the provisional state)
-
 * Create a DirectPath using the new tree (which includes any new members).  The
   GroupContext for this operation uses the `group_id`, `epoch`, `tree`, and
   `prior_confirmed_transcript_hash` values in the initial GroupInfo object.
```
ericcornelissen: This change leaves a reference to a GroupInfo object that is not mentioned before. If I'm not mistaken, I think this can be rewritten as:

> Create a DirectPath using the new tree (which includes any new members). The GroupContext for this operation uses the `group_id`, `epoch`, and `tree` in the provisional GroupContext and the `confirmed_transcript_hash` for the current state of the group (not the provisional state).

On this note, I'm not entirely sure I understand why an entire GroupContext (specifically the `group_id`, `epoch`, and `confirmed_transcript_hash`) is needed to compute a DirectPath...

Issue #302, opened 2020-02-06
Use masking instead of AES-GCM for sender data


bifurcation:
Earlier, we had considered a "masking" approach for sender data, much like QUIC does for [encrypting packet sequence numbers](https://github.com/quicwg/base-drafts/blob/master/draft-ietf-quic-tls.md#header-protection-header-protect).  This approach would save 28 bytes of overhead (nonce + tag), but was abandoned because of uncertainty around the security of the masking approach.

Since then, Bellare et al. published ["Nonces are Noticed"](https://eprint.iacr.org/2019/624.pdf), which includes what is effectively a proof of the security properties of the QUIC scheme.  The QUIC scheme is essentially scheme HN1 of the paper, the main difference being that QUIC encrypt not the nonce directly, but a packet number which is used to compute the nonce.  Our application would be similar, since we would encrypt metadata used to encrypt the key and the nonce.

Given this new analysis, is it time to bring back masking?

bifurcation:
Fixed by #360

PR #303, opened 2020-02-06
Add some per-message entropy

### LINKED ISSUE #216: Address risk of re-using nonces after state loss
Nonces and keys are generated deterministically. If a client's state is rolled back to a previous version (because the process crashed, for instance), then the client is very likely to re-use the same key+nonce pair to encrypt a different application message. This breaks the confidentiality of the AEAD.

Consider using explicit nonces (even small ones will do) or SIV-like constructions.

bifurcation:
Fixes #216

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1584,12 +1584,17 @@ object of the following form:
 struct {
     uint32 sender;
     uint32 generation;
+    opaque reuse_guard[4];
 } MLSSenderData;
 ~~~~~
 
 MLSSenderData.sender is assumed to be a `member` sender type.  When constructing
 an MLSSenderData from a Sender object, the sender MUST verify Sender.sender_type
 is `member` and use Sender.sender for MLSSenderData.sender.
+
+The `reuse_guard` field contains a fresh random value used to avoid nonce reuse
+in the case of state loss or corruption, as described in
+{{content-signing-and-encryption}}.
 
 The Additional Authenticated Data (AAD) for the SenderData ciphertext
 computation is its prefix in the MLSCiphertext, namely:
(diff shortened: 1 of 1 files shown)


bifurcation:
cc @Bren2010

beurdouche:
Quick question without reading much, but why can't we reuse the sender data nonce which is always fresh to do this?

bifurcation:
You could do that; in fact you could just use that nonce for the content encryption as well as the sender data encryption, and forget about the key-schedule derived nonces as well.

The reason I didn't go down that path is that I'm generally leery of requiring fresh randomness in the protocol.  In particular, for the sender data, I think we should be moving toward masking instead of GCM (cf #302).  If we do masking, then we won't need to change this, and we'll have basically the minimum amount of fresh entropy needed.

beurdouche:
Nope you can't really do that. The xoring of the two is strictly stronger, even if your PRNG is bad, you'll benefit from the entropy of the deterministic nonce because of the current epoch secret to which many people have contributed to.

Until we do that move to masking, I think I would prefer to have the xor use the sender data nonce.

bifurcation:
I'm confused about what you're talking about.  In the current proposal:

```
aead_nonce = key_sched_nonce XOR (reuse_guard || 0)
```

I agree that that's better than just using the `sender_data_nonce` for the reasons you say.  But I thought what you were proposing was :

```
aead_nonce = key_sched_nonce XOR sender_data_nonce
```

Either way, we're incorporating the key schedule nonce.  

How about the following proposal?
* For now: `aead_nonce = key_sched_nonce XOR sender_data_nonce`
* If we do masking later, we do something like what's in this PR

beurdouche:
> How about the following proposal?
> 
>     * For now: `aead_nonce = key_sched_nonce XOR sender_data_nonce` 
>     * If we do masking later, we do something like what's in this PR

Works for me !

bifurcation:
@Bren2010 if the revised scheme looks ok to you, I'll consider this ready to merge

Bren2010:
If I had a concern with the new scheme, it's that it's no longer resistant to nonce misuse. For example, consider somebody setting `sender_data_nonce = key_sched_nonce`.

bifurcation:
That would violate the requirement that the `sender_data_nonce` be fresh.  I've clarified that requirement in the latest commit.

Bren2010:
Putting a requirement in the spec is good. But that won't stop a nonce-misusing implementation from running, which was the point of deterministically generating the nonce in the first place.

bifurcation:
I'm not sure why you think a freshness requirement is less plausible for the `sender_data_nonce` than for the `reuse_guard`.  Neither is enforceable in the protocol.

image

Bren2010:
The value of `reuse_guard` is that it's only 4 bytes, so it doesn't mask the whole thing. Worst case, `reuse_guard` can reliably fix 32 bits, leaving only 64 bits of actual randomness in the nonce. Not good, but better than 0, and better than invoking a hash `H(key_sched_nonce | sender_data_nonce)` to actually mix things.

Bren2010:
The fact that `reuse_guard` is not called a nonce and is a different size also makes me feel better about it not being reused

beurdouche:
@bifurcation @Bren2010 I don't have a strong opinion here, I was just pointing that we already has a fresh value. I can see why Brendan prefers the smaller guard using a fresh distinct value, so I'll support this or any variant for which you both agree ; )

bifurcation:
Reverted to the `reuse_guard` approach, which it seems everyone is OK with.

PR #305, opened 2020-02-10
Update HKDFLabel variable to match naming conventions

GaPhil:
This PR changes the variable name `HkdfLabel` to `HKDFLabel` in order to match the previously used naming conventions.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1994,7 +1994,7 @@ message at the same time, by taking the following steps:
 * For each new member in the group:
   * Identify the lowest common ancestor in the tree of the new member's
     leaf node and the member sending the Commit
-  * Compute the path secret corresponding to the commonn ancestor node
+  * Compute the path secret corresponding to the common ancestor node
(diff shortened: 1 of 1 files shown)


beurdouche:
Hi Philip @GaPhil  !
This is following the conventions we used in RFC 8446 so I think we should just keep that... I'll let Richard @bifurcation close this if he agrees with me but anyway, thank you for having a look! : )
B.

GaPhil:
Sure, noted.

However, one might consider whether one wishes to maintain the same conventions for acronyms in this document. See `HPKECiphertext` and `MLSPlaintextCommitAuthData` as opposed to `HpkeCiphertext` and `MlsPlaintextCommitAuthData`.

bifurcation:
I actually agree with @GaPhil here.  It's also more like Go than Java, so let's go with the more modern language :)

beurdouche:
Well Phillip's point was certainly valid. The one about programming language is completely irrelevant.

bifurcation:
I agree that it's irrelevant.  Just kidding.

PR #307, opened 2020-02-13
DE-related text

seanturner:
Here's to hoping I did this merge properly.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2430,26 +2430,26 @@ Initialization keys are intended to be used only once and then
 deleted. Reuse of init keys can lead to replay attacks.
 
 
-# IANA Considerations
+# IANA Considerations {#iana-considerations}
 
-This document requests the creation of the following new IANA registries:
+This document requests the creation of the following new IANA
+registries: MLS Ciphersuites {{ciphersuites}}. All of these
+registries should be under a heading of "Message Layer Security",
+and assignments are made via the Specification Required policy
+{{!RFC8126}}. See {{de}} for additional information about the
+MLS Designated Experts (DEs).
 
-* MLS Ciphersuites
-
-All of these registries should be under a heading of "Message Layer Security",
-and administered under a Specification Required policy {{!RFC8126}}.
-
-## MLS Ciphersuites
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2430,26 +2430,26 @@ Initialization keys are intended to be used only once and then
 deleted. Reuse of init keys can lead to replay attacks.
 
 
-# IANA Considerations
+# IANA Considerations {#iana-considerations}
 
-This document requests the creation of the following new IANA registries:
+This document requests the creation of the following new IANA
+registries: MLS Ciphersuites {{ciphersuites}}. All of these
```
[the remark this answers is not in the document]
bifurcation: As above, just use `{{mls-ciphersuites}}` and there's no need for an explicit tag.

draft-ietf-mls-protocol.md
```diff
@@ -2430,26 +2430,26 @@ Initialization keys are intended to be used only once and then
...
+Ciphersuite names follow the naming convention:
 
 ~~~
    CipherSuite MLS_LVL_KEM_AEAD_HASH_SIG = VALUE;
 ~~~
 
-Where VALUE is a is represented as two 8bit octets:
+Where VALUE is represented as two 8bit octets:
 
 uint8 CipherSuite[2];
```
bifurcation: This needs to be guarded with `~~~`

draft-ietf-mls-protocol.md
```diff
@@ -2479,19 +2479,60 @@ defined in {{RFC5116}}. AEAD_CHACHA20_POLY1305 is defined
...
 Values with the first byte 255 (decimal) are reserved for Private Use.
 
-New cipher suite values are assigned by IANA as described in
+New ciphersuite values are assigned by IANA as described in
 {{iana-considerations}}.
 
+## MLS Designature Expert Pool {#de}
+
+[[ OPEN ISSUE: pick DE mailing address.
+Maybe mls-des@ or mls-de-pool. ]]
```
bifurcation: Suggest `mls-ciphersuites@`

seanturner: But, we're going to end up with more than one registry (e.g., extensions) so I was going for the more general name.

PR #308, opened 2020-02-27
Remove nonce from SenderData AAD.

Bren2010:
With any AEAD, the nonce is necessarily already authenticated. Also, the name of the struct was changed to be more clear.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1580,8 +1580,7 @@ struct {
     uint64 epoch;
     ContentType content_type;
     opaque authenticated_data<0..2^32-1>;
-    opaque sender_data_nonce<0..255>;
-} MLSCiphertextSenderDataAAD;
+} MLSSenderDataAAD;
 ~~~~~
 
 When parsing a SenderData struct as part of message decryption, the


brittahale:
Does this force use of an implicit nonce such as in TLS, vice an explicit nonce such as in DTLS?

Bren2010:
Sender data has always been encrypted with an explicit nonce, and the explicit nonce is still send in MLSCiphertext. This just removes the explicit nonce from the AAD.

bifurcation:
I think @beurdouche proposed this originally, with the idea that the AAD at each point would include everything above it.  But I agree that it doesn't seem very useful.  I'm OK to merge if other folks are.

beurdouche:
The thing is designed so that the entire prefix of the AEAD ciphertext is authenticated.
Screenshot 2020-03-04 at 19 01 50
I would prefer to keep it that way...

bifurcation:
@Bren2010 unless you think this is really important, I'm inclined to just close this PR.

Bren2010:
@beurdouche Put "sender nonce" under "Encrypted Sender Data" because the nonce is part of the ciphertext. Then there's no gap in your picture, and "Encrypted Sender Data" will fit within the bounds of the arrow.

beurdouche:
That was not meant to be merged, that was meant for raphael and I to review. So, I’ll revert that until we can do this.

PR #311, opened 2020-03-03
Review of section 7.

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -744,8 +739,8 @@ above:
 ## Synchronizing Views of the Tree
 
 The members of the group need to keep their views of the tree in
-sync and up to date.  When a client proposes a change to the tree
-(e.g., to add or remove a member), it transmits a handshake message
+sync and up to date.  When a client commits a change to the tree
+(e.g., to add or remove a set of members), it transmits a handshake message
```
bifurcation: This is just an example, so it doesn't cost any generality just to say "a member".  And I think it reads more clearly.

PR #320, opened 2020-03-10
Use node type instead of leaf index in tree hashes.

Bren2010:
This is more friendly with recursive implementations of Ratchet Trees and is more similar to the construction used in similar places, like RFC 6962.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1173,7 +1173,7 @@ structure is used:
 
 ~~~~~
 struct {
-    uint32 node_index;
+    NodeType node_type = parent;
     optional<ParentNode> parent_node;
     opaque left_hash<0..255>;
     opaque right_hash<0..255>;
@@ -1186,7 +1186,7 @@ a leaf node, the hash of a `LeafNodeHashInput` object is used:
 
 ~~~~~
 struct {
-    uint32 leaf_index;
+    NodeType node_type = leaf;
     optional<KeyPackage> key_package;
 } LeafNodeHashInput;
 ~~~~~


bifurcation:
We had something like that before, and I think some folks liked having the index better because, for example, blank leaves at different positions have different hashes.

Bren2010:
The change was made in #254 with no apparent discussion. Not sure what you mean wrt to blank nodes -- I don't believe this is any more or less malleable than the current scheme.

beurdouche:
There was discussion. The security proofs rely on being able to disambiguate between different empty leaves or intermediate nodes when you don't balance the levels, and the type is not enough to do that.

Bren2010:
Sorry, I still don't see where the ambiguity is?

bifurcation:
@Bren2010  You said in one of the issue discussions that this is to be friendlier to recursive implementations.  But it seems like as long as your recursive function knows (1) how big the tree is and (2) where it is in the tree, then you can include the node index with no problem.  This is in fact [how the Go implementation does it](https://github.com/bifurcation/mls/blob/master/ratchet-tree.go#L844).

Bren2010:
This implementation doesn't have a recursive structure, it [uses a slice of nodes](https://github.com/bifurcation/mls/blob/master/ratchet-tree.go#L378) and the first thing you do is `len(t.Nodes)`. A recursive implementation would be defined like:
```
type Node struct {
    left, right *Node
}
```

In which case, yeah, a method like `func (n *Node) Hash` wouldn't know how big its subtree is, or whether it's an intermediate or the root.

Bren2010:
I don't believe you need that information to make anything in the *ratchet tree* work. So you're arguing that we should include unnecessary information in one component of the protocol, because we happen to have that information somewhere else?

PR #327, opened 2020-03-28
Fix bugs in tree math and cleanup docs.

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -2883,93 +2882,97 @@ def level(x):
...
         return right(p, n)
-    elif x > p:
+    else:
         return left(p)
 
-    return p
-
-# The direct path of a node, ordered from the root
-# down, not including the root or the terminal node
+# The direct path of a node, ordered from leaf to root.
```
bifurcation, line 2943:
```suggestion
# The direct path of a node, ordered from leaf to root,
# including the leaf but not the root.
```

bifurcation, line 2943:
```suggestion
# The direct path of a node, ordered from leaf to root,
# including the root but not the leaf.
```

draft-ietf-mls-protocol.md
```diff
@@ -2883,93 +2882,97 @@ def level(x):
...
 def root(n):
     w = node_width(n)
     return (1 << log2(w)) - 1
 
-# The left child of an intermediate node.  Note that because the
-# tree is left-balanced, there is no dependency on the size of the
-# tree.  The child of a leaf node is itself.
+# The left child of an intermediate node. Note that because the tree is
+# left-balanced, there is no dependency on the size of the tree. The
+# child of a leaf node is itself.
```
mheinzel, line 2896:
```suggestion
# left-balanced, there is no dependency on the size of the tree.
```

To stay consistent with the code change.

draft-ietf-mls-protocol.md
```diff
@@ -2883,93 +2882,97 @@ def level(x):
...
 # The common ancestor of two leaves is the lowest node that is in the
 # lowest-level node that is in the direct paths of both leaves.
-def common_ancestor(x, y):
-    xn, yn = x, y
-    k = 0
-    while xn != yn:
-       xn, yn = xn >> 1, yn >> 1
-       k += 1
-    return (xn << k) + (1 << (k-1)) - 1
+def common_ancestor(x, y, n):
```
bifurcation, line 2967: If you're going to the semantic clarity route, you might as well go all the way.

```
def common_ancestor(x, y, n):
    dx = set(direct_path(x, n))
    dy = set(direct_path(y, n))
    dxy = dx & dy
    if len(dxy) == 0:
        raise Exception('failed to find common ancestor')

    return min(dxy, key=level)
```

I would prefer we also provide the old, direct version as an alternative implementation.  It is index magic, but it's not really any more magical than the stuff above.

```
# The same function as common_ancestor, but using the indices 
# directly instead of constructing the direct paths
def common_ancestor_dir(x, y, n):
    ...
```

raphaelrobert: I also think we should have both implementations

draft-ietf-mls-protocol.md
Bren2010: This is redundant with the definition of direct path

draft-ietf-mls-protocol.md
Bren2010: The code you've given is not correct, `common_ancestor(0, 7, 11) = 15 (should be 7)`.
The old code is also not correct, `common_ancestor(0, 7, 11) = 3 (should be 7)`, which is why I removed it.

bifurcation: The semantic code fails when x == y because the leaf is not in the direct path; the direct code fails when one node is in the direct path of the other.  I had not worried about these because MLS only ever needs to call `ancestor(x, y)` with x and y  leaf nodes and x != y.  But both are easy to fix.  See fixed code and validation here:

https://repl.it/@RichardBarnes/ImmediateCrookedScience

draft-ietf-mls-protocol.md
```diff
@@ -2883,89 +2882,107 @@ def level(x):
...
+    if len(dxy) == 0:
+        raise Exception('failed to find common ancestor')
+
+    return min(dxy, key=level)
+
 # The common ancestor of two leaves is the lowest node that is in the
 # lowest-level node that is in the direct paths of both leaves.
-def common_ancestor(x, y):
+def common_ancestor_direct(x, y, _):
+    # Handle cases where one is an ancestor of the other
```
raphaelrobert, line 2978: How can a leaf be the ancestor of another leaf?

Bren2010:
@bifurcation Merge!

yihuang:
```
The basic
rule is that the high-order bits of parent and child nodes have the
following relation (where x is an arbitrary bit string):

    parent=01x => left=00x, right=10x
```

I think it should change like this: "where x is an arbitrary bit string" => "where x is an sequence of one bits".

Bren2010:
You can have cases where x isn't all ones, for example when the left and right children are intermediate nodes

draft-ietf-mls-protocol.md
```diff
@@ -2883,89 +2882,107 @@ def level(x):
...
-# The parent of a node.  As with the right child calculation, have
-# to walk back until the parent is within the range of the tree.
+# The parent of a node. As with the right child calculation, we have to
+# walk back until the parent is within the range of the tree.
 def parent(x, n):
     if x == root(n):
-        return x
+        raise Exception('root node has no parent')
 
     p = parent_step(x)
```
yihuang, line 2929: If passed in x is larger than `node_width(x)`, the loop might not finish.

draft-ietf-mls-protocol.md
```diff
@@ -2883,89 +2882,107 @@ def level(x):
         k += 1
     return k
 
-# The number of nodes needed to represent a tree with n leaves
+# The number of nodes needed to represent a tree with n leaves.
 def node_width(n):
     return 2*(n - 1) + 1
```
yihuang, line 2887:
```suggestion
    if n == 0:
        return 0
    else:
        return 2*(n - 1) + 1
```

draft-ietf-mls-protocol.md
```diff
@@ -2883,89 +2882,107 @@ def level(x):
...
-    d = dirpath(x, n)
-    if x != sibling(x, n):
-        d.append(x)
+    if x == root(n):
+        return []
 
+    d = direct_path(x, n)
+    d.insert(0, x)
+    d.pop()
     return [sibling(y, n) for y in d]
```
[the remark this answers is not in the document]
Bren2010: d doesn't contain the root because of the last d.pop()

draft-ietf-mls-protocol.md
Bren2010: We don't assert anywhere else, probably fine

PR #331, opened 2020-04-27
Make ratcheting optional for Adds

bifurcation:
When we converted to the Proposal/Commit scheme, one of the consequences was that Adds became O(N) instead of O(1), since they had to Committed and commits encrypt the committer's direct path to the remainder of the tree.  This is ~log(N) in the nice case and ~N in the worst case.  At the time, we noted that one could special-case Add-only Commits in the future and make the update to the direct path optional in that case.

This PR implements that suggestion.  The leaf KeyPackage for the Commit is moved into the DirectPath struct, and the whole DirectPath is made optional.  The optional value MUST be provided for:

* A Commit committing an update or remove
* A "just update my path" Commit covering no proposals

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -139,6 +139,10 @@ shared keys with costs that scale as the log of the group size.
 
 RFC EDITOR PLEASE DELETE THIS SECTION.
 
+draft-10
+
+- Re-enable constant-time Add (\*)
+
 draft-09
 
 - Remove blanking of nodes on Add (\*)
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2060,9 +2060,10 @@ message at the same time, by taking the following steps:
 * For each new member in the group:
   * Identify the lowest common ancestor in the tree of the new member's
     leaf node and the member sending the Commit
-  * Compute the path secret corresponding to the common ancestor node
+  * If the `path` field was populated above: Compute the path secret
+    corresponding to the common ancestor node
   * Compute an EncryptedGroupSecrets object that encapsulates the `init_secret`
-    for the current epoch and the path secret for the common ancestor.
+    for the current epoch and the path secret (if present).
 
 * Construct a Welcome message from the encrypted GroupInfo object and the
   encrypted group secrets.
@@ -2167,8 +2168,12 @@ struct {
 } GroupInfo;
 
 struct {
-  opaque epoch_secret<1..255>;
   opaque path_secret<1..255>;
+} PathSecret;
+
+struct {
+  opaque epoch_secret<1..255>;
+  optional<PathSecret> path_secret;
 } GroupSecrets;
 
 struct {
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2021,17 +2022,22 @@ message at the same time, by taking the following steps:
...
+   * Assign this DirectPath to the `path` field in the Commit.
 
    * Apply the DirectPath to the tree, as described in
      {{synchronizing-views-of-the-tree}}. Define `commit_secret` as the value
      `path_secret[n+1]` derived from the `path_secret[n]` value assigned to
      the root node.
 
+* If not populating the `path` field: Set the `path` field in the Commit to the
+  null optional.  Define `commit_secret` as the all-zero vector of the same
+  length as a `path_secret` value would be.
```
raphaelrobert, line 2039: Why not set `commit_secret` to zero length vector? Seems easier to me.

kkohbrok:
So we _do_ go forward one epoch, but with an all-zero commit secret. Do I understand that correctly?

draft-ietf-mls-protocol.md
bifurcation: That would also work for me.  I just copied what TLS does, e.g., for missing PSK inputs.

bifurcation:
@kkohbrok That's correct.  The idea is that you don't need to send the current group any additional entropy, you just need them to hash forward.

Note that this could change the FS/PCS guarantees available.  But I don't think we ever proposed that Add would cause any FS/PCS changes to the group, so maybe this doesn't matter.

kkohbrok:
Add should give FS guarantees in so far as that the new member shouldn't be able to decrypt old messages or derive old keys. But I think ratcheting forward should do the trick. We should note, though, that this breaks the (implicit) assumption that a commit always introduces fresh key material to the group. As far as I can see, that's not breaking anything right now, but we should keep it in mind.

draft-ietf-mls-protocol.md
```diff
@@ -2002,11 +2002,12 @@ The Commit MUST NOT combine proposals sent within different epochs. In the event
...
-participants are no longer expected to have access to the Commit message adding
-them to the group.  It might be we need to re-introduce this assumption, though
-it seems like the information confirmed by the welcome_info_hash is confirmed at
-the next epoch change anyway. ]]
+The `path` field of a Commit message MUST be populated if the Commit covers at
+least one Update or Remove proposal, i.e., if the length of the `updates` or
+`removes vectors is greater than zero.  The `path` field MUST also be populated
+if the Commit covers no proposals at all (i.e., if all three proposal vectors
+are empty).  The `path` field MAY be omitted if the Commit covers only Add
+proposals.
```
bifurcation, line 2010: TODO: Add pseudocode here to express the algorithm.

bifurcation:
Note: The latest commit fixes a gap that was discovered in [implementing this PR in Go](https://github.com/cisco/go-mls/pull/69).

draft-ietf-mls-protocol.md
```diff
@@ -2076,21 +2095,28 @@ A member of the group applies a Commit message by taking the following steps:
   applied at the leftmost unoccupied leaf, or appended to the right edge of the
   tree if all leaves are occupied.
 
-* Process the `path` value using the ratchet tree the provisional GroupContext,
-  to update the ratchet tree and generate the `commit_secret`:
+* Verify that the `path` value is populated if either of the `updates` or
```
raphaelrobert, line 2098: I think we could be more flexible here: Since the path value MAY be omitted if the Commit only contains Adds, it MAY also not be omitted.
The sender should be able to choose whether they want to use the commit to also do an update and the receivers should simply process it.

bifurcation: I think this is already the case, but I added a bit more text to clarify

PR #332, opened 2020-05-05
adding a table of contents

bast:
We did this together with @chrisbrzuska live on Twitch.


raphaelrobert:
The table of contents is auto-generated in the text format, there is no need to add it:

https://www.ietf.org/id/draft-ietf-mls-protocol-09.txt

bast:
OK. Thank you for the clarification!

PR #333, opened 2020-05-05
add code highlighting for the python code part

bast:


raphaelrobert:
Code highlighting wouldn't work in the text format.

bast:
OK I considered only the rendered Markdown page on GitHub but forgot that the webpages are generated and possibly does not work there. Where are the generated webpages? Feel free to close it.

PR #336, opened 2020-05-16
PSK Injection, Group recovery, Re-Init, Sub-group Branching

br-hale:
Changes to key schedule and corresponding directions on use of PSKs for group recovery, re-initialization, sub-group branching, and external PSK injection.


draft-ietf-mls-protocol.md
```diff
@@ -2158,7 +2413,8 @@ struct {
 struct {
   opaque epoch_secret<1..255>;
   opaque path_secret<1..255>;
-} GroupSecrets;
+  optional<PSKId> psk<1..2^32-1;
+} KeyPackage;
```
raphaelrobert, line 2417: There already is a struct called `KeyPackage`

kkohbrok: Hm. Not sure how that got there. I don't think we intended renaming the struct. The change was supposed to be that the `psk`s are included in the Welcome message.

draft-ietf-mls-protocol.md
```diff
@@ -1353,66 +1353,131 @@ Given these inputs, the derivation of secrets for an epoch
...
+PSKLabel(psktype) =
+  select (psktype) {
+    case re-initialization: "re-inizialization";
+    case branch: "branch";
+    case recovery: "recovery";
+    case external: "external";
+}
+
+Derived_PSK =
+  Derive-Secret(PSK, Hash(PSKLabel(psktype) || psk_nonce))
```
chris-wood, line 1468: Over in TLS, we worked on an interface for injecting PSKs:

   https://tools.ietf.org/html/draft-ietf-tls-external-psk-importer-05

We concluded binding the imported PSK to the target protocol and KDF was sufficient. Maybe the same thing should be done here, i.e., to ensure that an injected PSK for MLS v1.0 does not collide with one injected for a different version?

draft-ietf-mls-protocol.md
```diff
@@ -1810,11 +1906,68 @@ followed by reconstructing the Add and Commit messages and verifying that the
...
+of maintained `recovery_secret` secrets and the group ID of the group they want
+to re-initialize. They also must include a nonce to avoid key re-use.
+
+~~~~~
+struct {
+  PSKType psktype;
+  select (psktype) {
+
+    case external:
+      opaque psk_id<0..255>;
```
chris-wood, line 1934: Are "external" PSKs fixed? Would it make sense for them to also include an epoch, to allow applications to roll them forward for some FS?

br-hale: External PSKs are required to be of an "equivalent security level" (although that terminology was added to the draft a long time ago and not yet defined). As such, any FS guarantees or other requirements must be met before being injected. It does not make sense to ratchet external PSK forward per MLS spec, as then we would be trying to correlate the time of existence of the PSK to MLS (i.e. parallel ratcheting); in that case the PSK is no longer external to the protocol but somehow correlated to it. The nonce prevents reuse, but any additional guarantees should be met by the application, not the protocol (going back to meeting requirements before being injected).

draft-ietf-mls-protocol.md
br-hale: Adding the MLS version number to the PSK type sounds like a solid addition.

draft-ietf-mls-protocol.md
chris-wood: I’m not sure I agree. You can certainly do the ratcheting outside of the protocol, but you can also do it inside. For example, send an epoch and psk_id, and hash forward the external PSK corresponding to the ID based on the epoch. Some applications might not be able to augment their external PSK provision machinery to ratchet, so handling it in the protocol can be appealing. Food for thought.

kkohbrok: So the proposal would be to extend MLS to allow ratcheting of "external" PSKs that continuously get injected into the protocol?

It's interesting, but it seems to me like we'd be creating a degree of redundancy here. MLS doesn't have very good FS capabilities (although they can be improved by something like RTreeKEM), so it's tempting to create a secondary mechanism, i.e. ratcheting a PSK along the epoch-chain. However, then the FS guarantees hinge on the FS properties of whatever means was used to distribute the PSK. So what is a scenario where the means of PSK distribution gives you better FS than MLS? If you have this really good out-of-band distribution channel for PSKs, do you really still need MLS?

I don't think it's a bad idea, but I'm not sure how often it will be used and if it warrants the added complexity in the protocol. But that's maybe an issue we should discuss on the list or in the next virtual interim? 

In any case, I think it's a good idea that's worth thinking about and that might match the use case of a number of people, but it's relatively orthogonal enough to our PR. Maybe it would be worth a new github issue?

bifurcation:
I had some long-form comments on this PR, so I sent them to the mailing list.

draft-ietf-mls-protocol.md
```diff
@@ -1279,71 +1279,145 @@ Given these inputs, the derivation of secrets for an epoch
...
-| `exporter_secret`     | "exporter"    |
-| `confirmation_key`    | "confirm"     |
+| Secret                  | Label         |
+|:------------------------|:--------------|
+| `sender_data_secret`    | "sender data" |
+| `handshake_secret`      | "handshake"   |
+| `application_secret`    | "app"         |
+| `exporter_secret`       | "exporter"    |
+| `confirmation_key`      | "confirm"     |
+| `confirmation_key`      | "confirm"     |
```
ericcornelissen, line 1321: Seems like the `confirmation_key` got duplicated accidentally

```suggestion
| `confirmation_key`      | "confirm"     |
```

kkohbrok: Whoops you're right. For some reason I can't commit your suggestion, but I'll change the PR accordingly.

draft-ietf-mls-protocol.md
```diff
@@ -1279,71 +1279,145 @@ Given these inputs, the derivation of secrets for an epoch
...
+    members wish to branch based on the current group state.
+
+The injection of one or more PSKs into the Key Schedule can be signalled in two
+ways: By including a `pre_shared_keys` extension either in a Commit message or
+in the `GroupSecrets` object of a Welcome message.
+
+~~~~~
+enum {
+  external(0),
+  re-initialization(1),
```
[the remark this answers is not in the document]
br-hale: Dash removed to address first comment.

For second comment: a single resumption case does not cover key differences between the three options here, and mixing them into one (which was considered at length) causes issues down the line. In reinit, the original group must be fully terminated as a requirement, which is not the case in branching and recovery. Branching "copies" a subset of the current group into a new one, and it is essential that members know that this is not done with recovery (where a member is added back in, but there is no divergence in history). If the PSKtype does not signal distinctions sufficiently well, then there will likely be group authentication attacks over the history of group evolution (i.e. disagreement on what has occurred).

draft-ietf-mls-protocol.md
```diff
@@ -1279,71 +1279,145 @@ Given these inputs, the derivation of secrets for an epoch
...
 
+A PSK may also be used within MLS in the following cases:
+
+  - Re-Initialization: If, during the lifetime of the group, a change in the
+    fixed group parameters becomes necessary, e.g. if the ciphersuite used
+    by the group is deprecated or if the protocol version should be upgraded,
+    a PSK can be used to reboot the group with the desired parameters.
+
+  - Recovery: If the group state of one or more members of the group deviates
+    from the rest, they can be re-added to the group securely.
```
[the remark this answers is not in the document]
br-hale: Good observation. Clarification added.

draft-ietf-mls-protocol.md
```diff
@@ -1279,71 +1279,145 @@ Given these inputs, the derivation of secrets for an epoch
...
+the PSK security expires before the next update (shorter PSK lifetime than
+update), does that constitute a weaker security level]]
+
+[[Open ISSUE: Define "security level", and what it means to match the
+security level of the ciphersuite used in the group.]]
+
 <!-- OPEN ISSUE: We have to decide if we want an external coordination
 via the application of a Handshake proposal. -->
 
+A PSK may also be used within MLS in the following cases:
```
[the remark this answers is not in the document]
br-hale: Addressed.

draft-ietf-mls-protocol.md
```diff
@@ -1279,71 +1279,145 @@ Given these inputs, the derivation of secrets for an epoch
...
+} PSKType;
+
+struct {
+  PSKType psktype;
+  select (psktype) {
+
+    case external:
+      opaque psk_id<0..255>;
+
+    default:
```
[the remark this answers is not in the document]
br-hale: Addressed, according to what you seem to be requesting.

draft-ietf-mls-protocol.md
```diff
@@ -1279,71 +1279,145 @@ Given these inputs, the derivation of secrets for an epoch
...
+includes them in the key schedule in the order listed in the commit. For
+internal PSKs, the psk is defined as the `recovery_secret` of the group and
+epoch specified in the `PreSharedKey` object. Finally, the `psk_secret` to be
+included in the Key Schedule is derived as follows (where `n` is the number of
+PSKs):
+
+~~~~~
+struct {
+    PSKType type;
+    PreSharedKey(type);
```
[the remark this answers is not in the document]
kkohbrok: Good point. Should be fixed now.

draft-ietf-mls-protocol.md
```diff
@@ -1867,6 +2019,42 @@ A member of the group applies a Remove message by taking the following steps:
...
+    PreSharedKeyID psk;
+} PSK;
+~~~~~
+
+When processing a commit message including one or more PSK proposals, a group
+member MUST derive the `psk_secret` for the inclusion in the Key Schedule as
+described in {{pre-shared-keys}}, where the order of the PSKs corresponds to the
+order of the PSK proposals in the commit.
+
+### ReInit
```
bifurcation, line 2038: Honestly, I'm not sure how much value this adds.  The new group already has a pointer to the old group (by way of the PSK ID).  The best that this can do is cause folks not to use the old group any more, and it seems like doing that at the application layer is sufficient.  The only thing doing it in MLS adds is that it gets into the transcript -- and by definition, the transcript doesn't matter after this!

If we're going to have a message here, I would propose we do it a bit differently:

* Use a different message type, parallel to Proposal or Commit.  Commit advances the group to a new epoch; this operation does not.
* ... and explain that this needs the same sequencing as Commit.
* Have the semantic be just "close", with an optional forward pointer.
* For the forward pointer, just use a hash of the Welcome message.

```
opaque WelcomeID<0..255>; 

struct {
    optional<WelcomeID> welcome_hash;
} Close;
```

draft-ietf-mls-protocol.md
```diff
@@ -2557,11 +2807,11 @@ participant D:
     /     \
    E       F
   / \     / \
-A0  B0  C0  D0 -+- KD0
+A0  B0  C0  D0 -|- KD0
```
bifurcation, line 2810: Please revert these unrelated changes.

draft-ietf-mls-protocol.md
```diff
@@ -1290,71 +1290,159 @@ Given these inputs, the derivation of secrets for an epoch
...
 
+<!-- OPEN ISSUEClarify lifetime vs security level mandated above. E.g. if
+the PSK security expires before the next update (shorter PSK lifetime than
+update), does that constitute a weaker security level -->
+
+<!-- OPEN ISSUE Define "security level", and what it means to match the
+security level of the ciphersuite used in the group. -->
+
 <!-- OPEN ISSUE: We have to decide if we want an external coordination
 via the application of a Handshake proposal. -->
```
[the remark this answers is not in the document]
br-hale: Actually not -- the "security level" is mentioned at different points in the document and still requires clarification as it is never defined. 

For example, what does it mean for a PSK to be of the same security level as the intended group? In reality, we need to match intended PCS and FS levels, and that is determined by how often keys are ratcheted (including signature keys). If there is a mis-match with incoming external PSKs, then the value of use of a PSK may be minor. If the external PSK was a TLS exporter key for example, then its PCS/FS guarantees are partially dependent on its freshness. 

If deniable signature keys are ever considered, then PSKs should also not undermine that "security level".

Security level explanation is something probably belongs in the architecture doc.

draft-ietf-mls-protocol.md
```diff
@@ -1290,71 +1290,159 @@ Given these inputs, the derivation of secrets for an epoch
...
 
+Each PSK in MLS has a type that designates how it was provisioned.
+External PSKs are provided by the application logic. Other resumption PSKs
+are derived from the MLS key schedule, and used in cases where it is
+necessary to authenticate a member's participation in a prior group state.
+In particular, in addition to external PSK types, a PSK derived from within MLS
+may be used in the following cases:
```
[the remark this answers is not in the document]
kkohbrok: Good point in that the types relate their use rather than their provisioning. What do you mean by "at this level"? We do agree that we need strong key separation between the different PSK types, right? That was one of the lessons of various TLS Key Schedule analysis. It's not quite clear to me what you are suggesting we change.

draft-ietf-mls-protocol.md
```diff
@@ -1669,17 +1785,17 @@ that the recipient of the message can use it to compute the nonce to be used for
 decryption.
 
 ~~~~~
-+-+-+-+-+---------...---+
++-|-|-|-|---------...---+
```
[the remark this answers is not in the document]
kkohbrok: Sorry about that. There must have been a mixup when merging. I'll fix it.

draft-ietf-mls-protocol.md
```diff
@@ -1750,11 +1866,45 @@ introduced.  The only choices the creator makes here are its own
 KeyPackage, the leaf secret from which the Commit is built, and the
 intermediate key pairs along the direct path to the root.
 
+## Group Creation from existing PSKs
```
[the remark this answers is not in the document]
kkohbrok: Works for me.

draft-ietf-mls-protocol.md
kkohbrok: I understand that you're reluctant to bloat the proposal space like this. However, I think a ReInit proposal serves two purposes. 1) It allows an external party (e.g. the DS) to indicate to the group that they should change/upgrade the group. A ReInit proposal authenticates the DS' intent and even the upgrade/change it wants to do.
2) It makes sure that the group agrees on the ReInit. The PSK in the Welcome message is derived from the epoch _after_ the commit, so everyone must have processed it and registered that the group should be closed.

I think both authentication and agreement are valuable properties for an action that has the power to change ciphersuite, version and extensions.

Other than potentially bloating the number of proposals we have: Why not have a ReInit proposal?

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
kkohbrok: Never mind. It was actually emacs, which "cleared up" all markdown tables upon saving the file -.-

draft-ietf-mls-protocol.md
kkohbrok: > As to "why not" -- typically we try not to put things in the protocol that don't do anything ;)

Point taken  :smile: 

Regarding the handshake message: I wasn't aware that they're also pretty extensible, so sure, that would work as well. Just to make sure I understand correctly: The server would be able to send the message themselves (if they're "preconfigured") and the message would be authenticated, right? Also, we would have to add the data fields to the handshake message instead like so:
```
struct {
    opaque group_id<0..255>;
    uint64 epoch;
    Sender sender;

    ContentType content_type;
        case application:

        case proposal:
          Proposal proposal;

        case commit:
          Commit commit;
          opaque confirmation_tag<0..255>;

        case close:
          ProtocolVersion version;
          CipherSuite cipher_suite;
          Extension extensions<0..2^32-1>;
    }

    opaque signature<0..2^16-1>;
} MLSPlaintext;
```
For maximum security guarantees, we could also add a `confirmation_tag` just like for the commit.

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
kkohbrok: I think we do to properly separate the keys depending on their use. This is comparable to TLS, where improper separation between application- and resumption PSKs led to problems down the road. However, we sill only need to derive a single value off of the Key Schedule, as the type is used later when the PSK gets injected into the Key Schedule.

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
kkohbrok: Tweaked the title a little. It's now "Linking a New Group to an Existing Group".

draft-ietf-mls-protocol.md
```diff
@@ -1750,11 +1866,45 @@ introduced.  The only choices the creator makes here are its own
...
+re-initialization of an existing group, or branching of a sub-group.
+
+Re-initialization of an existing group may be used, for example, to re-start the
+group based on the current group state but under a different ciphersuite.
+
+Branching may be used to bootstrap a new group consisting of a subset of
+current group members, based on the current group state.
+
+In both cases, the `psk_nonce` included in the `PreSharedKeyID` object must be a
+randomly sampled nonce to avoid key re-use.
```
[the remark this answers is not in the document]
kkohbrok: I've specified the nonce length to be `KDF.Nh` and added an open issue regarding the possibility to avoid the nonce altogether using the GroupId of the new group.

draft-ietf-mls-protocol.md
kkohbrok: Very good points. Using the Welcome message seems like an elegant way to link the groups.
Regarding ReInit vs. Replace, I have no strong opinion on that. Replace seems a little more direct, so I'll go with that for now. I'll make the changes either today or over the weekend so hopefully we can get the green light for a merge on Tuesday :-)

kkohbrok: I just noticed that if we use a Welcome message (or its hash) in the Replace message, then the Server won't be able to send it, because it lacks the necessary information to craft it. So until we find a better way of doing it, I'm going with the way I originally sketched. Let me know if you have another idea.

kkohbrok: If we go for a Handshake Message as opposed to a Proposal, it's also not clear, which group member sends the Welcome message for the new group, as the Close message can come from the Server itself. The Server could nominate a member, but it doesn't know if that member is online and so there is a risk that other members would then have to wait for that member to come online. The other possibility is to just have any member do it, which would create a race condition and possibly multiple follow-up groups.

 Do you have an idea on how to deal with that? Since there is otherwise not that much of a difference between Handshake Message and Proposal, I'm inclined to stick with a Proposal, but rename it to "Close" and make the content optional.

draft-ietf-mls-protocol.md
```diff
@@ -1290,70 +1290,150 @@ Given these inputs, the derivation of secrets for an epoch
...
+    from the rest, they can be re-added to the group such that participants who
+    were group members at the time of deviation and are still in the group can
+    verify that the added member was previously in the group.
+
+  - Branching: A PSK may be used to bootstrap a subset of current group
+    members into a new group. This applies if a subset of current group
+    members wish to branch based on the current group state.
+
+The injection of one or more PSKs into the key schedule is signaled in two
+ways: 1) as an extension to a Commit message that initiates a new epoch and 2)
```
bifurcation, line 1380: Not an extension, a new Proposal.  Seems like the content of the Proposal and the extension can be the same object?  Or since we're doing multiplicities here, the Welcome extension would have the union of the Proposals.

draft-ietf-mls-protocol.md
```diff
@@ -1867,6 +2018,43 @@ A member of the group applies a Remove message by taking the following steps:
...
+### PSK
+
+A PSK proposal can be used to request that a PSK be injected into the
+key schedule in the process of advancing the epoch.
+
+~~~~~
+struct {
+    PreSharedKeyID psk;
+} PSK;
+~~~~~
```
bifurcation, line 2030: Nit: I would prefer to spell out PreSharedKey.

Do we need a separate structure here, or could we re-use the PreSharedKeys struct defined above?

draft-ietf-mls-protocol.md
```diff
@@ -1867,6 +2018,43 @@ A member of the group applies a Remove message by taking the following steps:
...
+    opaque group_id<0..255>;
+    ProtocolVersion version;
+    CipherSuite cipher_suite;
+    Extension extensions<0..2^32-1>;
+} ReInit;
+~~~~~
+
+A member of the group applies a ReInit message by waiting for the committer to
+send the Welcome message and by checking that the `group_id` and the parameters
+of the new group corresponds to the ones specified in the proposal.
```
bifurcation, line 2055: Do we want to specify a PSK that carries over from the old group to the new group?  So you would define here something like: The new Welcome MUST have a PreSharedKey extension with a `reinit` entry pointing to the epoch initiated by the Commit that covers the ReInit proposal.

draft-ietf-mls-protocol.md
```diff
@@ -1933,6 +2121,19 @@ Update for the leaf if there are no Removes. If there are multiple Add proposals
 for the same client, the committer again chooses one to include and considers
 the rest invalid.
 
+If the committer recieved multiple PSK proposals, they MUST include all of them
+in an ordering of their choice.
```
bifurcation, line 2125: I would strike this.  It's up to the committer to decide what they commit.  For example, if someone proposes adding a nonsense PSK, that shouldn't get committed.

draft-ietf-mls-protocol.md
```diff
@@ -1933,6 +2121,19 @@ Update for the leaf if there are no Removes. If there are multiple Add proposals
...
+
+If there are one or more multiple ReInit proposals, the commiter MUST choose one
+where
+
+* the `group_id` is unique among its own groups,
+* each group member provides a KeyPackage that satisfies the parameters
+specified in the proposal and
+* the `mls_version` is greater or equal to that of the current group
+
+The committer MUST consider all other ReInit proposals invalid.
```
bifurcation, line 2135:
```suggestion
If there are multiple ReInit proposals during an epoch, the committer MUST choose exactly one to include in the Commit message.  The `version` field in the ReInit proposal MUST be no less than the version for the current group.  A client processing a Commit that covers multiple ReInit proposals or a ReInit proposal with a lower version MUST reject it as malformed.  The committer SHOULD choose a ReInit proposal with a unique `group_id` and parameters that are supported by all clients in the group.
```

draft-ietf-mls-protocol.md
```diff
@@ -1997,6 +2198,13 @@ message at the same time, by taking the following steps:
     be updated by the implementation to include any new proposals added by
     negotiated group extensions.
 
+  * PSK proposals are processed later when deriving the `psk_secret` for the Key
+    Schedule.
+
+  * ReInit proposals are not applied. They are simply present in the `proposals`
+    vector to indicate to other members if and how the group should be
+    re-initialized.
```
bifurcation, line 2206:
```suggestion
  * A ReInit proposal is applied by marking the group as "closed".  Further proposals and commits for this group MUST be rejected.  The group's only further purpose is to provide a PSK used by the successor group.
```

bifurcation: Actually, this probably goes in the receiver processing below...

draft-ietf-mls-protocol.md
```diff
@@ -2056,6 +2278,12 @@ A member of the group applies a Commit message by taking the following steps:
   public key from the credential stored at the leaf in the tree indicated by
   the `sender` field.
 
+* Verify that all PSKs specified in any PSK proposals in the `proposals` vector
+  are available.
+
+* Verify that only one ReInit proposal is present and that the `mls_version`
```
bifurcation, line 2284:
```suggestion
* Verify that at most one ReInit proposal is present and that the `mls_version`
```

bifurcation (commented):
This is getting real close.  Just a few things to tighten up.

draft-ietf-mls-protocol.md
```diff
@@ -1290,70 +1290,150 @@ Given these inputs, the derivation of secrets for an epoch
...
+~~~~~
+struct {
+    PreSharedKeyID;
+    uint16 index;
+    uint16 count;
+} PSKLabel;
+
+psk_input_[i] = KDF.Extract(0, psk_[i])
+psk_secret_[i] = ExpandWithLabel(psk_input_[i], "derived psk", PSKLabel, KDF.Nh)
+psk_secret     = psk_secret_[i] || ... || psk_secret_[n]
```
Bren2010, line 1432: Maybe for the sake of efficiency, concat all the `psk_[i]` before they're put into Extract?

draft-ietf-mls-protocol.md
```diff
@@ -1445,6 +1525,34 @@ the Group.
...
 
+## Recovery Keys
+
+The main MLS key schedule provides a `recovery_secret` which can provide extra
+security in some cross-group operations.
+
+The application SHOULD specify an upper limit on the number of past
+epochs for which the `recovery_secret` may be stored.
+
+There are three ways in which a `recovery_secret` can be used to re-initialize
```
Bren2010, line 1536:
```suggestion
There are three ways in which a `recovery_secret` can be used: to re-initialize
```

draft-ietf-mls-protocol.md
```diff
@@ -1445,6 +1525,34 @@ the Group.
...
+
+## State Authentication Keys
+
+The main MLS key schedule provides a per-epoch `authentication_secret`
+which MAY be used for authenticating the current group state. As with recovery
+keys, these are distinguished from exporter keys in that they have specific
+use affecting the MLS protocol.
+
+Authentication keys MAY be used to authenticate current or past group
+members out-of-band.
```
Bren2010, line 1554: I think you definitely need to elaborate on what this means / how it would be done?

draft-ietf-mls-protocol.md
```diff
@@ -1290,70 +1290,150 @@ Given these inputs, the derivation of secrets for an epoch
...
+  external(0),
+  reinit(1),
+  branch(2),
+  (255)
+} PSKType;
+
+struct {
+  PSKType psktype;
+  select (psktype) {
+
```
Bren2010, line 1395: - empty line

draft-ietf-mls-protocol.md
```diff
@@ -1290,70 +1290,150 @@ Given these inputs, the derivation of secrets for an epoch
...
+  - Re-Initialization: If, during the lifetime of the group, a change in the
+    fixed group parameters becomes necessary, e.g. if the ciphersuite used
+    by the group is deprecated or if the protocol version should be upgraded,
+    a PSK can be used to carry entropy from the old group forward into a new
+    group with the desired parameters.
+
+  - Recovery: If the group state of one or more members of the group deviates
+    from the rest, they can be re-added to the group such that participants who
+    were group members at the time of deviation and are still in the group can
+    verify that the added member was previously in the group.
```
Bren2010, line 1373: Please elaborate on how recovery happens. From skimming, it's not clear how I would do this

Bren2010 (changes requested):

draft-ietf-mls-protocol.md
kkohbrok: Right, it's a Proposal (again). Hm. Right now, the Proposal contains a single `PreSharedKeyID`, while the `Welcome` contains an array of `PreSharedKeyID`s. So the `Welcome` needs to contain the PSKs referenced by the Proposals in the order in which the individual Proposals are contained in the Commit. Is that what you meant?

draft-ietf-mls-protocol.md
kkohbrok: I'll change it such that it spells it out. As I mentioned in my comment above, the Proposal only contains a single `PreSharedKeyID`, while `PreSharedKeys` contains an array. We could change that, though. I don't have a strong opinion on this, so I'll leave it as it is for now.

draft-ietf-mls-protocol.md
kkohbrok: That's exactly what we want to do, so you're right indeed.

draft-ietf-mls-protocol.md
kkohbrok: Good point. That's a relic from a past design that actually supported recovery. I'll remove it, as it is not relevant anymore.

draft-ietf-mls-protocol.md
kkohbrok: Note, that it's not an extension in the `Welcome` anymore, but instead an optional `PreSharedKeys` field in the `GroupSecrets` object.

draft-ietf-mls-protocol.md
kkohbrok: Terminology is a bit tricky. We have "PSK"s (i.e. the concrete keys), but now we also have "PreSharedKey" which is the name of the Proposal. However, the Proposal in turn only contains `PreSharedKeyID`s, i.e. pointers to how the PSK can be derived.

draft-ietf-mls-protocol.md
kkohbrok: I've added another sentence or two on what this is supposed to achieve and how.

draft-ietf-mls-protocol.md
kkohbrok: I'm not sure what the assumptions would be under which we could prove that secure. Other than a random oracle that is. Since I don't think it's likely that people will inject more than one PSK at once very often, I'd rather leave this as it is.

bifurcation (approved):

draft-ietf-mls-protocol.md
```diff
@@ -1290,70 +1290,143 @@ Given these inputs, the derivation of secrets for an epoch
...
+    members into a new group. This applies if a subset of current group
+    members wish to branch based on the current group state.
+
+The injection of one or more PSKs into the key schedule is signaled in two ways:
+1) as a `psk` proposal and 2) in the `GroupSecrets` object of a Welcome message
+sent to new members added in that epoch.
+
+~~~~~
+enum {
+  external(0),
```
Bren2010, line 1380: Please make zero `reserved`

draft-ietf-mls-protocol.md
```diff
@@ -1290,70 +1290,143 @@ Given these inputs, the derivation of secrets for an epoch
...
+  reinit(1),
+  branch(2),
+  (255)
+} PSKType;
+
+struct {
+  PSKType psktype;
+  select (psktype) {
```
Bren2010, line 1388:
```suggestion
  select (PreSharedKeyID.psktype) {
```

draft-ietf-mls-protocol.md
```diff
@@ -1290,70 +1290,143 @@ Given these inputs, the derivation of secrets for an epoch
...
+
+    case reinit:
+      opaque psk_group_id<0..255>;
+      uint64 psk_epoch;
+
+    case branch:
+      opaque psk_group_id<0..255>;
+      uint64 psk_epoch;
+  }
+  opaque psk_nonce<0..255>;
```
Bren2010, line 1400: Never used

draft-ietf-mls-protocol.md
```diff
@@ -2156,6 +2402,7 @@ struct {
 struct {
   opaque joiner_secret<1..255>;
   optional<PathSecret> path_secret;
+  optional<PreSharedKeys> psks<1..2^32-1>;
```
Bren2010, line 2405:
```suggestion
  optional<PreSharedKeys> psks;
```

draft-ietf-mls-protocol.md
```diff
@@ -1290,70 +1290,143 @@ Given these inputs, the derivation of secrets for an epoch
...
+} PreSharedKeys;
+~~~~~
+
+On receiving a Commit with a `pre_shared_keys` extension or a GroupSecrets
+object with the psk field set, the receiving Client
+includes them in the key schedule in the order listed in the commit. For
+internal PSKs, the psk is defined as the `recovery_secret` of the group and
+epoch specified in the `PreSharedKeyID` object. Finally, the `psk_secret` to be
+included in the Key Schedule is derived as follows (where `n` is the number of
+PSKs):
```
Bren2010, line 1414: References Commit extensions

draft-ietf-mls-protocol.md
```diff
@@ -1290,70 +1290,143 @@ Given these inputs, the derivation of secrets for an epoch
...
+
+~~~~~
+struct {
+    PreSharedKeyID;
+    uint16 index;
+    uint16 count;
+} PSKLabel;
+
+psk_input_[i] = KDF.Extract(0, psk_[i])
```
Bren2010, line 1423:
```suggestion
psk_input_[i] = KDF.Extract(psk_[i], "")
```

draft-ietf-mls-protocol.md
```diff
@@ -1445,6 +1518,37 @@ the Group.
...
+
+There are three ways in which a `recovery_secret` can be used: to re-initialize
+the group with different parameters, to re-add lost group members or to create a
+sub-group of an existing group, as detailed in {{pre-shared-keys}}.
+
+Recovery keys are distinguished from exporter keys in that they have specific
+use inside the MLS protocol, whereas the use of exporter secrets may be
+decided by an external application. They are thus derived separately to avoid
+key material reuse resulting from poor exporter key management external to
+the protocol.
```
Bren2010, line 1537: Probably can get rid of the last sentence in this paragraph, it comes off as hostile

draft-ietf-mls-protocol.md
```diff
@@ -1445,6 +1518,37 @@ the Group.
...
+The main MLS key schedule provides a per-epoch `authentication_secret`
+which MAY be used for authenticating the current group state. As with recovery
+keys, these are distinguished from exporter keys in that they have specific
+use affecting the MLS protocol.
+
+Members of a group MAY compare their `authentication_secrets` out-of-band to
+authenticate that they share the same view of the group.
+
+If one of the parties is being actively impersonated by an attacker, their
+`authentication_secret` will differ from that of the other group members.
```
Bren2010, line 1550: Try to compress this section into one paragraph

Bren2010 (changes requested):

draft-ietf-mls-protocol.md
kkohbrok: It's used in the derivation of `psk_secret`. `PSKLabel` contains the `PreSharedKeyID`, which in turn contains the nonce.

draft-ietf-mls-protocol.md
kkohbrok: Good catch. I thought I'd gotten rid of all of that stuff.

draft-ietf-mls-protocol.md
kkohbrok: What's wrong with calling `Extract` like this?

draft-ietf-mls-protocol.md
kkohbrok: I've shortened it a little bit and left it with two paragraphs. It should be in keeping with the style of the rest of the spec now.

draft-ietf-mls-protocol.md
bifurcation: Nope, the IKM is on the right: `Extract(salt, ikm)`.  I agree this is unlikely to be an issue, but it seems more conservative to use the IKM slot.

https://github.com/cfrg/draft-irtf-cfrg-hpke/blob/master/draft-irtf-cfrg-hpke.md#cryptographic-dependencies

kkohbrok: Right, sorry. That's what I meant to say, just got the directions confused. Anyway, we're calling with the psk on the right right now, so we should be good.

bifurcation: Ah right.  I got mixed up on which was the thing in the PR and which was Brendan's edit.  I agree that the PR has the PSK in the IKM slot.

bifurcation (approved):

PR #337, opened 2020-05-17
Replace DKDF by n-PRF in key schedule; minor fix

chrisbrzuska:
We propose to replace the two-stage Extract-Expand-
Extract-Expand method to derive a single key from
multiple input keys by an Expand-XOR-Expand method.

As a small fix we added the Context argument to the
Derive-Secret function and pass it on to the
HKDF-Extract-Label function. In the key schedule
graph the function was called with the Context
argument, but it wasn't present in the function
definition of Derive-Secret.

Motivation for the Expand-XOR-Expand method:
- Extract was not used before, KDF was used as DPRF only.
- Previous construction relied on ad hoc assumption on DPRF/DKDF
- Previous construction required uniqueness of secrets to obtain distinct and independent keys (can be fragile when keys are under adversarial control)
- n-PRF reduces reliance on uniqueness of secrets and instead relies on the uniqueness of a public value under the protocol designer's control

Todo: Find/Design a unique public value.

This commit was co-authored by @keks and @chrisbrzuska.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1319,7 +1319,7 @@ functions as defined in {{!RFC5869}}, as well as the functions
 defined below:
 
 ~~~~~
-HKDF-Expand-Label(Secret, Label, Context, Length) =
+HKDF-Expand-Label(Secret, Label, Context, Count, Index, Length) =
     HKDF-Expand(Secret, HKDFLabel, Length)
 
 Where HKDFLabel is specified as:
@@ -1329,18 +1329,30 @@ struct {
     uint16 length = Length;
     opaque label<7..255> = "mls10 " + Label;
     opaque context<0..2^32-1> = Context;
+    uint8_t count = Count;
+    uint8_t index = Index;
 } HKDFLabel;
 
-Derive-Secret(Secret, Label) =
-    HKDF-Expand-Label(Secret, Label, "", Hash.length)
+Derive-Secret(Secret, Label, Context) =
+    HKDF-Expand-Label(Secret, Label, Context, 0, 1, Hash.length)
+
+Derive-Secrets(Secrets, Count, Label, Context) =
+        HKDF-Expand-Label(Secret0, Label, Context, 0,   n, Hash.length)
+    xor HKDF-Expand-Label(Secret1, Label, Context, 1,   n, Hash.length)
+              ⁝
+    xor HKDF-Expand-Label(Secret1, Label, Context, n-1, n, Hash.length)
+
+Where n is the number of secrets in Secrets.
 ~~~~~
 
 The Hash function used by HKDF is the ciphersuite hash algorithm.
 Hash.length is its output length in bytes.  In the below diagram:
 
-* HKDF-Extract takes its salt argument from the top and its IKM
-  argument from the left
 * Derive-Secret takes its Secret argument from the incoming arrow
+* Derive-Secrets takes its Secrets list argument from the
+  incoming arrows
+* `unique_value` needs to be the same unique value for all
+  derivation steps in a phase (and SHOULD be different between different phases)
 
 When processing a handshake message, a client combines the
 following information to derive new epoch secrets:
@@ -1353,32 +1365,32 @@ Given these inputs, the derivation of secrets for an epoch
 proceeds as shown in the following diagram:
 
 ~~~~~
-               init_secret_[n-1] (or 0)
-                     |
-                     V
-    PSK (or 0) -> HKDF-Extract = early_secret
-                     |
-               Derive-Secret(., "derived", "")
-                     |
-                     V
-commit_secret -> HKDF-Extract = epoch_secret
+ init_secret_[n-1] (or 0)-+
+                          |
+   commit_secret-----+    |
+                     |    |
+     PSK (or 0)-+    |    |
+                |    |    |
+                |    |    |
+                V    V    V
+               Derive-Secrets([., ., .], "combine", unique_value) = epoch_secret
                      |
                      +--> HKDF-Expand(., "mls 1.0 welcome", Hash.length)
                      |    = welcome_secret
                      |
-                     +--> Derive-Secret(., "sender data", GroupContext_[n])
+                     +--> Derive-Secret(., "sender data", (unique_value, GroupContext_[n]))
                      |    = sender_data_secret
                      |
-                     +--> Derive-Secret(., "handshake", GroupContext_[n])
+                     +--> Derive-Secret(., "handshake", (unique_value, GroupContext_[n]))
                      |    = handshake_secret
                      |
-                     +--> Derive-Secret(., "app", GroupContext_[n])
+                     +--> Derive-Secret(., "app", (unique_value, GroupContext_[n]))
                      |    = application_secret
                      |
-                     +--> Derive-Secret(., "exporter", GroupContext_[n])
+                     +--> Derive-Secret(., "exporter", (unique_value, GroupContext_[n]))
                      |    = exporter_secret
                      |
-                     +--> Derive-Secret(., "confirm", GroupContext_[n])
+                     +--> Derive-Secret(., "confirm", (unique_value, GroupContext_[n]))
                      |    = confirmation_key
                      |
                      V
@@ -1387,6 +1399,8 @@ commit_secret -> HKDF-Extract = epoch_secret
                      V
                init_secret_[n]
 ~~~~~
+
+[[OPEN ISSUE: `unique_value` needs to be chosen to be derived from information that can be expected to be unique. Choose such information]]
 
 ## Pre-Shared Keys
 


draft-ietf-mls-protocol.md
```diff
@@ -1329,18 +1329,30 @@ struct {
...
-Derive-Secret(Secret, Label) =
-    HKDF-Expand-Label(Secret, Label, "", Hash.length)
+Derive-Secret(Secret, Label, Context) =
+    HKDF-Expand-Label(Secret, Label, Context, 0, 1, Hash.length)
+
+Derive-Secrets(Secrets, Count, Label, Context) =
+        HKDF-Expand-Label(Secret0, Label, Context, 0,   n, Hash.length)
+    xor HKDF-Expand-Label(Secret1, Label, Context, 1,   n, Hash.length)
+              ⁝
+    xor HKDF-Expand-Label(Secret1, Label, Context, n-1, n, Hash.length)
```
wbl, line 1343: Are you sure you don't mean Secretn here?

draft-ietf-mls-protocol.md
```diff
@@ -1388,6 +1400,8 @@ commit_secret -> HKDF-Extract = epoch_secret
                init_secret_[n]
 ~~~~~
 
+[[OPEN ISSUE: `unique_value` needs to be chosen to be derived from information that can be expected to be unique. Choose such information]]
```
bifurcation, line 1403: Is there a reason that `unique_value` needs to be folded in  both at the `epoch_secret` level and for the secrets derived from it?  Seems like it would be simpler to fold it into the `epoch_secret`.

draft-ietf-mls-protocol.md
```diff
@@ -1353,32 +1365,32 @@ Given these inputs, the derivation of secrets for an epoch
...
+                |    |    |
+                |    |    |
+                V    V    V
+               Derive-Secrets([., ., .], "combine", unique_value) = epoch_secret
                      |
                      +--> HKDF-Expand(., "mls 1.0 welcome", Hash.length)
                      |    = welcome_secret
                      |
-                     +--> Derive-Secret(., "sender data", GroupContext_[n])
+                     +--> Derive-Secret(., "sender data", (unique_value, GroupContext_[n]))
```
bifurcation, line 1381: The `(x, y)` syntax is not defined.  And in any case, I think `GroupContext_[n]` is the unique value you're looking for.

draft-ietf-mls-protocol.md
bifurcation: Also, "unique" in what sense?

draft-ietf-mls-protocol.md
chrisbrzuska: Thanks! Fixed that.

draft-ietf-mls-protocol.md
chrisbrzuska: Excellent questions! Thanks :-)

**Double use of unique value:**
Derive-Secrets derives a keys from each input keys and xors them and the result is assigned to epoch_secret. It is important that the inputs to the xor have not been used before and the unique value ensures that. tl;dr: 1st use of unique_value is needed for pseudorandomness.

The second use of the unique value ensures that different sessions derive different key values. For honest sessions, this is the case anyway (since pseudorandom key values are unlikely to collide), but for dishonest sessions, the unique value is needed since the xor does not maintain collision-resistance.

**Definition of unique:**
In a call to a Derive-Secrets function with fixed label, e.g.,

Derive-Secrets([., ., .], "combine", unique_value),

the unique-value should not repeat (across protocol runs), even with different input keys. In Derive-Secret with fixed label, e.g., 

Derive-Secret(., "sender data", (unique_value,.)),

the unique-value should not repeat (across protocol runs), even with different input keys. However, it is allowed that unique_values are re-used when the labels are different.

draft-ietf-mls-protocol.md
chrisbrzuska: (x,y) should refer to an injective encoding of pairs. If GroupContext_[n] is already unique, this simplifies the construction and would be very nice. We were hoping that GroupContext_[n] is already unique, but we don't have the full MLS context to make this judgement. Is GroupContext_[n] also already available at the point when we compute the epoch_secret?

draft-ietf-mls-protocol.md
```diff
@@ -1329,18 +1329,30 @@ struct {
...
+    uint8_t index = Index;
 } HKDFLabel;
 
-Derive-Secret(Secret, Label) =
-    HKDF-Expand-Label(Secret, Label, "", Hash.length)
+Derive-Secret(Secret, Label, Context) =
+    HKDF-Expand-Label(Secret, Label, Context, 0, 1, Hash.length)
+
+Derive-Secrets(Secrets, Count, Label, Context) =
+        HKDF-Expand-Label(Secret0, Label, Context, 0,   n, Hash.length)
```
Bren2010, line 1340: I think the correct syntax here would be `Secrets[0]`, `Secrets[1]` ...

Bren2010: Also, you've added extra spaces before "n, " but they don't seem to align with the last call

draft-ietf-mls-protocol.md
```diff
@@ -1319,7 +1319,7 @@ functions as defined in {{!RFC5869}}, as well as the functions
 defined below:
```
Bren2010, line 1319: The line just above here references HKDF-Extract, which is no longer used anywhere. Please remove

draft-ietf-mls-protocol.md
```diff
@@ -1329,18 +1329,30 @@ struct {
     uint16 length = Length;
     opaque label<7..255> = "mls10 " + Label;
     opaque context<0..2^32-1> = Context;
+    uint8_t count = Count;
+    uint8_t index = Index;
```
Bren2010, line 1333: Just "uint8" no "_t"

draft-ietf-mls-protocol.md
```diff
@@ -1329,18 +1329,30 @@ struct {
...
+Derive-Secret(Secret, Label, Context) =
+    HKDF-Expand-Label(Secret, Label, Context, 0, 1, Hash.length)
+
+Derive-Secrets(Secrets, Count, Label, Context) =
+        HKDF-Expand-Label(Secret0, Label, Context, 0,   n, Hash.length)
+    xor HKDF-Expand-Label(Secret1, Label, Context, 1,   n, Hash.length)
+              ⁝
+    xor HKDF-Expand-Label(Secretn-1, Label, Context, n-1, n, Hash.length)
+
+Where n is the number of secrets in Secrets.
```
Bren2010, line 1345:
```suggestion
where n is the length of Secrets.
```

draft-ietf-mls-protocol.md
```diff
@@ -1353,32 +1365,32 @@ Given these inputs, the derivation of secrets for an epoch
...
-commit_secret -> HKDF-Extract = epoch_secret
+ init_secret_[n-1] (or 0)-+
+                          |
+   commit_secret-----+    |
+                     |    |
+     PSK (or 0)-+    |    |
+                |    |    |
+                |    |    |
+                V    V    V
+               Derive-Secrets([., ., .], "combine", unique_value) = epoch_secret
```
Bren2010, line 1376: Not enough arguments to Derive-Secrets

draft-ietf-mls-protocol.md
Bren2010: Yes GroupContext_[n] is what you're looking for, and I also don't like the tuples.

draft-ietf-mls-protocol.md
Bren2010: The label "combine" is weird, maybe "epoch" instead

chris-wood:
@chrisbrzuska are you certain this general construction gives us an n-PRF with the desired properties? What if we wanted to replace HMAC with a different PRF?

draft-ietf-mls-protocol.md
bifurcation: Thanks for the clarification on the double input.

I think the GroupContext value does meet the requirement for uniqueness here.  It contains the following values:

* Group ID
* Epoch
* Tree hash
* Transcript hash

Obviously, GroupID and Epoch provide a first level of protection, and the tree hash diverges if there are any different choices of keys. Even if the participants make the same choices in two protocol runs, HPKE uses random keys which are fed into the transcript hash, so will be different.

draft-ietf-mls-protocol.md
bifurcation: Thanks for the clarification.  We would want a defined injective encoding here.  But if we agree that GroupContext is unique enough, maybe we can just use that?

jhoyla:
> I don't like that before, we had Extract calls which theoretically turn non-uniform randomness into uniform randomness, and now that's all replaced with HKDF-Expand calls that theoretically always assume the randomness is uniform.

I agree with @Bren2010, what's the theory justification for changing to an Expand-XOR-Expand?
Given that Expand and Extract are both calls to an underlying HMAC this is potentially OK, but it's definitely not obvious. 

The unique value in the `Derive-Secrets` acts as a salt, and the key material acts as the IKM, so the first call to `Derive-Secrets` can act as `HKDF-Extract`, and the subsequent calls to `Derive-Secret` can act as `HKDF-Expand`s, but I think there should be an explicit requirement / mentioning that the two functions must always have different types when used as `Extract`s vs `Expand`s. 

As far as I can tell this mostly seems to be a labelling change, renaming `Extract` to overload `Expand`. A security section expanding on why this is secure, preferably with some references to computational proofs would be really good here. 

It might be better to just define a new `HKDF-Extract`-type function as `(Expand-XOR)`, and enforce a different label in the inner `Extract` step. That would hopefully prevent any collisions of subkeys. 

It also seems like you could define 
```
Derive-Secret(Secret, Label, Context) =
        Derive-Secrets([Secret], 1, Label, Context)
```
so we need to be aware of type collisions there as well.

jhoyla:
@chrisbrzuska Have you submitted your paper somewhere for publication / peer review?
I had a read of it, and I tried to follow some of your proofs, but despite the excellent diagrams I really struggled. 

There are many very similarly named things that it's unclear what the difference is. For example I couldn't figure out the difference between CORE_{GNPRF}^0 and CORE^0_{NPRF}. 
The text says:
> CORE_{GNPRF}^0 is defined in Figure 20a

But 20a is labelled `CORE^0_{NPRF},R^0_{NPRF}`

I also was a bit confused by the overloading of `Eval`.

Another issue I had was finding all the parts of the proof. To follow the NPRF proof, for example, requires huge amounts of paging back and forth through the document, which large gaps between sections (for the other proofs). Could you possibly share the TeX file, or provide a re-ordered document with all the pieces of each proof next to each other. It would really help people to evaluate the proof. 
(A glossary which explains what each thing is and provides a link to the definition would be awesome too.)

Even though I couldn't completely follow the proofs it does look like all the pieces are there, so maybe add a section to the Security Considerations with a link to the paper. I think if people find this proof satisfactory then this change seems reasonable.

I know this is getting a bit off topic, but I found a couple of nits in your paper. Things like on page 12 you have `skX_i = skY = ⊥`, which seems like it should be `skX_i = skY_i = ⊥`. Let me know if a list would be helpful to you.

chrisbrzuska:
>>> Let me know if a list would be helpful to you.

Thank you for the close read, very appreciated. Yes, a list of nits as well as any other feedback would be very useful to us. Especially, also: How did you proceed to read the proof? What distracted, confused or annoyed you? Which were the bits which required page-scrolling? What came first to my mind is the Key package, but I guess there are more parts of the proof which require scrolling? The idea of a glossary is nice.

jhoyla:
Hi @chrisbrzuska

I tried to read the proof by following just the proof for the NPRF. I wanted to see all the stages in their simplest form.
I read the paper from start to finish, just skimming the proofs, just to get a sense of what was going on, and then I started to work backwards from Claim 1.
The first step of the proof is to compare Figure 20a, Figure 17a, and Figure 15a, or pages 35, 26, and 21. 
Then one must refer to Fig. 14, Fig. 13a, and Fig. 10 or pages 18, 16, and 12.

Figure 19 brings them all together, but I like to compare the details. The big jumps from 35 to 26 and 21 are particularly frustrating.
I ended up screenshotting all the definitions and putting them side by side in powerpoint. 

I was able to convince myself that claim 1 is true, but I didn't really follow the proof in Section 6.2. 

I struggled with both the logic of the steps, i.e. that each step followed from the previous step, and with understanding the conclusion.
In very informal terms, my understanding of the reduction in Fig 20a-d is:
Replace the PRF with the collision resistant PRF.
The XOR can now be replaced with the collision resistant XOR.
The PRF can then be replaced with the non-collision resistant PRF.
Fig. 20d thus describes the collision resistant NPRF.
Is that approximately right?

With respect to the result, the probability equation has terms that I can't evaluate.
What does e_{PRF}(B→R^1_{NPRF}) actually evaluate to? 
Is it obvious that it is ≤ e_{PRF}(A)?

Nits / Clarification questions:
1. In the Key package, is there a requirement that the public name be _cryptographically_ bound to the key? 
2. Figure 7. could use a more permanent reference, i.e. "current" will change eventually.
3. On pg. 12 should `skXi = skY = ⊥` read `skXi = skYi = ⊥`?
4. Is there a significance to `sort` being unstable?
5. On pg. 14 you write "Finally in Section 3.2", when 3.2 is not the last subsection of section 3.
6. Some of your references have gone a bit weird, specifically [1,3,4,5,8,10,11,12,16]. "in [name], editor" seems wrong.
7. Please add a reference to [Kra10] by the min-entropy def. 
8. The pseudo-code in Fig. 12 is out of order. `Set(h, k, ctr)` should read `Set(ctr, k, h)` I think.
9. In Fig. 13b. should the lower `Key^{00}` read `Key^{b0}`?
10. Ditto with Fig. 13c.
11. In Fig. 13c. `k <- xtr(S, pkX^Y)` should read `k <- xtr(S, pkX^y)` (i.e. the `Y` should be in lower case.)
12. In Fig. 14 `S: {0, 1}^{l}` should probably read `S: {0,1}^{sl}`. 
13. "thus duped" should probably read "thus dubbed".
14. The package CRPRF is never defined as such. The definition appears in Fig. 14 with the label GcrPRF^b. Without a glossary it's hard to know whether GcrPRF ≟ NPRF ≟ CRPRF.
15. The final column of Fig. 19 replaces code with f_{cfNPRF}, but the code does not exactly match. `f_{PRF}(k_i, (ctx1, i), ol)`  is replaced with `f_{PRF}(k_i, (ctx1, i))`, i.e. the final parameter disappears.

I didn't try and go through the other 2 code comparisons exactly, but if I have time I will.

draft-ietf-mls-protocol.md
chrisbrzuska: Sorry for the delay in my response. Is the length of Secrets equal to the number of secrets?

chrisbrzuska:
Hi @jhoyla 

Thanks a lot for the thorough read and the reading feedback! It took me a while to get around integrating your suggested changes into the document. I started with the nits and clarifications (comments below). Will post the new document within the next days.

> Nits / Clarification questions:
> 
> 1. In the Key package, is there a requirement that the public name be _cryptographically_ bound to the key?

No, it can be a purely administrative value. However, in the NamePRF and the DHNPRF construction, we explore the use of cryptographic public names (handles).

> 2. Figure 7. could use a more permanent reference, i.e. "current" will change eventually.

Good point. Done.

> 3. On pg. 12 should `skXi = skY = ⊥` read `skXi = skYi = ⊥`?

Thank you for catching this. Fixed.

> 4. Is there a significance to `sort` being unstable?

What does 'unstable' mean in this context?

> 5. On pg. 14 you write "Finally in Section 3.2", when 3.2 is not the last subsection of section 3.

Fixed.

> 6. Some of your references have gone a bit weird, specifically [1,3,4,5,8,10,11,12,16]. "in [name], editor" seems wrong.

Thank you, not yet fixed.

> 7. Please add a reference to [Kra10] by the min-entropy def.

Fixed.

> 8. The pseudo-code in Fig. 12 is out of order. `Set(h, k, ctr)` should read `Set(ctr, k, h)` I think.

That's correct. Thanks! Fixed.

> 9. In Fig. 13b. should the lower `Key^{00}` read `Key^{b0}`?
> 10. Ditto with Fig. 13c.

Indeed. Fixed.

> 11. In Fig. 13c. `k <- xtr(S, pkX^Y)` should read `k <- xtr(S, pkX^y)` (i.e. the `Y` should be in lower case.)
> 12. In Fig. 14 `S: {0, 1}^{l}` should probably read `S: {0,1}^{sl}`.

Fixed.

> 13. "thus duped" should probably read "thus dubbed".

Fixed :-D

> 14. The package CRPRF is never defined as such. The definition appears in Fig. 14 with the label GcrPRF^b. Without a glossary it's hard to know whether GcrPRF ≟ NPRF ≟ CRPRF.

Clarified Fig. 14 to describe the packages crPRF and PRF. There should be no "CRPRF", this is a typo. Removed throughout the paper. NPRF is not equal to CPRF, too bad there was a confusion.

> 15. The final column of Fig. 19 replaces code with f_{cfNPRF}, but the code does not exactly match. `f_{PRF}(k_i, (ctx1, i), ol)`  is replaced with `f_{PRF}(k_i, (ctx1, i))`, i.e. the final parameter disappears.

Right. We are still wondering whether different output lengths are needed or not and are inconsistent w.r.t. this currently in the paper. Will make it uniform once we determined whether this is needed.

seanturner:
@chrisbrzuska Here's a link to the [CFRG](https://irtf.org/cfrg) (Crypto Forum Research Group).  It's in IRTF (Internet Research Task Force) the sister organization of the IETF. As noted during the interim, there a lot of cryptographers there and IETF WGs often go to get a review of crypto-related items. You can subscribe [here](https://www.irtf.org/mailman/listinfo/cfrg).

draft-ietf-mls-protocol.md
chrisbrzuska: Okay, this sounds very good, thank you!

chrisbrzuska:
> @chrisbrzuska Here's a link to the [CFRG](https://irtf.org/cfrg) (Crypto Forum Research Group). It's in IRTF (Internet Research Task Force) the sister organization of the IETF. As noted during the interim, there a lot of cryptographers there and IETF WGs often go to get a review of crypto-related items. You can subscribe [here](https://www.irtf.org/mailman/listinfo/cfrg).

@seanturner: Excellent, thank you! I subscribed and will be there tomorrow in case I get an opportunity to present the design.

bifurcation:
This PR has fallen a bit behind, and the key schedule has been updated so that it's less relevant.  Still open to the idea of n-PRF, but it will need a new PR.

PR #338, opened 2020-05-19
Rely More on HPKE

rozbb:
The first commit is editorial. I removed all the DH terminology I could find and replaced it with KEM terminology.

The second commit is related to https://github.com/cfrg/draft-irtf-cfrg-hpke/pull/79. It would massively reduce complexity of implementations if people were able to rely on HPKE to do all private key parsing, key validation, public key computation, etc. Currently, all the information in the deleted CFRG and NIST sections is either redundant (performed by HPKE) or outdated and performed correctly by HPKE (e.g., old P-256 DH secret representation). Instead of having all this duplicated, it would be easier to keep it in one place.

An issue with the new definition of `Derive-Key-Pair`: Repeatedly calling `HKDF-Expand-Label` can be computationally expensive. Each call requires computing the group hash (which, okay, it could be cached) and serializing a struct. It would be nice to do something that's cheaper but still ties context into the secret key generation.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -862,70 +862,32 @@ the same as the signature algorithm specified in the credential field of the
 KeyPackage objects in the leaves of the tree (including the InitKeys
 used to add new members).
 
+For all ciphersuites defined in this document, the `Derive-Key-Pair` function
+rejection-samples candidate private key byte sequences until the `UnmarshalSk`
+function defined in HPKE {{!I-D.irtf-cfrg-hpke}} succeeds. Concretely,
+`Derive-Key-Pair(Secret)` will derive candidate octet strings of length `Nsk`
+(a KEM-specific constant defined in HPKE) as below
+
+~~~~~
+        "derive key pair"
+              |
+              V
+Secret -> HKDF-Extract
+              |
+              +--> HKDF-Expand-Label(., "sk candidate 1.", "", Nsk)
+              |
+              +--> HKDF-Expand-Label(., "sk candidate 2.", "", Nsk)
+              |
+              .
+              .
+              .
+~~~~~
+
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -862,68 +862,32 @@ the same as the signature algorithm specified in the credential field of the
 KeyPackage objects in the leaves of the tree (including the InitKeys
 used to add new members).
 
-The ciphersuites are defined in section {{mls-ciphersuites}}.
+For all ciphersuites defined in this document, the `Derive-Key-Pair` function
```
Bren2010, line 865: I think it would make more sense for the ciphersuite / HPKE to define how it generates a private key, given a seed, than to specify an Unmarshal function and do rejection sampling. Rejection sampling isn't guaranteed to work for many schemes -- RSA being the one that comes to mind, but possibly some PQ schemes also.

rozbb: This is a great point. In this case, `Derive-Key-Pair` should just be HPKE's version of `Derive-Key-Pair`, possibly after some HKDF step. This also lets us avoid doing an `Expand-Label` for every iteration.

rozbb:
Update: This now uses HPKE's `DeriveKeyPair` function to do all the heavy lifting

draft-ietf-mls-protocol.md
```diff
@@ -720,9 +720,9 @@ node, from which the node's key pair is derived.
 
 ~~~~~
 path_secret[0] = HKDF-Expand-Label(leaf_hpke_secret,
-                                   "path", "", Hash.Length)
+                                   "path", "", Nsk)
```
bifurcation, line 723: Might be clearer to write something like `KEM.Nsk`.

PR #345, opened 2020-06-10
no need extra layer of optional in GroupInfo.tree

yihuang:
since the Node is optional inside.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2181,7 +2181,7 @@ Commit.
 struct {
   opaque group_id<0..255>;
   uint64 epoch;
-  optional<Node> tree<1..2^32-1>;
+  Node tree<1..2^32-1>;
   opaque confirmed_transcript_hash<0..255>;
   opaque interim_transcript_hash<0..255>;
   Extension extensions<0..2^16-1>;


bifurcation:
@yihuang I think your analysis is correct here, but I would eliminate the other `optional<>`.  That is, something like:

```
struct {
    NodeType node_type;
    select (Node.node_type) {
        case leaf:   KeyPackage key_package;
        case parent: ParentNode node;
    };
} Node;

optional<Node> tree<1..2^32-1>;
```

If we agree on the approach here, I can implement it in #348, which makes the tree optional as @raphaelrobert suggests.

yihuang:
> @yihuang I think your analysis is correct here, but I would eliminate the other `optional<>`. That is, something like:
> 
> ```
> struct {
>     NodeType node_type;
>     select (Node.node_type) {
>         case leaf:   KeyPackage key_package;
>         case parent: ParentNode node;
>     };
> } Node;
> 
> optional<Node> tree<1..2^32-1>;
> ```
> 
> If we agree on the approach here, I can implement it in #348, which makes the tree optional as @raphaelrobert suggests.

Your approach is even better.

bifurcation:
Thanks @yihuang.  This has been folded into #348

Issue #344, opened 2020-06-02
Tree in GroupInfo not optional


raphaelrobert:
I remember we discussed this in the past, but I don't think we ever spec'ed this out fully:

Right now the whole tree is part of `GroupInfo` (which in turn is part of `Welcome`). This can grow quite big in large groups (~17MB with 100k members with the MTI). The practical limitations are:

 - The TLS representation of the tree is hard to parse efficiently (no skipping and no a priori information of how many nodes are in the tree)
 - Out-of-band transmission (like server assist) is not possible
 - Large messages like that are hard to handle on mobile devices, when they have to be downloaded in the background

We should introduce an indirection, like a commit hash of the tree.

PR #347, opened 2020-06-12
Explicitly specify which nodes are targeted

bifurcation:
Currently, processing a DirectPath requires that the recipient have a full view of the ratchet tree, so that it can compute the resolutions of copath nodes.  This PR changes the DirectPath structure to explicitly state which tree node a given encryption of a path secret is targeted to.  This is a prerequisit for making full knowledge of the tree optional, as in #344.


seanturner:
Plan is to close this PR.

PR #348, opened 2020-06-12
Make the tree in the Welcome optional

### LINKED ISSUE #344: Tree in GroupInfo not optional

bifurcation:
Fixes #344


bifurcation:
A couple of shortcomings to note on this PR:

* The joiner can't compute their private keys because they're not told where they are in the tree or how big the tree is.  The former should be added to `GroupSecrets` and the latter to `GroupInfo`.

* This approach assumes that the joiner gets the tree OOB before processing the Welcome, otherwise they can't verify the signature on the Welcome.  (Since Welcome presumes they get the credential for the signer from the tree.)

We could avoid the latter problem by including a credential for the signer and a "membership proof" comprising the hashes needed to tie that leaf to the tree hash in the Welcome.  That could allow participants to remain "passive" (without a full tree, just their private keys) for longer, especially if such a proof could be provided on Commit as well.  This might be a good candidate for an extension, though.

bifurcation:
Discussion at the 2020-07-14 interim indicated that we should go ahead with this strategy, and just note the requirement that you have to get the tree before you process the Welcome.

PR #349, opened 2020-06-16
Use opaque epoch IDs and encrypt content type

### LINKED ISSUE #142: Prevent suppression of Handshake messages
In current proposals, it is possible for the delivery service to distinguish handshake and application messages.  This enables the DS to lock in a compromise by forbidding key updates while allowing application messaging to continue.  We should make it harder to distinguish handshake and application messages, e.g., by encrypting the content type.

### LINKED ISSUE #269: Randomize values in the common framing header
We can't really touch the ContentType because of the ordering constraints.

We could try to improve for the Epoch Number but loosing lexicographical ordering of the epoch numbers is a problem for transport loss + DS reordering.

We can definitely make sure that only group members can recognize the Group Identifier *across epochs* by generating the GID from the Key Schedule.
We also can do significantly better if we allow the DS to assist and modify the GID by deterministically blinding the GID with a pseudo random recipient ID. In that case seeing the GID after dispatching in the delivery queue will make it significantly more difficult to determine the group.

bifurcation:
Fixes #269 
Fixes #142

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1614,6 +1614,7 @@ object of the following form:
 struct {
     uint32 sender;
     uint32 generation;
+    ContentType content_type;
     opaque reuse_guard[4];
 } MLSSenderData;
 ~~~~~
@@ -1632,7 +1633,6 @@ computation is its prefix in the MLSCiphertext, namely:
 ~~~~~
 struct {
     EpochID epoch_id;
-    ContentType content_type;
     opaque authenticated_data<0..2^32-1>;
     opaque sender_data_nonce<0..255>;
 } MLSCiphertextSenderDataAAD;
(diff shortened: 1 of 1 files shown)


bifurcation:
@raphaelrobert - You asked today if I had implemented this.  I tried to today and failed!  The problem is that we use different keys for the content encryption depending on the content type.  So the content_type can't go in the encrypted content, it has to go in the sender data.  With that modification, the remainder of the PR worked as expected.  Pushing an update to this PR shortly.

PR on the Go implementation here: https://github.com/cisco/go-mls/pull/77

uhoreg:
FWIW, making the epoch opaque rather than an increasing number will help with the decentralized case where users' views of the ratchet tree may branch and merge.  There are of course still other things that will need to change for this case, but this will be one fewer difference.

Bren2010:
@bifurcation Re-iterating my points from the mailing list discussion:
- Implementations that can branch will still have to handle branching on their own, consistent with the way in which they branch.
- The privacy properties of the opaque epoch id are very weak and privacy-preserving applications will still have to invent their own way to hide this metadata.

Please decline this PR!

PR #350, opened 2020-06-16
IANA updates and their consequences

### LINKED ISSUE #312: Add IANA considerations for Extensions
Now that we have a more robust extension mechanism, we need to have an IANA registry to manage extensions, including:

* Registry structure
* Registry policy and (probably) designated experts
* Reserved vendor space

bifurcation:
This started off just fixing #312, then grew a bit.

* Defines an IANA registry for ExtensionType values
* Defines an IANA registry for CredentialType values
* Updates the Ciphersuites registry to share formatting with the other two, and to be more explicit about its ties to HPKE algorithms
* Removes the unnecessary `signature_scheme` field from `BasicCredential`
* ... and as a result removes the definition of `SignatureScheme` (which is used nowhere else)

Fixes #312


draft-ietf-mls-protocol.md
```diff
@@ -75,6 +75,11 @@ informative:
        - name: Trevor Perrin(ed)
        - name: Moxie Marlinspike
 
+  SECG:
```
seanturner, line 78: Okay I see that we are referring to the uncompressed key format in SECG. I know there's a reference for that somewhere else, but I can't think of it right now. Go ahead and don't block on me and I can submit reference updates later.

PR #352, opened 2020-06-16
Use node_index for both hashes

### LINKED ISSUE #328: Use the same index for hashing parent and leaf nodes
Right now the LeafNodeHashInput struct uses the index of the node among the leaves, while the ParentNodeHashInput struct uses the index of the node among all nodes.  This means that, for example, the index value will be the same for the second leaf and its parent (`1` in both cases).

We should probably just use the node index in both cases.

bifurcation:
Fixes #328


bifurcation:
@Bren2010 - I seem to recall you had some opinions on node hashing?

PR #353, opened 2020-06-16
Explain the meaning of a Commit with no proposals

### LINKED ISSUE #340: Clarify what it means to send a Commit with no Proposals
Right now, it is possible to send a Commit that covers no proposals.  This has the result that the leaf of the sender and its direct path are refreshed, but no other change to the tree is made.  (The epoch also advances, of course.)  This can be is a useful operation.  For example, an Add wipes out the new joiner's direct path, so having the new joiner send a Commit over no proposals would fill in that direct path and make subsequent operations more efficient.

We should clarify in some prose that (a) this is possible and (b) why it might be useful.

bifurcation:
Fixes #340


draft-ietf-mls-protocol.md
```diff
@@ -2006,6 +2006,14 @@ The Commit MUST NOT combine proposals sent within different epochs. In the event
 that a valid proposal is omitted from the next Commit, the sender of the
 proposal SHOULD retransmit it in the new epoch.
 
+A member of the group MAY send a Commit that references no proposals at all,
+which would thus have empty vectors for `updates`, `removes`, and `adds`.  Such
+a Commit resets the sender's leaf and the nodes along its direct path and
```
Bren2010, line 2011: I don't believe such a commit would change the leaf. Makes more sense to me to not allow empty commits

bifurcation: It would reset the leaf, via `Commit.path.leaf_key_package`.

ISTM there is a very concrete use case for this feature, namely keeping the tree full.  If you want to maintain a full tree, then you would want to have each new joiner should send an empty commit immediately on joining, have the neighbor of a removed node send an empty commit after a remove, etc.

draft-ietf-mls-protocol.md
```diff
@@ -2006,6 +2006,14 @@ The Commit MUST NOT combine proposals sent within different epochs. In the event
...
+advances the group to a new epoch, with the forward secrecy and post-compromise
+security guarantees that that entails.  An Update proposal can be regarded as a
+"lazy" version of this operation, where only the leaf changes and intermediate
+nodes are blanked out.
```
kkohbrok, line 2015: So an empty commit is essentially what used to be an update, i.e. fresh key material for the committers leaf, which is then KDFed up the tree. Functionally, I am ok with this, but it seems a little counter-intuitive to me. Wouldn't it be clearer if we disallowed empty commits and a party that would like to keep the tree full simply committed an Update proposal targeting their own leaf? That would yield the same result, wouldn't it?

Also: Advancing an epoch doesn't always give PCS guarantees. If an Add proposal is committed without an Update and `path` is empty, the epoch changes, but we don't get PCS (this is why I was critical of allowing Adds without Updates. It breaks the assumption that epoch change implies PCS).

bifurcation: Advancing the epoch gives you PCS with respect to the committer and and any nodes that were updated/removed in the epoch.  That doesn't change here.

kkohbrok: I'm pretty sure it doesn't if there is no `path` value. If I understand correctly, an `Add` only commit doesn't update the committers secrets (and thus doesn't get you PCS), but _does_ advance the epoch. I was only pointing out that advancing the epoch doesn't imply PCS if we allow empty `path` fields.

bifurcation: You are correct that a Commit with no `path` doesn't cause any PCS.  But we're talking about empty Commits here, which are required to have a `path`.  I've edited to be more precise.

bifurcation:
I think we need to clearly define "FS / PCS with regard to", but once we do that, it seems like a fine phrase.  That seems like a good topic for a review / update of the Security Considerations once we have the protocol mostly in shape, for which I have filed #365.

PR #354, opened 2020-06-23
misc little fixes

uhoreg:
I've put a bunch of little fixes in here.  If you disagree with any of these changes, I can back them out or change them.

- maximal is defined but never used
- k-th leaf/leaf index is defined in two places; merge definitions
- change KeyPackage struct definition to reflect that ~2~ 3 extensions are required
- `lifetime` is also a required extension
- grammar fix
- add missing semicolon

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -980,7 +980,7 @@ struct {
     CipherSuite cipher_suite;
     HPKEPublicKey hpke_init_key;
     Credential credential;
-    Extension extensions<3..2^16-1>;
+    Extension extensions<12..2^16-1>;
     opaque signature<0..2^16-1>;
 } KeyPackage;
 ~~~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -982,13 +980,14 @@ struct {
     CipherSuite cipher_suite;
     HPKEPublicKey hpke_init_key;
     Credential credential;
-    Extension extensions<0..2^16-1>;
+    Extension extensions<3..2^16-1>;
```
bifurcation, line 983: Each extension is at least 4 bytes long (2 for type, 2 for length), so if you want the real minimum here, it should be 12.  I'm generally not sanguine about using the TLS struct definition to enforce this sort of thing, though.

uhoreg: Fixed

PR #355, opened 2020-07-10
Validate external proposals from preconfigured senders

d1vyank:


d1vyank:
This stuck out to me while reading because verification is explicitly specified for the 'new_member' sender type in the paragraph above this and not for the 'preconfigured' sender type.

I think it would be better to specify this as the verification flow is different for these external proposals i.e. the KeyPackage in the proposal is used for new_member sender types, and the key corresponding to an authorized sender is used for preconfigured senders.

PR #357, opened 2020-07-10
Fix all compiler warnings.

Bren2010:
Previously we used [[ ... ]] for comments, and the body of the comment was only getting elided because the compiler didn't know what that syntax meant.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -43,14 +43,6 @@ author:
 
 
 normative:
-  X962:
-    title: "Public Key Cryptography For The Financial Services Industry: The Elliptic Curve Digital Signature Algorithm (ECDSA)"
-    date: 1998
-    author:
-      org: ANSI
-    seriesinfo:
-      ANSI: X9.62
-  IEEE1363: DOI.10.1109/IEEESTD.2009.4773330
   SHS: DOI.10.6028/NIST.FIPS.180-4
 
 informative:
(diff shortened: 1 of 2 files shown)


Bren2010:
Please don't remove the .editorconfig file again! It's a [normal thing](https://github.com/quicwg/base-drafts/blob/master/.editorconfig) for projects to have!!

draft-ietf-mls-protocol.md
```diff
@@ -2732,8 +2724,8 @@ New ciphersuite values are assigned by IANA as described in
 
 ## MLS Designated Expert Pool {#de}
 
-[[ OPEN ISSUE: pick DE mailing address.
-Maybe mls-des@ or mls-de-pool. ]]
+<!-- OPEN ISSUE: pick DE mailing address.
```
bifurcation, line 2727: I don't think this quite matches the intent -- we want these OPEN ISSUES to be visible in the rendered HTML

draft-ietf-mls-protocol.md
```diff
@@ -2610,8 +2602,8 @@ tree is secret and thus so are all values derived from it.
 
 Initial leaf keys are known only by their owner and the group creator,
 because they are derived from an authenticated key exchange protocol.
-Subsequent leaf keys are known only by their owner. [[TODO: or by
-someone who replaced them.]]
+Subsequent leaf keys are known only by their owner.
+<!-- TODO: or by someone who replaced them. -->
```
bifurcation, line 2606: I think we can delete this TODO.

draft-ietf-mls-protocol.md
Bren2010: I noticed this is how QUIC WG does it: https://github.com/quicwg/base-drafts/commit/09338d0e237e67c648aa172c32b872e10bf737ff

PR #360, opened 2020-07-14
#302: Reduce encryption overhead to just auth tag

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -1513,82 +1513,20 @@ struct {
...
-    ContentType content_type;
-    opaque authenticated_data<0..2^32-1>;
-} MLSSenderDataAAD;
-~~~~~
-
-When parsing a SenderData struct as part of message decryption, the
-recipient MUST verify that the sender field represents an occupied
-leaf in the ratchet tree.  In particular, the sender index value
-MUST be less than the number of leaves in the tree.
+MLSPlaintext object and how to convert it to an MLSCiphertext object.
```
bifurcation, line 1527: I think the overview steps are important to have here, especially since the ciphertext sampling makes the order of things a bit complex (you have to encrypt the content, then loop back to the sender data).  But I think we streamline a little bit:

```
* Set group_id, epoch, content_type and authenticated_data fields from the
  MLSPlaintext object directly

* Identify the key and key generation depending on the content type

* Encrypt an MLSCiphertextContent for the ciphertext field using the key
  identified and MLSPlaintext object

* Encrypt the sender data using the sender data key for the epoch and a
   nonce sampled from  encrypted MLSCiphertextContent.
```

draft-ietf-mls-protocol.md
```diff
@@ -1686,24 +1626,70 @@ decryption.
...
 struct {
     opaque group_id<0..255>;
     uint64 epoch;
     ContentType content_type;
     opaque authenticated_data<0..2^32-1>;
-    opaque sender_data_nonce<0..255>;
-    opaque encrypted_sender_data<0..255>;
-} MLSCiphertextContentAAD;
+    opaque ciphertext<0..2^32-1>;
+} MLSCiphertextAAD;
```
bifurcation, line 1686: You've removed the AAD from the main content encryption, which seems wrong.  it seems like we should have the unencrypted sender data feed in the ciphertext AAD, and the ciphertext (or the derived nonce) in the sender data AAD.  Something like:

```
      /  group_id            \
      |  epoch               |--+
   +--|  content_type        /  |
   |  |  sender_data         <--+
   |  \  authenticated_data  \__|
   +-->  content             /

struct {
    opaque group_id<0..255>;
    uint64 epoch;
    ContentType content_type;
    opaque authenticated_data<0..2^32-1>; // <---
    opaque ciphertext<0..2^32-1>;
} SenderDataAAD;

struct {
    opaque group_id<0..255>;
    uint64 epoch;
    ContentType content_type;
    SenderData sender_data;
    opaque authenticated_data<0..2^32-1>; // <--- 
} ContentAAD;
```

draft-ietf-mls-protocol.md
```diff
@@ -2691,39 +2677,48 @@ Where VALUE is represented as two 8bit octets:
...
-| LVL       | The security level |
-| KEM       | The KEM algorithm used for HPKE in TreeKEM group operations |
-| AEAD      | The AEAD algorithm used for HPKE and message protection |
-| HASH      | The hash algorithm used for HPKE and the MLS KDF |
-| SIG       | The Signature algorithm used for message authentication |
+| LVL       | The security level                                                     |
+| KEM       | The KEM algorithm used for HPKE in TreeKEM group operations            |
+| AEAD      | The AEAD algorithm used for HPKE and message protection                |
+| HASH      | The hash algorithm used for HPKE and the MLS KDF                       |
+| PRF       | The pseudorandom function used for masking metadata                    |
```
bifurcation, line 2687: Can't we just use the existing KDF we have on hand?

draft-ietf-mls-protocol.md
```diff
@@ -1686,24 +1626,70 @@ decryption.
 +-+-+-+-+---------...---+
 ~~~~~
 
-The Additional Authenticated Data (AAD) input to the encryption
-contains an object of the following form, with the values used to
-identify the key and nonce:
+## Masking Sender Data
```
bifurcation, line 1629: Since we're going to have an auth tag anyway, it seems like we could simplify this a fair bit by just using the ciphersuite's AEAD to encrypt the SenderData (as we do now), and just sampling the *nonce* from the ciphertext.

That obviously removes the need to do XOR in addition to normal AEAD, and may allow us to simplify the sampling as well.  Since the nonce only needs to be random and not secret, we might be able to get away with just using the first `nonce_size` bytes of the ciphertext (or HKDF-Extract thereof).  So it gets rid of the AES-ECB dependency as well.

And since this changes from (AEAD + XOR + AEAD) to (AEAD + AEAD), we no longer need the `authentication_key`.

```
ciphertext = AEAD.Seal(key=generation.key,
                       nonce=generation.nonce,
                       pt=content,
                       aad=header_with_plaintext_sender_data)
sender_data_nonce = ciphertext[:nonce_size]
encrypted_sender_data = AEAD.Seal(key=sender_data_key,
                                  nonce=sender_data_nonce,
                                  pt=sender_data,
                                  aad=header_without_sender_data_and_with_ciphertext)
```

draft-ietf-mls-protocol.md
Bren2010: The ciphertext's auth tag is truncated so there's no point in having AAD on the content. All AAD is against the sender data.

bifurcation:
Just to recap the options as I understand them:

a) Today:
   encrypted_sender_data = AEAD(sender_data_key, sender_data_nonce, (groupID, epoch), sender_data)
   encrypted_content = AEAD(app_key[i][j], app_key[i][j], (groupID, epoch, encrypted_sender_data), content)

b) Sample sender data nonce from ciphertext (saves explicit nonce)
   encrypted_content = AEAD(app_key[i][j], app_nonce[i][j], (groupID, epoch), content)
   encrypted_sender_data = AEAD(sender_data_key, sample(encrypted_content), (groupID, epoch, encrypted_content), sender_data)

c) Drop auth tag from content encryption (saves explicit nonce and content auth tag)
   encrypted_content = Enc(app_key[i][j], app_nonce[i][j], content)
   encrypted_sender_data = AEAD(sender_data_key, sample(encrypted_content), (groupID, epoch, encrypted_content), sender_data)

Bren2010:
Ok updated to strategy B

bifurcation (changes requested):
Thanks, @Bren2010, this looks pretty close.  Couple things that need adjusting before we can merge.

bifurcation (approved):

PR #372, opened 2020-07-27
Commit Generation Clarifications

chelseakomlo:
This PR tries to clarify these paragraphs by the following: 

- Differentiating between HPKE public/private keys when describing how these keys are rotated for a commit
- Clarifying who does the actions associated with a commit
- Adding consistency fixes between the two paragraphs

Let me know if these fixes are not quite right, as this PR tries to clarify where I had questions when reading.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -698,13 +698,13 @@ they receive the private keys for nodes, as described in
 
 ## Ratchet Tree Evolution
 
-When performing a Commit, the leaf KeyPackage of the committer and
-its direct path to the root are updated with new secret values.  The
+When performing a Commit, the generator of the Commit updates its leaf
+KeyPackage and its direct path to the root with new secret values.  The
 HPKE leaf public key within the KeyPackage MUST be derived from a freshly
 generated HPKE secret key to provide post-compromise security.
 
 The generator of the Commit starts by using the freshly generated HPKE secret
-key "leaf_hpke_secret" associated with the updated leaf KeyPackage (see
+key "leaf_hpke_secret" associated with its now-updated leaf KeyPackage (see
 {{key-packages}}) to compute "path_secret\[0\]" and generate a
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -698,13 +698,13 @@ they receive the private keys for nodes, as described in
...
-generated value to provide post-compromise security.
+When performing a Commit, the generator of the Commit updates its leaf
+KeyPackage and its direct path to the root with new secret values.  The
+HPKE leaf public key within the KeyPackage MUST be derived from a freshly
+generated HPKE secret key to provide post-compromise security.
 
-The generator of the Commit starts by using the HPKE secret key
-"leaf_hpke_secret" associated with the new leaf KeyPackage (see
+The generator of the Commit starts by using the freshly generated HPKE secret
+key "leaf_hpke_secret" associated with its now-updated leaf KeyPackage (see
```
bifurcation, line 707: This isn't quite right.  `leaf_hpke_secret` is a symmetric secret value, from which an HPKE secret key is derived, as illustrated below:

```
    path_secret[0] --> node_priv[0], node_pub[0]
         ^
         |
   leaf_hpke_secret
```

Maybe it would be helpful to annotate that diagram to reinforce that the `[0]` is the leaf?

```
    path_secret[0] --> node_priv[0], node_pub[0]
         ^             = leaf_priv, leaf_pub
         |             ~> leaf_key_package
         |
   leaf_hpke_secret
```

chelseakomlo: This annotation would be helpful.

bifurcation: Actually, as you pointed out to me offline, this is inconsistent with the diagram below.  I think what it should be instead is:

```
       path_secret[1] --> node_priv[1], node_pub[1]
            ^
            |
       path_secret[0] --> node_priv[0], node_pub[0]
            ^
            |
       leaf_secret    --> leaf_priv, leaf_pub
                          ~> leaf_key_package
```

Issue #366, opened 2020-07-24
Add extensions to the Commit message


bifurcation:
[This issue is part of a feature breakdown based on #336]

Applications might want to signal to group members  processing a commit that they should take certain actions in addition to ratcheting the group forward.  The most obvious example is folding in a PSK for the epoch, but you can also imagine addressing things like transcript consistency or updatable PKE through this channel.

This should be straightforward syntactically, but will also require updating the extensibility section to make clear that Commit extensions can also in principle affect the GroupContext extensions.  In other words:

* Each group member associates a list of extensions with the group, which are confirmed via the GroupContext
* New joiners get configured with these via extensions in the GroupInfo
* Existing members can find out about changes to the group's extensions via extensions in the Commit

kkohbrok:
I think making commit messages extensible is a great idea. However, I'd like to understand better what we want to use the extensions first. 

> Applications might want to signal to group members processing a commit that they should take certain actions in addition to ratcheting the group forward.

I thought that was what Proposals are for. I think for mechanisms that are part of the base protocol spec, there is no reason not to extend the list of proposals.

Regarding the nature of extensions: Do I understand correctly, that an extension is a property of the group first and foremost? And extensions in the commit are meant to alert other group members to changes in those properties? Or are extensions in the commit messages a mechanism to allow for the inclusion of additional data in the commits? Or both? I feel like the meaning of the word "extension" is a little stretched here.

Since we're discussing this also in the context of PSK injections: Do you think PSK injections should be something that can be switched on and off for any given group? If so, why? I would consider it part of the base protocol and do not see what benefits we get from disallowing it in a group.

bifurcation:
It's a fair point, that we have basically two extension points here -- proposal types and extensions.  I think the theory I had in mind was something like:

1. Proposals describe modifications you make to the tree before you use it for TreeKEM (to get the path) (if you do)
2. The DirectPath describes the outcome of TreeKEM.Encap
3. Extensions tell you anything else you need to do

If that philosophy makes sense, I can update the PR to make that clearer.

Note that proposal extensibility is currently broken because we don't actually have a way for the Commit to include any Proposals with new proposal types.  I've filed #383 for that.

bifurcation:
On the nature of extensions: My thinking here is that an "extension" is purely syntactic; its semantics are defined entirely by its ExtensionType.  One thing that extensions do (in the context of GroupContext) is express what attributes/parameters the group has agreed to.  But they can also do other things.

[...]

PR #376, opened 2020-07-28
Negotiate PSKs

### LINKED ISSUE #367: Negotiate PSKs
[This issue is part of a feature breakdown based on #336]

There are a few subissues here:

# 1. Signaling which PSK(s) to add to the key schedule

When a PSK is to be added to the key schedule in an epoch, the sender of the Commit needs to signal this to existing members in a Commit, and to any new members in the GroupInfo.  In both cases, there should be an extension to signal this.  This would work like the [TLS ServerHello PSK extension](https://tools.ietf.org/html/rfc8446#section-4.2.11), which simply indicates which PSK is being used.  In the MLS case, we would probably include the full ID instead of just an index, since there's not a clear list like in TLS.  And just like the server's ability to complete the TLS handshake proves that it has the PSK, the confirmation value in MLS would validate the same.

```
struct {
    opaque identity<1..2^16-1>;
} PreSharedKeyExtension;
```

# 2. Signaling which PSKs are available

It might also be useful for clients to advertise in a KeyPackage extension what PSKs they have available.  This would be analogous to the ClientHello PSK extension, which includes a binding value to prove that the client has the PSK.  If we define such an extension, we will need to decide (a) whether we need such a binding and (b) what parts of the KeyPackage should be covered by the binding MAC.

Unlike TLS, however, the set of available PSKs might change over the lifetime of the group.  Since the KeyPackages in the leaves of the tree can be modified, in principle, clients could signal the availability of new PSKs by updating, and you would get a clear assurance that everyone has the key before you try to use it.  But this seems pretty complex, and for not much benefit if the application can coordinate which PSKs are available.

# 3. Structuring PSK input / multiple PSKs

It may be desirable for the input to the key schedule not to be a PSK directly, but some derived value that adds some context to the PSK.  This is the approach taken in the [TLS external PSK draft](https://tools.ietf.org/html/draft-ietf-tls-external-psk-importer-05) and in #336.  We might also want to support the application of multiple PSKs in a single epoch.

For the former need, we would need to define what context is added, how it is signaled in the relevant extensions, and how it is combined with the PSK itself to derive the ultimate PSK.  For the latter need, we need a way to signal multiple PSKs and their order, and an algorithm for combining them all into one value to be added to the key schedule.  (The schemes in both #336 and #337 are relevant here.)

bifurcation:
Fixes #367

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1893,6 +1893,7 @@ struct {
     ProposalID adds<0..2^32-1>;
 
     optional<DirectPath> path;
+    Extension extensions<0..2^16-1>;
 } Commit;
 ~~~~~
 
@@ -2275,15 +2276,21 @@ places:
 
 * In KeyPackages, to describe client capabilities and aspects of their
   participation in the group (once in the ratchet tree)
-* In the Welcome message, to tell new members of a group what parameters are
-  being used by the group
+* In the GroupInfo object carried in a Welcome message, to tell new members of a
+  group what parameters are being used by the group
 * In the GroupContext object, to ensure that all members of the group have the
   same view of the parameters in use
+* In Commit messages, to describe changes to the group's parameters
 
 In other words, clients advertise their capabilities in KeyPackage
 extensions, the creator of the group expresses its choices for the group in
 Welcome extensions, and the GroupContext confirms that all members of the group
 have the same view of the group's extensions.
+
+In addition to the function of managing the metadata associated to the group as
+described above, extensions in all four of these places may be used to attach
+additional information to a message, convey additional instructions, etc.,
+subject to the compatibility constraints below.
 
 This extension mechanism is designed to allow for secure and forward-compatible
 negotiation of extensions.  For this to work, implementations MUST correctly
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2321,23 +2369,26 @@ handle extensible fields:
...
+* If any extension in a Commit message is unrecognized, then the client MUST
+  reject the Commit.  The sender of a Commit message MUST verify that the
+  extension is supported by all members of the group, as indicated in the
+  KeyPackages in the leaves of the tree.
+
+* The extensions populated into a GroupContext object by a joining member are
+  constructed from those in the GroupInfo object, according to the definitions
+  of those extensions.  Extensions in a Commit object may define modifications
+  to the list of GroupContext extensions, which recipients of the Commit apply
+  on receiving it.
```
kkohbrok, line 2381: As I noted on the mailinglist, the issue with the PSKId being (part of) an extension and those extensions in turn being part of the `GroupInfo` object, means that the PSK can't be required for decrypting that `GroupInfo` object in the first place. In #336 , we suggested that the PSKId be part of the `GroupSecrets` object along with the other secrets. This would also ensure compatibility with the pending nPRF key schedule proposal.

bifurcation:
Closing this one for now.  After discussion with @kkohbrok, he's going to refactor the relevant bits out of #336 once we have the required tools from fixing #366.

PR #378, opened 2020-07-30
Readability edits for section 'Synchronizing Views of the Tree'

chelseakomlo:
This PR does the following: 

- Adds a lead-in sentence describing how this section builds on the prior. 
- Differentiates between public keys for nodes and public keys for leaves (users)
- Clarifies when secret key material needs to be deleted. 

I also wanted to make a couple other changes, but wanted feedback first. 

1. This section describes the private value corresponding to a node's public key as 'ns[0]' (or node secret), whereas the prior section referred to it as 'np[0]' (or node_priv). It would be better to have these terms be consistent.
2.  On line 773, it says that the client "...transmits a handshake message containing a set of **public** values" and on line 776 it says that "other members of the group can use these **public** values to update". However, this values seems to be contradictory with the value listed on line 785. 
3. The current terminology is unclear regarding 1) a Commit to the tree, 2) an update to the tree, and and 3) a handshake message. I think these are very similar but using the same language consistently would be better (such as defining a Commit as what updates the tree, and then describing how the handshake message is different from a Commit message, which is described earlier in the spec).

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -825,7 +825,7 @@ values:
 | pk(ns\[1\])  | E(pk(C), ps\[1\]), E(pk(D), ps\[1\]) |
 | pk(ns\[0\])  | E(pk(A), ps\[0\])                    |
 
-In this table, the value pk(ns\[1\]) represents the public key
+In this table, the value pk(ns\[X\]) represents the public key
 derived from the node secret X, whereas pk(X) represents the public leaf key
 for user X.  The value E(K, S) represents
 the public-key encryption of the path secret S to the
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -828,6 +828,8 @@ derived from the node secret X.  The value E(K, S) represents
 the public-key encryption of the path secret S to the
 public key K.
 
+After applying the commit, all group members MUST delete outdated path secrets.
+
 
 # Cryptographic Objects
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -823,8 +823,9 @@ values:
 | pk(ns\[1\])  | E(pk(C), ps\[1\]), E(pk(D), ps\[1\]) |
 | pk(ns\[0\])  | E(pk(A), ps\[0\])                    |
 
-In this table, the value pk(X) represents the public key
-derived from the node secret X.  The value E(K, S) represents
+In this table, the value pk(ns\[1\]) represents the public key
+derived from the node secret X, whereas pk(X) represents the public leaf key
+for user X.  The value E(K, S) represents
 the public-key encryption of the path secret S to the
 public key K.
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -766,7 +766,9 @@ delete the leaf_secret.
 
 ## Synchronizing Views of the Tree
 
-The members of the group need to keep their views of the tree in
+After generating a Commit as described in the prior section, the generator of
+the Commit must broadcast this update to other members of the group, who will need to
+apply it to keep their views of the tree in
 sync and up to date.  When a client commits a change to the tree
 (e.g., to add or remove a member), it transmits a handshake message
 containing a set of public
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -835,7 +835,7 @@ After processing the update, each recipient MUST delete outdated key material,
 specifically:
 
 * The path secrets used to derive each updated node key pair.
-* Each outdated key pair that was replaced by the update.
+* Each outdated node key pair that was replaced by the update.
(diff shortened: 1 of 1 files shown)


bifurcation:
Thanks for looking through this, @chelseakomlo.  There's a lot of historical baggage here, so it's good to get it cleaned up.  I'll review the diff in a moment, but to respond to your points:

1. Just to confirm terminology here, there are three values in play, (1) a symmetric secret, (2) a KEM private key derived from the secret, and (3) the corresponding public key.  It _should_ be the case that you only need to store the latter two in the tree (priv+pub), since the secrets are used transiently: For deriving the tree key pairs, for deriving the commit secret, and for initializing new members added in an epoch.  [My implementation](https://github.com/cisco/mlspp/blob/master/include/treekem.h#L42) stores the node secrets to simplify these operations, but it feels like it should be possible to arrange things so that this isn't necessary.  So I think the right answer here is to just refer to node private keys.

2. Line 785 is correct, the Commit contains a public key and encrypted copies of the path secret for each node.  So the text should refer to public **and encrypted private** values as necessary.

3. The right taxonomy here is: "Handshake message" encompasses `Proposal` and `Commit`.  Processing the Commit is the thing that causes an update to the tree; more specifically, applying a DirectPath.  So we should only talk about handshake messages when we want to cover both Proposals and Commits, and most instances of "update" should probably refer to either Commit or DirectPath.

draft-ietf-mls-protocol.md
```diff
@@ -766,7 +766,9 @@ delete the leaf_secret.
 
 ## Synchronizing Views of the Tree
 
-The members of the group need to keep their views of the tree in
+After generating a Commit as described in the prior section, the generator of
+the Commit must broadcast this update to other members of the group, who will need to
```
bifurcation, line 770: `s/will need to//`

draft-ietf-mls-protocol.md
```diff
@@ -766,7 +766,9 @@ delete the leaf_secret.
...
+apply it to keep their views of the tree in
```
bifurcation, line 771: Might rephrase as "apply it to their local views of the tree to keep them in sync with the sender's."

draft-ietf-mls-protocol.md
```diff
@@ -766,7 +766,9 @@ delete the leaf_secret.
 
 ## Synchronizing Views of the Tree
 
-The members of the group need to keep their views of the tree in
+After generating a Commit as described in the prior section, the generator of
```
bifurcation, line 769: This stuff might be better phrased in terms of a `DirectPath` (which maybe we should rename to `UpdatePath`?), since a Commit no longer has to have an update of this character.

draft-ietf-mls-protocol.md
```diff
@@ -823,11 +825,14 @@ values:
...
 
-In this table, the value pk(X) represents the public key
-derived from the node secret X.  The value E(K, S) represents
+In this table, the value pk(ns\[X\]) represents the public key
+derived from the node secret X, whereas pk(X) represents the public leaf key
+for user X.  The value E(K, S) represents
 the public-key encryption of the path secret S to the
 public key K.
 
+After applying the commit, all group members MUST delete outdated path secrets.
```
bifurcation, line 834: Unclear whether "outdated" here means "from before the Commit" or "transmitted in the Commit, but now obsolete because the keys have been generated from them".  Like I said above, I think we can focus on the latter (so that you'll never even have the former stored), we just need to make sure that the wording of the requirement doesn't interfere with the secondary uses of the path secrets.

draft-ietf-mls-protocol.md
chelseakomlo: `UpdatePath` sounds like a better name. I'll do this renaming in a separate PR, but will rephrase this section to refer to a `DirectPath` as opposed to a `Commit` in the meantime.

chelseakomlo: Actually, you know, I think it would be better to do this change in a separate commit as well, as it will involve re-writing the "Ratchet Tree Evolution" section to refer to `DirectPath` as well, no?

draft-ietf-mls-protocol.md
chelseakomlo: Do you mind suggesting phrasing here? This doesn't match my mental model of "when replacing key material in the tree after receiving a `DirectPath`, the outdated key material (that was replaced by key material in the `DirectPath` message) should be deleted, but there could be something I'm profoundly missing.

bifurcation: The question is: if your nodes have `(secret, sk, pk)`, and you move from `(secret0, sk0, pk0) -> (secret1, sk1, pk1)`, do you need to delete secret0, secret1, or both?  (If you delete `secret1`, then you never have a `secret0` to delete)

draft-ietf-mls-protocol.md
bifurcation: Given this and the discussion below, do you want to close this and come back after the other refactors are done?

PR #385, opened 2020-08-06
Require the initial init_secret to be a random value

ericcornelissen:
This suggestion is based on a discussion in the MLS mailing list [[1]](https://mailarchive.ietf.org/arch/msg/mls/qrpAoK7aNSzKxX4O60pOLvM3cp8). It prevents issues that currently arrise when the very first commit of a group is an "add-only" commit because both the `commit_secret` and `init_secret` would be **0**.

Let me know if the wording or anything else should be changed.

/cc @bifurcation

---

1. https://mailarchive.ietf.org/arch/msg/mls/qrpAoK7aNSzKxX4O60pOLvM3cp8

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1271,7 +1271,7 @@ the below diagram:
 When processing a handshake message, a client combines the
 following information to derive new epoch secrets:
 
-* The init secret from the previous epoch.
+* The init secret from the previous epoch
 * The commit secret for the current epoch
 * The GroupContext object for current epoch
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1714,7 +1714,7 @@ The creator of a group MUST take the following steps to initialize the group:
   * Tree hash: The root hash of the above ratchet tree
   * Confirmed transcript hash: 0
   * Interim transcript hash: 0
-  * Init secret: a fresh random value
+  * Init secret: a fresh random value of size `KDF.Nh`
 
 * For each member, construct an Add proposal from the KeyPackage for that
   member (see {{add}})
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1714,7 +1714,7 @@ The creator of a group MUST take the following steps to initialize the group:
   * Tree hash: The root hash of the above ratchet tree
   * Confirmed transcript hash: 0
   * Interim transcript hash: 0
-  * Init secret: 0
+  * Init secret: a fresh random value
```
bifurcation, line 1717:
```suggestion
  * Init secret: a fresh random value of size `KDF.Nh`
```

... just to ensure you have enough entropy here.

ericcornelissen:
I believe this change invalidates the last paragraph of the [Group Creation section](https://github.com/mlswg/mls-protocol/blob/d1d5f56a5d83781042d19c830497ab5aa104907f/draft-ietf-mls-protocol.md#group-creation), as it is impossible to verify the group creation process for the joiner without knowing the random `init_secret` that the group creator used.


> A new member receiving a Welcome message can recognize group creation if the number of entries in the members array is equal to the number of leaves in the tree minus one. A client receiving a Welcome message SHOULD verify whether it is a newly created group, and if so, **SHOULD verify that the above process was followed by reconstructing the Add and Commit messages and verifying that the resulting transcript hashes and epoch secret match those found in the Welcome message.**

tomtau:
maybe one alternative is to keep the zero `init_secret`, but make it compulsory to fill in the path for this group creation welcome/commit (so that `commit_secret` isn't zero)?

ericcornelissen:
That would also work @tomtau (I also suggested that as an alternative in my original report in the mailing list), to me it is a little bit less elegant though. Personally, I also wonder why it is necessary to verify that group creation just happened. (Moreover, this way of recognizing group creation would prevent any group from which all but one person left from continuing, which my be confusing to end-users)

PR #387, opened 2020-08-10
Make composite types 2^32 and make 0 invalid for enums

Bren2010:
- 0 is invalid for all enums except a few, and I wasn't sure if it mattered? So I made them consistent
- `DirectPath.nodes` needs to be 2^32 to fit the largest DirectPathNode
- Make the `extensions` array 2^32 as well to fit the largest Extension, and also 2^16 for *all* extensions seems small

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -954,7 +954,8 @@ except for the signature field.
 
 ~~~~~
 enum {
-    mls10(0),
+    reserved(0),
+    mls10(1),
     (255)
 } ProtocolVersion;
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -954,7 +954,8 @@ except for the signature field.
 
 ~~~~~
 enum {
-    mls10(0),
+    invalid(0),
```
seanturner, line 957: Change to reserved.

draft-ietf-mls-protocol.md
```diff
@@ -2257,8 +2258,9 @@ type `ratchet_tree`, containing a `ratchet_tree` object of the following form:
 
 ~~~~~
 enum {
-    leaf(0),
-    parent(1),
+    invalid(0),
```
seanturner, line 2261: Change to reserved.

PR #388, opened 2020-08-11
Clarify the transcript hash computation

### LINKED ISSUE #386: Remove remaning raw Hash or add hash function to ciphersuite
In #364, we changed from using HKDF to using the KDF from HPKE, which I thought removed the need for the ciphersuite to specify a hash function. As a result, ciphersuites no longer needed to specify a hash.

However, the transcript computations still have a raw `Hash()` invocation, not a KDF invocation.  So we need to either:

1. Change the transcript hash computations to use the KDF, or
2. Clarify that ciphersuites still specify a hash function (as before)

bifurcation:
Addresses ambiguity pointed out on the mailing list by @ericcornelissen.

Fixes #386

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -67,6 +67,8 @@ informative:
     target: https://secg.org/sec1-v2.pdf
     date: 2009
 
+  SHS: DOI.10.6028/NIST.FIPS.180-4
+
 --- abstract
 
 Messaging applications are increasingly making use of end-to-end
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1173,11 +1175,13 @@ struct {
     opaque signature<0..2^16-1>;
 } MLSPlaintextCommitAuthData;
 
+interim_transcript_hash_[0] = 0
```
raphaelrobert, line 1178: Still not clear to me what `0` means here.

PR #391, opened 2020-08-14
Rename DirectPath to UpdatePath

chelseakomlo:
This is a simple rename of `DirectPath` to `UpdatePath` to be more descriptive of the action being performed by the message.

I'll open a separate PR after this PIR is merged with updates to the spec where it uses `Commit` when instead it should say `UpdatePath` (since not every `Commit` actually ratchets forward key material, as I understand). Some examples of where these clarifications should happen are: 

- Line 1203 should instead say "each MLS UpdatePath message" 
- The sections "Ratchet Tree Evolution" and "Sychronizing Views of the Tree" should describe the process of generating the key material for an `UpdatePath` message

cc @bifurcation

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1198,7 +1198,7 @@ without having to store the whole MLSPlaintextCommitAuthData structure.
 As shown above, when a new group is created, the `interim_transcript_hash` field
 is set to the zero-length octet string.
 
-## Direct Paths
+## Update Paths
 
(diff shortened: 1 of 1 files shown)


Bren2010:
DirectPath corresponds to the graph theory term "direct path" of a leaf with the root

bifurcation:
"Direct path" is still a thing, and the object in question here operates along a direct path, but I think "UpdatePath" indicates more clearly what it's doing.

PR #393, opened 2020-08-14
Application messages MUST be encrypted

raphaelrobert:


draft-ietf-mls-protocol.md
```diff
@@ -1449,10 +1449,11 @@ The two main structures involved are MLSPlaintext and MLSCiphertext.
 MLSCiphertext represents a signed and encrypted message, with
 protections for both the content of the message and related
 metadata.  MLSPlaintext represents a message that is only signed,
-and not encrypted.  Applications SHOULD use MLSCiphertext to encode
-both application and handshake messages, but MAY transmit handshake
-messages encoded as MLSPlaintext objects in cases where it is
-necessary for the delivery service to examine such messages.
+and not encrypted.  Applications MUST use MLSCiphertext to encrypt 
+application messages and SHOULD use MLSCiphertext to encode
```
bifurcation, line 1453: I don't think this MUST makes sense -- for example, there may be cases where you want to take an MLS key exchange and protect application data with an exported key.

Instead, can we just delete "both application and" here?  I think the intent of the sentence was to encourage encryption of handshake messages, not to say anything about application messages.

raphaelrobert: > I don't think this MUST makes sense -- for example, there may be cases where you want to take an MLS key exchange and protect application data with an exported key.

If you just want to use MLS for key exchange and then use key material from the exporter that's fine, but that's orthogonal to using Application Messages.

I really want to make sure that client never send unencrypted Application Messages. If you don't care about confidentiality you can always use the AAD for that.

bifurcation: Actually, on re-read, seems clear from context.

PR #396, opened 2020-08-18
Authenticate group membership in MLSPlaintext

bifurcation:
[Joël pointed out on the list](https://mailarchive.ietf.org/arch/msg/mls/LFz5nC4BnNTf7-OGIac5_QdlCRs/) that MLSPlaintext does not currently authenticate the sender's membership on the group, and as a result, some important attacks arise.  Discussion on the list indicated that adding a MAC with a key derived from the key schedule would provide this authentication.  This PR implements that suggestion.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1651,7 +1651,7 @@ verifying the content signature.
 
 The following sections describe the encryption and signing processes in detail.
 
-## Content Signing
+## Content Authentication
 
 The `signature` field in an MLSPlaintext object authenticates the sender of the
 MLSPlaintext message.  The input to the signature is an MLSPlaintextTBS
@@ -1660,10 +1660,10 @@ context for the signature that associates to the group messages sent within the
 group.
 
 In cases where the sender is a member of the group, the context contains the
-GroupContext for the current epoch and a "membership token" that authenticates
+GroupContext for the current epoch and a "membership tag" that authenticates
 that the sender is a member of the group.  As a result, for messages sent within
-a group, the signature authenticates not only the member's individual identity,
-but also their membership in the group.
+a group, the signature authenticates the member's individual identity, and the
+membersip tag authenticates their membership in the group.
 
 ~~~~~
 struct {
@@ -1738,6 +1738,10 @@ struct {
     opaque padding<0..2^16-1>;
 } MLSCiphertextContent;
 ~~~~~
+
+Note that the `membership_tag` is not carried forward from MLSPlaintext to
+MLSCiphertext.  In the context of an MLSCiphertext, the AEAD authentication
+provides the function that the `membership_tag` provides for MLSPlaintext.
 
 The key and nonce used for the encryption of the message depend on the
 content type of the message.  The sender chooses the handshake key for a
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1336,6 +1336,7 @@ A number of secrets are derived from the epoch secret for different purposes:
 | `application_secret`  | "app"         |
 | `exporter_secret`     | "exporter"    |
 | `confirmation_key`    | "confirm"     |
+| `membership_key`      | "membership"  |
 
 ## Pre-Shared Keys
 
@@ -1505,6 +1506,7 @@ struct {
     }
 
     opaque signature<0..2^16-1>;
+    opaque membership_tag<0..2^8-1>;
 } MLSPlaintext;
 
 struct {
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1611,6 +1623,11 @@ struct {
           Commit commit;
           opaque confirmation_tag<0..255>;
     }
+} MLSPlaintextContent;
+
+struct {
+    MLSPlaintextSignatureContext context;
+    MLSPlaintextTBS tbs;
 } MLSPlaintextTBS;
```
ericcornelissen, line 1631: Looks like something went wrong here. The new `MLSPlaintextContext` and `MLSPlaintextContent` structs are not used and the `MLSPlaintextTBS` struct is now self-referencing.

If I had to guess, I think you meant:

```suggestion
struct {
    MLSPlaintextContext context;
    MLSPlaintextContent content;
} MLSPlaintextTBS;
```

draft-ietf-mls-protocol.md
```diff
@@ -1614,14 +1615,21 @@ struct {
...
 ~~~~~
 
-<!-- OPEN ISSUE: group_id and epoch are duplicated in the TBS and in
-GroupContext. Think about how to de-duplicate. -->
```
Bren2010, line 1618: You didn't fix this issue?

draft-ietf-mls-protocol.md
```diff
@@ -1614,14 +1615,21 @@ struct {
...
-by the ciphersuite in use.  The plaintext input contains content and
-signature of the MLSPlaintext, plus optional padding.  These values
-are encoded in the following form:
+~~~~~
+membership_token = KDF.Extract(confirmation_key, MLSPlaintextTBS);
+~~~~~
+
+The ciphertext field of the MLSCiphertext object is produced by supplying the
+inputs described below to the AEAD function specified by the ciphersuite in use.
+The plaintext input contains content and signature of the MLSPlaintext, plus
```
Bren2010, line 1629: "contains the content"

draft-ietf-mls-protocol.md
```diff
@@ -1614,14 +1615,21 @@ struct {
...
+optional padding. (The membership token is not included in the encrypted
+content, because the AEAD encryption authenticates the sender's membership in
```
Bren2010, line 1631: "the AEAD authenticates"

draft-ietf-mls-protocol.md
```diff
@@ -1614,14 +1615,21 @@ struct {
...
+than `member`, this field MUST be set to the zero-length octet string.  For
+messages sent by members, it MUST be set to the following value:
 
-The ciphertext field of the MLSCiphertext object is produced by
-supplying the inputs described below to the AEAD function specified
-by the ciphersuite in use.  The plaintext input contains content and
-signature of the MLSPlaintext, plus optional padding.  These values
-are encoded in the following form:
+~~~~~
+membership_token = KDF.Extract(confirmation_key, MLSPlaintextTBS);
```
Bren2010, line 1624: Using a KDF output for authentication feels wrong

draft-ietf-mls-protocol.md
bifurcation: I thought about it, and decided not to worry about it.  If you feel strongly, we can keep it.

draft-ietf-mls-protocol.md
bifurcation: We already do this elsewhere:
```
MLSPlaintext.confirmation_tag =
    KDF.Extract(confirmation_key, GroupContext.confirmed_transcript_hash)
```

This seems OK with HKDF, since Extract is just HMAC.  But I agree that we should confirm that this generalizes properly.

bifurcation: Agreements on the 2020-08-25 call:
* Use an explicit MAC here instead of KDF.Extract
* Use a separate key from the key schedule, different from the confirmation_key

```
epoch
  |
  +-- confirmation_key
  |
  +-- membership_key
  |
  +-- authentication_secret (todo)
```

Bren2010:
Lgtm. Rebase and I'll properly approve.

bifurcation:
Generally not a fan of rebasing, but here you go.

Bren2010 (approved):

draft-ietf-mls-protocol.md
```diff
@@ -1543,13 +1545,17 @@ The following sections describe the encryption and signing processes in detail.
...
+MLSPlaintext message.  The input to the signature is an MLSPlaintextTBS
+structure, which encodes the content and metadata of the message as well a
+context for the signature that associates to the group messages sent within the
+group.
+
+In cases where the sender is a member of the group, the context contains the
+GroupContext for the current epoch and a "membership token" that authenticates
+that the sender is a member of the group.  As a result, for messages sent within
+a group, the signature authenticates not only the member's individual identity,
+but also their membership in the group.
```
Bren2010, line 1558: Actually this isn't quite right because the signature doesn't cover the membership token

draft-ietf-mls-protocol.md
```diff
@@ -1576,18 +1582,35 @@ struct {
           opaque confirmation_tag<0..255>;
     }
 } MLSPlaintextTBS;
+
```
Bren2010, line 1585: Please break this out into its own sub-section titled "Plaintext Authentication" or similar, and an explainer that `membership_tag` is only necessary when sending MLSPlaintext over-the-wire

psyoptix (changes requested):

PR #397, opened 2020-08-20
Always use path secrets with KDF.Expand

bifurcation:
@ericcornelissen pointed out on the mailing list that path secrets are currently used with both KDF.Expand and KEM.DeriveKeyPair.  This PR introduces an extra KDF.Expand call so that each key is used with exactly one algorithm.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -727,7 +727,7 @@ path_secret[n] = ExpandWithLabel(path_secret[n-1],
                                    "path", "", KEM.Nsk)
 node_secret[n] = ExpandWithLabel(path_secret[n],
                                    "node", "", KEM.Nsk)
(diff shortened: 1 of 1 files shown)


ericcornelissen:
I believe it was @chrisbrzuska who that [pointed this out](https://mailarchive.ietf.org/arch/msg/mls/EQLdrkuVBeB5ExDcFx4EMbJcpOg/), not me. Also, shouldn't the `node_priv[n]` and `node_pub[n]` derivation be updated as well?

```diff
- node_priv[n], node_pub[n] = KEM.DeriveKeyPair(path_secret[n])
+ node_priv[n], node_pub[n] = KEM.DeriveKeyPair(node_secret[n])
```

bifurcation:
Ah, sorry!  I'll make that change.

bifurcation:
Approved at the interim 2020-08-25.

PR #401, opened 2020-09-14
Derive handshake keys from a secret tree as well.

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -1359,8 +1359,50 @@ Note that, as a PSK may have a different lifetime than an update, it
...
+
+For the generation of encryption keys and nonces, the key schedule begins with a
+particular root secret and derives a tree of secrets with the same structure as
+the group's ratchet tree. Each leaf in the Secret Tree is associated with the
+same group member as the corresponding leaf in the ratchet tree. Nodes are also
+assigned an index according to their position in the array representation of the
+tree (described in {{tree-math}}). If N is a node index in the Secret Tree then
+left(N) and right(N) denote the children of N (if they exist).
+
+The root secret of the tree is either `handshake_secret` if the tree is for
```
bifurcation, line 1372: I think that on the last interim call, we agreed to split off at the leaves of the tree, instead of at the root.  So you would have one secret tree per epoch, and the leaves would split into application and handshake hash chains.

@Bren2010 if you could make that update, I think this is good to go.

Bren2010: Updated, thank you!

kkohbrok (approved):

bifurcation (approved):

PR #402, opened 2020-09-16
Allow for credentials supporting multiple signature schemes

kkohbrok:
The credentials we have defined at the moment only contain a single public key and no information regarding what signature scheme that keypair works with (even though the Spec actually says it should include the signature scheme as well).

It would be nice, however, to support potential CredentialTypes that, similar to KeyPackages, include multiple supported signature schemes, each accompanied by a corresponding public key.

The Key that should be used in any given group is then determined by the CipherSuite of the group (which includes a Signature Scheme). When choosing a KeyPackage of a new member, it has to be one that contains a Credential which supports that Signature Scheme.

Note, that I'm not suggesting that this be the case in all Credentials, or even that we change the BasicCredential.

I understand that it's possible to have multiple credentials per identity, but in some authentication settings it can be beneficial to have a 1-to-1 mapping between Credential and identity.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -889,26 +889,33 @@ The ciphersuites are defined in section {{mls-ciphersuites}}.
 
 ## Credentials
 
-A member of a group authenticates the identities of other
-participants by means of credentials issued by some authentication
-system, like a PKI.  Each type of credential MUST express the
-following data:
+A member of a group authenticates the identities of other participants by means
+of credentials issued by some authentication system, like a PKI. Each type of
+credential MUST express the following data in the context of the group it is
+used with:
 
-* The public keys of one or more signature key pairs
+* The public key of a signature key pair matching the SignatureScheme specified
+  by the CipherSuite of the group
 * The identity of the holder of the private keys
-* The signature schemes corresponding to the individual key pairs (only if more
-  than one key pair is included)
 
 Credentials MAY also include information that allows a relying party
 to verify the identity / signing key binding.
 
+Additionally, Credentials SHOULD specify the signature scheme corresponding to
+each contained public key.
+
 ~~~~~
+
+// See IANA registry for registered values
+uint16 SignatureScheme;
+
 // See IANA registry for registered values
 uint16 CredentialType;
 
 struct {
     opaque identity<0..2^16-1>;
     opaque identity_key<0..2^16-1>;
+    SignatureScheme signature_scheme;
 } BasicCredential;
 
 struct {
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -905,7 +905,6 @@ Additionally, Credentials SHOULD specify the signature scheme corresponding to
 each contained public key.
 
 ~~~~~
-
 // See IANA registry for registered values
 uint16 SignatureScheme;
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -894,9 +894,10 @@ participants by means of credentials issued by some authentication
 system, like a PKI.  Each type of credential MUST express the
 following data:
 
-* The public key of a signature key pair
-* The identity of the holder of the private key
-* The signature scheme that the holder will use to sign MLS messages
+* The public keys of one or more signature key pairs
+* The identity of the holder of the private keys
+* The signature schemes corresponding to the individual key pairs (only if more
+  than one key pair is included)
 
 Credentials MAY also include information that allows a relying party
 to verify the identity / signing key binding.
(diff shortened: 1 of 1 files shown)


Bren2010:
> It would be nice, however, to support potential CredentialTypes that, similar to KeyPackages, include multiple supported signature schemes, each accompanied by a corresponding public key.

KeyPackages don't support multiple signature schemes?

> I understand that it's possible to have multiple credentials per identity, but in some authentication settings it can be beneficial to have a 1-to-1 mapping between Credential and identity.

In what scenario is it beneficial?

kkohbrok:
> KeyPackages don't support multiple signature schemes?

Ah, right. I think that was the case in the past, but you're right, they don't.

> In what scenario is it beneficial?

Of course it's always possible to just bundle a bunch of certificates, each supporting a different signature scheme and then treat that as your "Multi-Scheme Credential" in the grand scheme of things. 

However, if you end up rotating them a lot, where you have to sign them and potentially store the whole chain for a longer period of time, being able to support multiple signature schemes in only one credential means less overhead.

Bren2010:
> However, if you end up rotating them a lot, where you have to sign them and potentially store the whole chain for a longer period of time, being able to support multiple signature schemes in only one credential means less overhead.

Credentials are stored in each group's ratchet tree though, so you're keeping a bunch of unnecessary key material in the group's tree which everyone has to store

kkohbrok:
That's fair. And it's not something you have to do if you don't want to. I'm just preparing for MLS to support that type of Credential in general.

br-hale:
To support hybrid PQ/standard schemes at any point in the future, allowing for multiple schemes is reasonable (in addition to the above arguments for it). Most applications probably will not use the feature, especially to start with, but providing support for just the option is not going to cause any of the problems noted.

Bren2010:
> To support hybrid PQ/standard schemes at any point in the future, allowing for multiple schemes is reasonable

PQ schemes can be used by switching the entire group to a new ciphersuite using those schemes. Requiring everybody in the group to store a bunch of unused key material isn't reasonable

br-hale:
The statement was not about PQ, but hybrid PQ/standard schemes. Those require use to two signature schemes, for which both types of keying material are used.

Bren2010:
Then you need a ciphersuite that specifies a hybrid scheme.

br-hale:
It would be ideal to have a ciphersuite specifying a single hybrid scheme, but even that does not imply use of a single credential, unfortunately. In fact, hybrid schemes may rely on two separate credentials, with the ciphersuite specifying the hybrid algorithm for combining them. Since the "main" method of building a solo hybrid credential is under patent, there is also additional motivation for applications to use two.

bifurcation:
I concur with Brendan here.  To be consistent with our earlier decisions to (a) keep KeyPackages to one ciphersuite and (b) include the signature algorithm in the ciphersuite, we should keep credentials to one signature algorithm as well.

Propose closing this PR without merging.

kkohbrok:
I don't think it's a good idea to not list the `SignatureScheme` in the credential. It only appears to be redundant in the context of a group, but credentials are also going to exist in contexts outside of individual groups, e.g., the AS. 

If we don't even potentially allow multiple signature schemes in one credential, a client that wants to support multiple signature schemes has to have multiple credentials. Now I have multiple credentials flying around, each with a `signature_key` field, but no information which credential belongs to which scheme, which could result in the a key being used with the wrong signature scheme. It's not clear what side effect such an "agile" use of key material is going to have.

If we're not going to support multiple schemes in a credential (not even proprietary ones), then let's at least keep the information which signature scheme the `signature_key` belongs to.

bifurcation:
Discussed at 2020-10-06 interim, TODO:
* Re-add SignatureScheme enum, re-using the TLS signature schemes
* Clarify that the main job of a Credential is to provide a public key for verifying messages
    * How it provides the credential might be different per credential type
    * e.g., in BasicCredential, it's provided directly, X509Credential, have to parse the leaf certificate

draft-ietf-mls-protocol.md
```diff
@@ -889,25 +889,33 @@ The ciphersuites are defined in section {{mls-ciphersuites}}.
...
+* The identity of the holder of the private keys
 
 Credentials MAY also include information that allows a relying party
 to verify the identity / signing key binding.
 
+Additionally, Credentials SHOULD specify the signature scheme corresponding to
+each contained public key.
+
 ~~~~~
+
```
bifurcation, line 908:
```suggestion
```

draft-ietf-mls-protocol.md
```diff
@@ -889,25 +889,33 @@ The ciphersuites are defined in section {{mls-ciphersuites}}.
...
+// See IANA registry for registered values
+uint16 SignatureScheme;
+
 // See IANA registry for registered values
 uint16 CredentialType;
 
 struct {
     opaque identity<0..2^16-1>;
     opaque signature_key<0..2^16-1>;
+    SignatureScheme signature_scheme;
```
bifurcation, line 918:
```suggestion
    SignatureScheme signature_scheme;
    opaque signature_key<0..2^16-1>;
```

bifurcation (approved):
LGTM with a couple nits.

PR #404, opened 2020-09-16
Remove long term and explicitly prevent changing of identity

kkohbrok:
Rotating keys is important and in some cases, Signature Keys are not going to be very "long-term". Instead they're going to be rotated periodically.

I thus propose we simply remove the assumption that Signature Keys are of a "long-term" nature (or otherwise "long-lived"). 

The second change in this PR is that I added a few words indicating explicitly that the identity must not change when updating a KeyPackage.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -943,12 +943,12 @@ provide some public information about a user. KeyPackage
 structures provide information about a client that any existing
 member can use to add this client to the group asynchronously.
 
-A KeyPackage object specifies a ciphersuite that the client
-supports, as well as providing a public key that others can use
-for key agreement. The client's identity key can be updated
-throughout the lifetime of the group by sending a new KeyPackage
-with a new identity; the new identity MUST be validated by the
-authentication service.
+A KeyPackage object specifies a ciphersuite that the client supports, as well as
+providing a public key that others can use for key agreement. The client's
+identity key can be updated throughout the lifetime of the group by sending a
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -329,7 +328,7 @@ This protocol is designed to execute in the context of a Service Provider (SP)
 as described in {{?I-D.ietf-mls-architecture}}.  In particular, we assume
 the SP provides the following services:
 
-* A long-term identity key provider which allows clients to authenticate
+* A identity key provider which allows clients to authenticate
```
Bren2010, line 331:
```suggestion
* An identity key provider which allows clients to authenticate
```

draft-ietf-mls-protocol.md
```diff
@@ -944,12 +943,12 @@ provide some public information about a user. KeyPackage
...
-for key agreement. The client's identity key can be updated
-throughout the lifetime of the group by sending a new KeyPackage
-with a new identity; the new identity MUST be validated by the
-authentication service.
+A KeyPackage object specifies a ciphersuite that the client supports, as well as
+providing a public key that others can use for key agreement. The client's
+identity key can be updated throughout the lifetime of the group by sending a
+new KeyPackage with a new Credential. However, the identity MUST be the same in
+both Credentials and the new Credential MUST be validated by the authentication
+service.
```
Bren2010, line 951: This doesn't necessarily make sense because not all credentials have an `identity`

Bren2010 (approved):

bifurcation (approved):

PR #408, opened 2020-09-23
Clarify X509Credential structure

suhasHere:
This PR adds clarification around X509 cert structure, the chain and its encoding X509.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -936,7 +936,7 @@ X509 certificate. The chain is ordered such that the first entry (chain[0])
 is the end-entity certificate and each subsequent certificate in the chain 
 MUST be the issuer of the previous certificate. The algorithm for the  
 `public_key` in the end-entity certificate MUST match the relevant 
-ciphersuite for the enclosing KeyPackage.
+ciphersuite.
 
 For ciphersuites using Ed25519 or Ed448 signature schemes, the public key is in
 the format specified {{?RFC8032}}.  For ciphersuites using ECDSA with the NIST
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -927,11 +927,16 @@ struct {
 ~~~~~
 
 A BasicCredential is a raw, unauthenticated assertion of an identity/key
-binding.  The format of the key in the `public_key` field is defined by the
+binding. The format of the key in the `public_key` field is defined by the
 relevant ciphersuite: the group ciphersuite for a credential in a ratchet tree,
 the KeyPackage ciphersuite for a credential in a KeyPackage object.
 
-For X509Credential, each entry in the chain represents a single DER-encoded X509 certificate. The chain is ordered such that the first entry (chain[0]) is the end-entity certificate and each subsequent certificate in the chain MUST be the issuer of the previous certificate. The algorithm for the `public_key` in the end-entity certificate MUST match the relevant ciphersuite for the enclosing KeyPackage.
+For X509Credential, each entry in the chain represents a single DER-encoded 
+X509 certificate. The chain is ordered such that the first entry (chain[0]) 
+is the end-entity certificate and each subsequent certificate in the chain 
+MUST be the issuer of the previous certificate. The algorithm for the  
+`public_key` in the end-entity certificate MUST match the relevant 
+ciphersuite for the enclosing KeyPackage.
 
 For ciphersuites using Ed25519 or Ed448 signature schemes, the public key is in
 the format specified {{?RFC8032}}.  For ciphersuites using ECDSA with the NIST
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -914,7 +914,6 @@ struct {
     opaque cert_data<0..2^16-1>;
 } Certificate;
 
-
 struct {
     CredentialType credential_type;
     select (Credential.credential_type) {
@@ -932,12 +931,7 @@ binding.  The format of the key in the `public_key` field is defined by the
 relevant ciphersuite: the group ciphersuite for a credential in a ratchet tree,
 the KeyPackage ciphersuite for a credential in a KeyPackage object.
 
-For X509Credential, each entry in the chain represents a single  DER-encoded
- X509 certificate. The chain is ordered such that the first entry (chain[0]) 
- is the end-entity certificate and each subsequent entry in the chain MUST be 
- a parent of the previous entry. The algorithm for the `public_key` in the 
- end-entity certificate MUST match the ciphersuite for the enclosing 
- KeyPackage.
+For X509Credential, each entry in the chain represents a single DER-encoded X509 certificate. The chain is ordered such that the first entry (chain[0]) is the end-entity certificate and each subsequent certificate in the chain MUST be the issuer of the previous certificate. The algorithm for the `public_key` in the end-entity certificate MUST match the relevant ciphersuite for the enclosing KeyPackage.
 
 For ciphersuites using Ed25519 or Ed448 signature schemes, the public key is in
 the format specified {{?RFC8032}}.  For ciphersuites using ECDSA with the NIST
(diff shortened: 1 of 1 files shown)


bifurcation (approved):
Thanks, @suhasHere.  I agree that separating out the certificates is a good idea.  While it's different from how TLS does things, I'm ready to declare TLS wrong here, since this saves mixing TLS-syntax and X.509 parsing logic.

As for the chain ordering, I really thought we had this in here already, so I'm glad to have it back.

DemiMarie (approved):
Looks good.  In particular, requiring that certificates be ordered from leaf to root allows for arbitrarily long certificate chains to be verified in O(1) memory.

draft-ietf-mls-protocol.md
```diff
@@ -910,14 +910,19 @@ struct {
     opaque signature_key<0..2^16-1>;
 } BasicCredential;
 
+struct {
+    opaque cert_data<0..2^16-1>;
+} Certificate;
+
+
```
Bren2010, line 917: remove newline

draft-ietf-mls-protocol.md
```diff
@@ -927,6 +932,13 @@ binding.  The format of the key in the `public_key` field is defined by the
 relevant ciphersuite: the group ciphersuite for a credential in a ratchet tree,
 the KeyPackage ciphersuite for a credential in a KeyPackage object.
 
+For X509Credential, each entry in the chain represents a single  DER-encoded
```
Bren2010, line 935:
```suggestion
For X509Credential, each entry in the chain represents a single DER-encoded
```

draft-ietf-mls-protocol.md
```diff
@@ -927,6 +932,13 @@ binding.  The format of the key in the `public_key` field is defined by the
...
+ X509 certificate. The chain is ordered such that the first entry (chain[0])
```
Bren2010, line 936:
```suggestion
X509 certificate. The chain is ordered such that the first entry (chain[0])
```

draft-ietf-mls-protocol.md
```diff
@@ -927,6 +932,13 @@ binding.  The format of the key in the `public_key` field is defined by the
 relevant ciphersuite: the group ciphersuite for a credential in a ratchet tree,
 the KeyPackage ciphersuite for a credential in a KeyPackage object.
 
+For X509Credential, each entry in the chain represents a single  DER-encoded
+ X509 certificate. The chain is ordered such that the first entry (chain[0]) 
+ is the end-entity certificate and each subsequent entry in the chain MUST be
```
Bren2010, line 937:
```suggestion
is the end-entity certificate and each subsequent certificate in the chain MUST be
```

draft-ietf-mls-protocol.md
```diff
@@ -927,6 +932,13 @@ binding.  The format of the key in the `public_key` field is defined by the
 relevant ciphersuite: the group ciphersuite for a credential in a ratchet tree,
 the KeyPackage ciphersuite for a credential in a KeyPackage object.
 
+For X509Credential, each entry in the chain represents a single  DER-encoded
+ X509 certificate. The chain is ordered such that the first entry (chain[0]) 
+ is the end-entity certificate and each subsequent entry in the chain MUST be 
+ a parent of the previous entry. The algorithm for the `public_key` in the
```
Bren2010, line 938:
```suggestion
the issuer of the previous certificate. The algorithm for the `public_key` in the
```

draft-ietf-mls-protocol.md
```diff
@@ -927,6 +932,13 @@ binding.  The format of the key in the `public_key` field is defined by the
 relevant ciphersuite: the group ciphersuite for a credential in a ratchet tree,
 the KeyPackage ciphersuite for a credential in a KeyPackage object.
 
+For X509Credential, each entry in the chain represents a single  DER-encoded
+ X509 certificate. The chain is ordered such that the first entry (chain[0]) 
+ is the end-entity certificate and each subsequent entry in the chain MUST be 
+ a parent of the previous entry. The algorithm for the `public_key` in the 
+ end-entity certificate MUST match the ciphersuite for the enclosing
```
Bren2010, line 939:
```suggestion
end-entity certificate MUST match the relevant ciphersuite.
```

draft-ietf-mls-protocol.md
```diff
@@ -927,6 +932,13 @@ binding.  The format of the key in the `public_key` field is defined by the
 relevant ciphersuite: the group ciphersuite for a credential in a ratchet tree,
 the KeyPackage ciphersuite for a credential in a KeyPackage object.
 
+For X509Credential, each entry in the chain represents a single  DER-encoded
+ X509 certificate. The chain is ordered such that the first entry (chain[0]) 
+ is the end-entity certificate and each subsequent entry in the chain MUST be 
+ a parent of the previous entry. The algorithm for the `public_key` in the 
+ end-entity certificate MUST match the ciphersuite for the enclosing 
+ KeyPackage.
```
Bren2010, line 940:
```suggestion

```

Bren2010 (changes requested):

suhasHere:
@Bren2010  thanks for the suggestions. updated PR to reflect the changes . pls let me know ..

draft-ietf-mls-protocol.md
```diff
@@ -910,23 +910,34 @@ struct {
...
+binding. The format of the key in the `public_key` field is defined by the
 relevant ciphersuite: the group ciphersuite for a credential in a ratchet tree,
 the KeyPackage ciphersuite for a credential in a KeyPackage object.
 
+For X509Credential, each entry in the chain represents a single DER-encoded 
+X509 certificate. The chain is ordered such that the first entry (chain[0]) 
+is the end-entity certificate and each subsequent certificate in the chain 
+MUST be the issuer of the previous certificate. The algorithm for the  
+`public_key` in the end-entity certificate MUST match the relevant 
+ciphersuite for the enclosing KeyPackage.
```
Bren2010, line 939:
```suggestion
ciphersuite.
```

Bren2010 (changes requested):
Many lines still have trailing spaces, please fix

draft-ietf-mls-protocol.md
```diff
@@ -910,23 +910,34 @@ struct {
...
-binding.  The format of the key in the `public_key` field is defined by the
+binding. The format of the key in the `public_key` field is defined by the
 relevant ciphersuite: the group ciphersuite for a credential in a ratchet tree,
 the KeyPackage ciphersuite for a credential in a KeyPackage object.
 
+For X509Credential, each entry in the chain represents a single DER-encoded 
+X509 certificate. The chain is ordered such that the first entry (chain[0]) 
+is the end-entity certificate and each subsequent certificate in the chain 
+MUST be the issuer of the previous certificate. The algorithm for the  
+`public_key` in the end-entity certificate MUST match the relevant
```
franziskuskiefer, line 938: This wasn't clearly defined before so I appreciate making it clear.
However, this `MUST` makes the mandatory ciphersuite pretty hard to use with x509 certificates. Public CAs don't issue ed25519 certificates.

kkohbrok:
I think this makes it clear that there is a need to include additional data in Credentials. However, instead of "hardcoding" this for X509 Credential, I think it would be more useful to allow for extensions in Credentials in general. Such extensions could then be used to include an X509 chain or additional signature schemes/public key pairs as proposed in #402 .

draft-ietf-mls-protocol.md
DemiMarie: Most uses for MLS don’t involve public CAs.

bifurcation: It seems a bit early to talk about "most uses for MLS", since AFAIK there aren't any yet :) 

But I think the overall point still stands -- precedent in the Web PKI doesn't really matter here, because (again in most of the envisioned use cases), these certificates will be for user identifiers that aren't even allowed in Web PKI certificates.  I suppose it's possible to have a bunch of entities with domain names and Web PKI certificates doing MLS, but (a) that doesn't seem like a core use cases and (b) they can just use a ciphersuite that is lines up with those certs.

suhasHere:
@Bren2010 i have pushed another commit .. can you please check. thanks

Bren2010 (approved):

PR #415, opened 2020-10-08
make IANA reference for Signature Schemes more precise

kkohbrok:
Thanks @Bren2010 pointing out the imprecision. Is this what you were looking for?

Thanks @bifurcation for the pointer to the correct registry.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -904,7 +904,7 @@ Additionally, Credentials SHOULD specify the signature scheme corresponding to
 each contained public key.
 
 ~~~~~
-// See See RFC 8446 and the IANA TLS SignatureScheme registry
+// See RFC 8446 and the IANA TLS SignatureScheme registry
 uint16 SignatureScheme;
 
 // See IANA registry for registered values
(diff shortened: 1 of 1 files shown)


Bren2010:
This doesn't make sense because we don't actually use this registry. The signature scheme is defined by the ciphersuite, and the ciphersuite registry doesn't map to this unrelated registry

kkohbrok:
The `SignatureScheme` is not authoritative in the group context and will always correspond with the one in the group (as is required by the generic description of the functionalities that Credentials have to provide). The SignatureScheme is there for when the Credential is handled in a context where there is no surrounding group. For example, when it's handled by the AS.

There is a mapping from ciphersuit to SignatureScheme in Table 5, the entries of which correspond to the entries in the referenced registry. Is that what you meant?

draft-ietf-mls-protocol.md
```diff
@@ -904,7 +904,7 @@ Additionally, Credentials SHOULD specify the signature scheme corresponding to
 each contained public key.
 
 ~~~~~
-// See IANA registry for registered values
+// See See RFC 8446 and the IANA TLS SignatureScheme registry
```
Bren2010, line 907:
```suggestion
// See RFC 8446 and the IANA TLS SignatureScheme registry
```

Bren2010 (changes requested):

bifurcation (approved):

Bren2010 (approved):

PR #416, opened 2020-10-10
Inlclude the signature in the confirmation tag

dajost:
As discussed in the mailing list, not having the signature contribute to the new epoch's key schedule leads to a one-off effect where parties stay in sync (with respect to keys) for one more epoch, before inevitably go out of sync. This pull request addresses this issue by including the signature in the confirmed transcript hash rather than the interim transcript hash. See the mailing list for a security consideration of this change.

Credit: This pull request was primarily authored by Marta Mularczk; I'm just posting it here for administrative reasons.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1220,10 +1220,11 @@ interim_transcript_hash_[n+1] =
 
 Thus the `confirmed_transcript_hash` field in a GroupContext object represents a
 transcript over the whole history of MLSPlaintext Commit messages, up to the
-confirmation tag field in the current MLSPlaintext message.  The confirmation tag
-is then included in the transcript for the next epoch.  The interim transcript hash
-is passed to new members in the GroupInfo struct, and enables existing members to
-incorporate a Commit message into the transcript without having to store the whole MLSPlaintextCommitAuthData structure.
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1723,10 +1723,10 @@ struct {
 
         case commit:
           Commit commit;
-          opaque confirmation_tag<0..255>;
     }
 
     opaque signature<0..2^16-1>;
+    optional<MAC> confirmation_tag;
```
dajost, line 1729: An alternative would be to use:
        case commit:
          MAC confirmation_tag;
}
sparing the need for the definition below that the confirmation tag must be present iff it is a commit.

bifurcation: Not a bad idea; it would save an octet and give you "syntactic" enforcement.  I'm still inclined to go with the `optional<>` approach for two reasons:

1. I'm not aware of TLS syntax implementatiosn that implement `select()` in any sort of automated fashion, so there would be manual logic here in any case.  It's just slightly easier to omit with `optional<>`.

2. If we do `select` for `confirmation_tag`, we should probably also do it with `membership_tag` (selecting based on `sender_type`), which gets to be ugly.

bifurcation (approved):

PR #417, opened 2020-10-12
use `KDF.Nh` instead of `Hash.length`

kkohbrok:


bifurcation (approved):
For the record: I asked @kkohbrok whether we could just update DeriveTreeSecret so that it always produces `KDF.Nh` octets.  But he pointed out that DeriveTreeSecret is also used to produce keys and nonces.  Might be something to clean up later.

PR #420, opened 2020-10-14
make confirmation_tag a MAC just like membership_tag

kkohbrok:


kkohbrok:
Ah, I missed that this is apparently already underway in #416.

PR #421, opened 2020-10-14
Review PSK PR.

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -1639,30 +1640,13 @@ the group.
 It is RECOMMENDED for the application generating exported values
 to refresh those values after a Commit is processed.
 
-## Recovery Keys
```
kkohbrok, line 1642: Why remove this section? If I understand correctly, it fits well into the subsections explaining the use of the various keys derived from the key schedule.

draft-ietf-mls-protocol.md
```diff
@@ -2013,18 +1995,6 @@ of the new group, provided the GroupId is long enough to prevent collisions.
 However, GroupId is currently 'hidden' in the GroupInfo, which is only available
 to the receiver _after_ the PSK would have to be injected. -->
 
-### Re-Initialization {#re-initialization}
```
kkohbrok, line 2016: Why remove this, but not the subsection below on Sub-group branching? I think both section serve the purpose of giving an overview over the respective process and should be left in.

draft-ietf-mls-protocol.md
```diff
@@ -2066,7 +2036,7 @@ enum {
     add(1),
     update(2),
     remove(3),
-    presharedkey(4)
+    psk(4)
```
kkohbrok, line 2039: @bifurcation wanted to have this spelled out. I'm happy with the abbreviation.

draft-ietf-mls-protocol.md
```diff
@@ -2260,14 +2236,6 @@ Update for the leaf if there are no Removes. If there are multiple Add proposals
 for the same client, the committer again chooses one to include and considers
 the rest invalid.
 
-If there are multiple ReInit proposals, the committer MUST choose exactly one to
```
kkohbrok, line 2263: Why remove this? I think we should at least specify that only one re-init proposal can be included in a commit.

draft-ietf-mls-protocol.md
bifurcation: I can live with this.

draft-ietf-mls-protocol.md
```diff
@@ -1379,15 +1379,14 @@ update), does that constitute a weaker security level -->
 security level of the ciphersuite used in the group. -->
 
 Each PSK in MLS has a type that designates how it was provisioned.
-External PSKs are provided by the application logic. Other resumption PSKs
-are derived from the MLS key schedule, and used in cases where it is
+External PSKs are provided by the application, while resumption PSKs
```
bifurcation, line 1382:
```suggestion
External PSKs are provided by the application, while recovery and re-init PSKs
```

draft-ietf-mls-protocol.md
bifurcation: Or `pre_shared_key(4)`

draft-ietf-mls-protocol.md
```diff
@@ -2156,19 +2126,20 @@ A member of the group applies a Remove message by taking the following steps:
...
 struct {
     PreSharedKeyID psk;
 } PreSharedKey;
 ~~~~~
 
-When processing a Commit message including one or more PreSharedKey proposals, a
-group member MUST derive the `psk_secret` for the inclusion in the Key Schedule
-as described in {{pre-shared-keys}}, where the order of the PSKs corresponds to
-the order of the PreSharedKey proposals in the commit.
+The `psktype` of the pre-shared key MUST be `external` and the `psk_nonce` MUST
```
bifurcation, line 2138: "MUST be `external`" is too strong here; this proposal is also how "recovery" PSKs are injected, right?  I think all you want to say is "MUST NOT be `reinit`"

draft-ietf-mls-protocol.md
```diff
@@ -1639,30 +1640,13 @@ the group.
...
-
 ## State Authentication Keys
 
 The main MLS key schedule provides a per-epoch `authentication_secret`.
 If one of the parties is being actively impersonated by an attacker, their
 `authentication_secret` will differ from that of the other group members.
 Thus, members of a group MAY use their `authentication_secrets` within
-an out-of-band authentication protocol to detect attacks or ensure that they
+an out-of-band authentication protocol to ensure that they
 share the same view of the group.
```
raphaelrobert, line 1650: I think detecting attacks is a valid goal of the authentication keys, we shouldn't remove it.

draft-ietf-mls-protocol.md
```diff
@@ -2186,13 +2157,18 @@ struct {
...
 of the new group corresponds to the ones specified in the proposal. The Welcome
-message MUST specify a `PreSharedKey` with `psktype = reinit` and with
+message MUST specify exactly one pre-shared key with `psktype = reinit`, and with
 `psk_group_id` and `psk_epoch` equal to the `group_id` and `epoch` of the
 existing group after the Commit containing the `reinit` Proposal was processed.
+The Welcome message may specify the inclusion of other pre-shared keys with a
+`psktype` different from `reinit`.
 
+If a ReInit proposal is included in a Commit, it MUST be the only proposal
+referenced by the Commit. The `version` field in the ReInit proposal MUST be no
```
raphaelrobert, line 2170: This sounds like existing proposals should be dropped, which should not be the case. I think it better to mandate an extra Commit that is issued when there are no more pending proposals.

raphaelrobert (changes requested):

draft-ietf-mls-protocol.md
```diff
@@ -1379,15 +1379,14 @@ update), does that constitute a weaker security level -->
...
+External PSKs are provided by the application, while resumption PSKs
+are derived from the MLS key schedule and used in cases where it is
 necessary to authenticate a member's participation in a prior group state.
 In particular, in addition to external PSK types, a PSK derived from within MLS
 may be used in the following cases:
 
-  - Re-Initialization: If, during the lifetime of the group, a change in the
-    fixed group parameters becomes necessary, e.g. if the ciphersuite used
-    by the group is deprecated or if the protocol version should be upgraded,
+  - Re-Initialization: If during the lifetime of a group, the group members
```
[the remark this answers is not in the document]
Bren2010: I don't see a provision for that in the External Proposals section. I think there's more to think through in that case, which we can do in another PR

draft-ietf-mls-protocol.md
Bren2010: You did not define how to do recovery. Define how to do recovery in another PR and you can add it back

draft-ietf-mls-protocol.md
Bren2010: It's redundant. The second clause of the sentence just says "detect attacks" again in a different way.

draft-ietf-mls-protocol.md
Bren2010: This is all (correctly) discussed in the section on ReInit proposals. Sub-group branching isn't discussed anywhere else

draft-ietf-mls-protocol.md
Bren2010: There are no recovery PSKs, `external` is currently the only valid PSK type to put here

draft-ietf-mls-protocol.md
Bren2010: Yes, the ReInit must be dropped or the rest of the proposals must be dropped. Applications can do what you say automatically: Commit outstanding proposals -> send new ReInit -> Commit ReInit. There's no reason to mandate it

draft-ietf-mls-protocol.md
Bren2010: This discussion was moved to the ReInit proposal section

> If a ReInit proposal is included in a Commit, it MUST be the only proposal
referenced by the Commit.

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
kkohbrok: In my mind, that was one of the main features of having a ReInit proposal. But you are right, we should have mentioned it in the External Proposals section. I'll make a PR.

draft-ietf-mls-protocol.md
kkohbrok: This Section is not about keys for recovery specifically, but about the `recovery_secret`, which is the basis for internal PSK derivation. I agree that the name is not ideal, though.

draft-ietf-mls-protocol.md
kkohbrok: You are right, that the of the currently specified PSK types, only `external` can be used in a psk proposal. However, implementors could define proprietary PSK types (e.g. for recovery) that they would want to use with a psk proposal as well. Not sure how we should write the spec so that option remains open.

draft-ietf-mls-protocol.md
kkohbrok: Why does it have to be the only proposal in the commit? I mean, updates are going to be useless, but Adds and Removes can still be processed.

bifurcation: Discussed at virtual interim 2020-10-20:
* ReInit MUST be only proposal
* If non-ReInit and ReInit received in same epoch, SHOULD prefer non-ReInit
* Options for recovering from contention: committer re-originates or original proposer retries

raphaelrobert (approved):

kkohbrok (approved):
Looks good! Thanks for cleaning up some of the bits we missed in the original PSK PR!

PR #423, opened 2020-10-16
Remove encrypted sender data from ciphertext AAD.

Bren2010:


bifurcation:
@psyoptix - We had talked about this circular dependency offline.  Does this resolution make sense to you?

bifurcation:
@Bren2010 It seems like at this point, we can provide the same AAD to the content and sender data encryption.  Should we consolidate into something like this?

```
struct {
    opaque group_id<0..255>;
    uint64 epoch;
    ContentType content_type;
} MLSCiphertextAAD;
```

Bren2010:
Well, we don't know how large `authenticated_data` will be. If it's large, doing multiple passes over it when they aren't necessary would be undesirable

bifurcation:
Discussed on virtual interim 2020-10-20:
* Sender data AAD is almost not necessary, since the key/nonce are derived from the ciphertext, which includes the whole header as AAD
* But sampling introduces some ambiguity here, and this argument could be difficult for composability
* So the original path of having minimal AAD in sender data encryption is the right compromise

PR #406, opened 2020-09-21
External Commits

raphaelrobert:

PR #433, opened 2020-10-26
Sign the PublicGroupState

bifurcation:
The unsigned PublicGroupState created by #406 leads to an attack on the new joiner.  The sender of a GroupPublicState can associate any HPKE it wants with the group, and thus learn the `init_secret` without being a member of the group.  (The UpdatePath and the resulting `commit_secret` partly defend against this attack, though they are subverted if the attacker has compromised any single key used in the Update Path.)

To guard against this attack, this PR adds a signature over the HPKE public key in the PublicGroupState.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2613,7 +2613,7 @@ following information for the group's current epoch:
 
 This information is aggregated in a `PublicGroupState` object as follows:
 
-```
+~~~
 struct {
     CipherSuite cipher_suite;
     opaque group_id<0..255>;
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2622,16 +2622,36 @@ struct {
...
+leaf node at position `signer_index`.  The signature covers the following
+structure, comprised of fields from the PublicGroupState object:
+
+~~~~~
+struct {
+    opaque group_id<0..255>;
+    uint64 epoch;
+    HPKEPublicKey external_pub;
+} PublicGroupStateTBS;
+~~~~~
```
Bren2010, line 2644: Is there a good reason not to sign everything? Would be simpler and more robust to do so

br-hale: I second this question. Following discussion elsewhere on the mailing list, deniability may be pushed to the application level. If deniability was a factor for not signing everything, then it may not apply anymore.

br-hale: That makes sense. Even if we are still aiming for some form of deniability here, I think that we can sign a little more. For example, the cipher_suite, extensions, should be fine. The opaque_signature should be the main issue with respect to deniability. @raphaelrobert : do you have a strong opinion about not signing the transcript hash?

kkohbrok: The difference between signing the transcript hash and signing just the public key is that you need knowledge of the key schedule (and how it was derived) to tie the public key to the rest of the group state, while (if you don't encrypt handshake messages), the transcript hash can be computed by any party observing the traffic. At least as far as I understand the transcript hash computation. So in terms of deniability it certainly gets easier for the adversary.

raphaelrobert: Sorry for the late reply. I think the PR really just covers the bare minimum of what needs to be signed. My initial proposal was to cover the whole PublicGroupState struct by the signature, except for the tree_hash field.
Just like @br-hale I think including the transcript hash would make this more robust. That being said, @kkohbrok has a very valid point here: the transcript hash can be computed by a passive network observer without any knowledge of group secrets (this is where we are missing the derived epoch IDs we discarded). My intuition is that if we include the transcript hash we might as well include everything.
As @br-hale pointed out, if we go for the beefier version of the parent hash anyway, seeking any deniability guarantees at this level makes little sense.

bifurcation:
In light of the feedback on the mailing list from @claucece re: deniability, I have upgraded this PR to sign the whole PublicGroupState.

Bren2010 (approved):

raphaelrobert:
Outstanding issue: One detail that is not fully covered by #436 is the strategy new joiners should apply to determine who the latest committer is and make sure that both the `PublicGroupState` and the last Commit are signed by the same member.

bifurcation:
The last node to send an UpdatePath is the last one to set the root.  So you can follow parent_hash to find which leaf it is.  But like I said on the IETF meeting call, I don't think we should actually require that PublicGroupState be signed by the last updater.  There are possible operational/sync issues if the PublicGroupState isn't fate-shared with the Commit, but that's a practical issue, not a security issue.

PR #435, opened 2020-11-03
Fix parent hash verification

dajost:
This is a spin-off from the work on tree-signing by Joël, Marta, and me. Note that this obsoletes #434 as I couldn't figure out how to do dependent PRs in Github.

Consider the following ratchet tree and assume that (1) the left subtree is fully occupied and (2) E last committed in the right subtree, i.e., the key at node [A..F] has been sampled by E. 
```
       [root]   
           \
            \
             \
            [A..F]
            /    \
           /      \
          /        \
         /          \
     [A..D]        [EF]
      /  \          /\
     /    \        /  \
    /      \      /    \
  [AB]    [CD]   [E]   [F]
   /\      /\
  /  \    /  \
[A] [B] [C] [D]
```
Now say party Z from the left subtree committs a proposal that adds party G. According to my understanding the tree will look as follows, where I use the syntax [A..F,G] to denote that parties A to F know the secret key and G is tracked as an unmerged leaf. 
```
       [root]   
           \
            \
             \
           [A..F,G]
            /    \
           /      \
          /        \
         /          \
     [A..D]        [X]
      /  \          /\
     /    \        /  \
    /      \      /    \
  [AB]    [CD]   [EF]  [G]
   /\      /\     /\
  /  \    /  \   /  \
[A] [B] [C] [D] [E] [F]
```

Now observe the following two issues:
* According to the draft, G now has to verify that either of [A..F,G]'s children store a correct parent hash. However, its parent-hash is no longer stored in its direct child [X], an additional blank node inserted for the add, but still in [EF] instead. This needs to be accounted for in the verification.
* Orignally, the ``parent_hash`` includes ``unmerged_leaves``. Hence, for [A..F,G] the value stored in [EF] (and recursively signed by E) will no longer match the recomputed value. Since partent hashes are only updated along the committer's path which at this point has no unmerged leaves, however, there is no point in including unmerged leaves.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1114,8 +1114,15 @@ opaque parent_hash<0..255>;
 
 This extension MUST be present in all Updates that are sent as part of a Commit
 message. If the extension is present, clients MUST verify that `parent_hash`
-matches the hash of the leaf's parent node when represented as a ParentNode
+matches the hash of the leaf's parent node when represented as a ParentHashInput
 struct.
+
+~~~~~
+struct {
+    HPKEPublicKey public_key;
+    opaque parent_hash<0..255>;
+} ParentHashInput;
+~~~~~
 
 <!-- OPEN ISSUE: This scheme, in which the tree hash covers the parent hash, is
 designed to allow for more deniable deployments, since a signature by a member
@@ -1150,9 +1157,8 @@ struct {
 } optional<T>;
 
 struct {
-    HPKEPublicKey public_key;
+    ParentHashInput parent_hash_input;
     uint32 unmerged_leaves<0..2^32-1>;
-    opaque parent_hash<0..255>;
 } ParentNode;
 ~~~~~
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2752,7 +2752,9 @@ welcome_key = KDF.Expand(welcome_secret, "key", key_length)
     children are non-empty and have the hash of this node set as their
     `parent_hash` value (if the child is another parent) or has a `parent_hash`
     extension in the KeyPackage containing the same value (if the child is a
-    leaf).
+    leaf). If either of the node's children is empty, and in particular does not
+	have a parent hash, then its respective children's `parent_hash` values have
+	to be considered instead.
 
   * For each non-empty leaf node, verify the signature on the KeyPackage.
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2759,8 +2759,8 @@ welcome_key = KDF.Expand(welcome_secret, "key", key_length)
     `parent_hash` value (if the child is another parent) or has a `parent_hash`
     extension in the KeyPackage containing the same value (if the child is a
     leaf). If either of the node's children is empty, and in particular does not
-	have a parent hash, then its respective children's `parent_hash` values have
-	to be considered instead.
+    have a parent hash, then its respective children's `parent_hash` values have
+    to be considered instead.
 
   * For each non-empty leaf node, verify the signature on the KeyPackage.
 


Bren2010 (changes requested):
I'm not sure this gets where you want.

parent_hash = H(subtree headed by my parent, excluding parent_hash related stuff)

dajost:
I am not sure I can follow your concern. Just to clarify, this is so far just about fixing the current scheme, preserving the deniability properties. The bigger change of reversing the order of the tree hashes in the parent hashes will be a different PR.

Bren2010:
Ah, if this is still about the old scheme, I would say we just wait for the change to the new one.

draft-ietf-mls-protocol.md
```diff
@@ -2752,7 +2758,9 @@ welcome_key = KDF.Expand(welcome_secret, "key", key_length)
     children are non-empty and have the hash of this node set as their
     `parent_hash` value (if the child is another parent) or has a `parent_hash`
     extension in the KeyPackage containing the same value (if the child is a
-    leaf).
+    leaf). If either of the node's children is empty, and in particular does not
+    have a parent hash, then its respective children's `parent_hash` values have
+    to be considered instead.
```
bifurcation, line 2763: This seems incorrect to me.  The  value of `parent_hash` is set from the direct parent, so it's useless to look at the blank node's children.  You don't have enough information to verify that `parent_hash`; you're missing what was in the blank node.  You also can't reset the `parent_hash` to "skip over" the blank node when you set it to blank, because that would invalidate `parent_hash` values below it in the tree and ultimately, the signature on the leaf.

draft-ietf-mls-protocol.md
```diff
@@ -1150,9 +1157,8 @@ struct {
 } optional<T>;
 
 struct {
-    HPKEPublicKey public_key;
+    ParentHashInput parent_hash_input;
```
bifurcation, line 1160: I prefer not to group things like this, and just assemble the hash input as necessary.  Especially since #436 is going to change `ParentHashInput` to diverge further from `ParentNode`.  Implementations are of course free to store things this way.

bifurcation (commented):
Given the issues noted, I am inclined to close this with no action.

bifurcation:
I think I see where the misunderstanding is here.  The tree you present is only constructed transiently, for use in generating the `UpdatePath` for the commit.  The tree that is actually sent to G has the `UpdatePath` merged:

```
       [root]   
           \
            \
             \
             [X3]               X3.ph = 0
            /    \
           /      \
          /        \
         /          \
      [X2]         [__]         X2.ph = H(X3.pk, X3.ph, [EF.pk, G.pk])
      /  \          /\
     /    \        /  \
    /      \      /    \
  [AB]    [X1]   [EF]  [G]      X1.ph = H(X2.pk, X2.ph, [AB.pk])
   /\      /\     /\
  /  \    /  \   /  \
[A] [B] [X] [D] [E] [F]         X.ph  = H(X1.pk, X1.ph, [D.pk])
```

Here `X` is the sender of the update, and the ParentHashes according to the scheme in #436 are shown.  In this tree, the updated nodes have valid parent hash values, and the new blank node doesn't need one.

MartaMularczyk:
> I think I see where the misunderstanding is here. The tree you present is only constructed transiently, for use in generating the `UpdatePath` for the commit. The tree that is actually sent to G has the `UpdatePath` merged:
> 
> ```
>        [root]   
>            \
>             \
>              \
>              [X3]               X3.ph = 0
>             /    \
>            /      \
>           /        \
>          /          \
>       [X2]         [__]         X2.ph = H(X3.pk, X3.ph, [EF.pk, G.pk])
>       /  \          /\
>      /    \        /  \
>     /      \      /    \
>   [AB]    [X1]   [EF]  [G]      X1.ph = H(X2.pk, X2.ph, [AB.pk])
>    /\      /\     /\
>   /  \    /  \   /  \
> [A] [B] [X] [D] [E] [F]         X.ph  = H(X1.pk, X1.ph, [D.pk])
> ```
> 
> Here `X` is the sender of the update, and the ParentHashes according to the scheme in #436 are shown. In this tree, the updated nodes have valid parent hash values, and the new blank node doesn't need one.

I think the misunderstanding may be in who's the committer: this picture shows only the right half of the tree and the committer X is in the left part (left child of [root]). So after X commits, [A..F,G].ph doesn't change and it's still computed using [EF]'s values.

bifurcation:
Ah, thanks for the clarification, @MartaMularczyk.  So the observation here is that committing Add proposals can result in blank nodes being inserted beneath a parent node, as the tree grows.  That means that you can't just go down one level, you have to recurse to the left until you find a non-blank node (equivalently, take the first node in the resolution of the right child).

It seems like the right parent hash validation is then to do the following check for each non-blank `parent_node`:

```
H = Hash(parent_node)
L = left_child(parent_node)
R = left_child^k(right_child(parent_node) // k >= 0 smallest value such that R != blank
return H == L.parent_hash || H == R.parent_hash;
```

The theory being that extending the tree always adds nodes to the right, so that when these situations arise, you just need to recurse to the left.

bifurcation:
Given that this PR will need to be rebased on #436 anyway, I'm closing this one and proposing that we just update #436 to do the right thing.  I have suggested text on that PR.

PR #436, opened 2020-11-12
Strong tree signing

psyoptix:
This pull request makes editorial changes to the definition of Tree Hash  to hopefully make it easier to understand.

It also redefines Parent Hash to include the HPKE public keys to which a nodes secret key was sent. This binds a key package introduced to the ratchet tree in a commit not just to the HPKE keys on the commits direct path but also (indirectly) to the other members that were sent each of the new (secret) keys. 

In particular, this prevents the attacks from the mailinglist that allowed an attacker to create artificial ratchet trees where the tree invariant was violated (which in turn lead to removes failing). But parent_hash only includes the HPKE pubkeys in the resolution so at least some amount of deniability might be preserved.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1122,7 +1122,7 @@ struct {
 } ParentHashInput;
 ~~~~~
 
-The Parent Hash of V with Co-Path Child S is the hash of a `ParentHashInput` object
+The Parent Hash of P with Co-Path Child S is the hash of a `ParentHashInput` object
 populated as follows. The field `public_key` contains the HPKE public key of P. If P
 is the root then `parent_hash` is set to the all-zeroes vector of size `Nh`.
 Otherwise `parent_hash` is the Parent Hash of P's parent with P's sibling as the
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1114,8 +1114,15 @@ opaque parent_hash<0..255>;
 
 This extension MUST be present in all Updates that are sent as part of a Commit
 message. If the extension is present, clients MUST verify that `parent_hash`
-matches the hash of the leaf's parent node when represented as a ParentNode
+matches the hash of the leaf's parent node when represented as a ParentHashInput
 struct.
+
+~~~~~
+struct {
+    HPKEPublicKey public_key;
+    opaque parent_hash<0..255>;
+} ParentHashInput;
+~~~~~
 
 <!-- OPEN ISSUE: This scheme, in which the tree hash covers the parent hash, is
 designed to allow for more deniable deployments, since a signature by a member
@@ -1150,9 +1157,8 @@ struct {
 } optional<T>;
 
 struct {
-    HPKEPublicKey public_key;
+    ParentHashInput parent_hash_input;
     uint32 unmerged_leaves<0..2^32-1>;
-    opaque parent_hash<0..255>;
 } ParentNode;
 ~~~~~
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1182,7 +1182,7 @@ Note that the `node_index` field contains the index of the leaf among the nodes
 in the tree, not its index among the leaves; `node_index = 2 * leaf_index`.
 
 Now the tree hash of any non-leaf node is recursively defined to be the hash of
-its `ParentNodeTreeHashInput`. This includes an optional `ParentNodeTreeHashData`
+its `ParentNodeTreeHashInput`. This includes an optional `ParentNode`
 object depending on if the node is blank or not.
 
 ~~~~~
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1122,16 +1122,16 @@ This extension MUST be present in the `leaf_key_package` Key Package field of an
 `UpdatePath` object. When processing a Commit message clients MUST recompute the 
 expected value of `parent_hash` for the committer's new leaf and verify that it 
 matches the `parent_hash` value in the `leaf_key_package`. Moreover, when joining
-a group new members MUST verify that if a leaf contains a `parent_hash` value than
+a group new members MUST verify that if a leaf contains a `parent_hash` value, then
 it matches the value obtained by recomputing `parent_hash` at the leaf.
 
 The `parent_hash` at the root is `0`. To compute the parent hash at a non-root
-node v with parent p and sibling u the `ParentHashInput` struct is used. It
-consists of three fields. The first contains the HPKE public key of p. The second
-contains the `parent_hash` at p. The third contains the list of HPKE public
-keys to which the HPKE secret key of v's parent was sent. That is, it consists of
+node V with parent P and sibling S, the `ParentHashInput` struct is used. It
+consists of three fields: The first contains the HPKE public key of P. The second
+contains the `parent_hash` at P. The third contains the list of HPKE public
+keys to which the HPKE secret key of P was sent to. That is, it consists of
 the array of `HPKEPublicKey` values of the nodes in the resolution of u but with
-the keys of p's `unmerged_leaves` omitted. 
+the`unmerged_leaves` of P omitted. 
 
 For example, in the ratchet tree depicted in {{resolution-example}} the
 `ParentHashInput` struct for node 6 would contain an empty array as the sibling of
@@ -1168,7 +1168,7 @@ struct {
 } optional<T>;
 ~~~~~
 
-The tree hash a leaf node is the hash of leaf's `LeafNodeHashInput` object which
+The tree hash of a leaf node is the hash of leaf's `LeafNodeHashInput` object which
 might include a Key Package depending on whether or not it is blank.
 
 ~~~~~
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1103,35 +1103,36 @@ opaque key_id<0..2^16-1>;
 
 ## Parent Hash
 
-The `parent_hash` extension serves to bind a KeyPackage to all the nodes
-above it in the group's ratchet tree. This enforces the tree invariant, meaning
-that malicious members can't lie about the state of the ratchet tree when they
-send Welcome messages to new members.
+The `parent_hash` extension serves to contain the adverse effect by a malicous 
+member lying about the state of the ratchet tree when they send Welcome messages 
+to new members. It binds a KeyPackage to all subtrees it is contained in of the 
+group's ratchet tree, enables joining members to verify that each subtree is 
+unmodified from when it has last been refreshed.
 
 ~~~~~
 opaque parent_hash<0..255>;
 ~~~~~
 
-This extension MUST be present in all Updates that are sent as part of a Commit
-message. If the extension is present, clients MUST verify that `parent_hash`
-matches the hash of the leaf's parent node when represented as a ParentHashInput
-struct.
+This extension MUST be present in all KeyPackages that are sent as part of an 
+UpdatePath in a Commit message. If the extension is present, clients MUST verify
+that `parent_hash` matches the expected value.
+
+To compute the parent hash of a node v, the `ParentHashInput` struct is used. It
+consists of v's parent, represented as a `ParentHashNode` struct, and the 
+resolution of v's sibling node, excluding all `unmerged_leaves` from v's parent
+node.
 
 ~~~~~
 struct {
-    HPKEPublicKey public_key;
-    opaque parent_hash<0..255>;
+	HPKEPublicKey public_key;
+	opaque parent_hash<0..255>;
+} ParentHashNode;
+
+struct {
+	ParentHashNode parent_hash_node;
+	HPKEPublicKey merged_sibling_resolution<0..2^32-1>;
 } ParentHashInput;
 ~~~~~
-
-<!-- OPEN ISSUE: This scheme, in which the tree hash covers the parent hash, is
-designed to allow for more deniable deployments, since a signature by a member
-covers only its direct path. The other possible scheme, in which the parent hash
-covers the tree hash, provides better group agreement properties, since a
-member's signature covers the entire membership of the trees it is in. Further
-discussion is needed to determine whether the benefits to deniability justify
-the harm to group agreement properties, or whether there are alternative
-approaches to deniability that could be compatible with the other approach. -->
 
 ## Tree Hashes
 
@@ -1157,7 +1158,7 @@ struct {
 } optional<T>;
 
 struct {
-    ParentHashInput parent_hash_input;
+    ParentHashNode parent_hash_node;
     uint32 unmerged_leaves<0..2^32-1>;
 } ParentNode;
 ~~~~~
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1122,9 +1122,10 @@ struct {
 ~~~~~
 
 The Parent Hash of P with Co-Path Child S is the hash of a `ParentHashInput` object
-populated as follows. Field `public_key` contains the HPKE public key of P. If P is
-the root then `parent_hash` is set to the all-zeroes vector of size `Nh`. Otherwise
-`parent_hash` is the Parent Hash of P's parent with P's sibling as the co-path child.
+populated as follows. The field `public_key` contains the HPKE public key of P. If P
+is the root then `parent_hash` is set to the all-zeroes vector of size `Nh`.
+Otherwise `parent_hash` is the Parent Hash of P's parent with P's sibling as the
+co-path child.
 
 Finally, `original_child_resolution` is the array of `HPKEPublicKey` values of the
 nodes in the resolution of S but with the `unmerged_leaves` of P omitted. For
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1,4 +1,4 @@
---- 
+---
 title: The Messaging Layer Security (MLS) Protocol
 abbrev: MLS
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
-The `parent_hash` extension serves to bind a KeyPackage to all the nodes
-above it in the group's ratchet tree. This enforces the tree invariant, meaning
-that malicious members can't lie about the state of the ratchet tree when they
-send Welcome messages to new members.
+The `parent_hash` extension serves to bind a KeyPackage inserted as part of an
+`UpdatePath` during a commit to the rest of the HPKE keys in that update path.
+Moreover, for each of those keys `parent_hash` also binds the set of parties
+that were told the corresponding secret key. This helps new members joining a group
+verify the tree invariant. That is, it prevents malicious members from
+constructing an artificial ratchet tree with a node v whose HPKE secret key is
```
Bren2010, line 1111:
```suggestion
constructing an artificial ratchet tree with a node whose HPKE secret key is
```

draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
+known to the malicious member yet where the member is not assigned a leaf in the
+subtree rooted at v. (Indeed, such a ratchet tree would violate the tree
```
Bren2010, line 1113:
```suggestion
corresponding subtree. (Indeed, such a ratchet tree would violate the tree
```

draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
 
-This extension MUST be present in all Updates that are sent as part of a Commit
-message. If the extension is present, clients MUST verify that `parent_hash`
-matches the hash of the leaf's parent node when represented as a ParentNode
-struct.
+This extension MUST be present in the `leaf_key_package` Key Package field of an
+`UpdatePath` object. When processing a Commit message clients MUST recompute the 
+expected value of `parent_hash` for the committer's new leaf and verify that it 
+matches the `parent_hash` value in the `leaf_key_package`. Moreover, when joining
+a group new members MUST verify that if a leaf contains a `parent_hash` value than
```
Bren2010, line 1125:
```suggestion
a group new members MUST verify that if a leaf contains a `parent_hash` value, then
```

draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
 
-<!-- OPEN ISSUE: This scheme, in which the tree hash covers the parent hash, is
-designed to allow for more deniable deployments, since a signature by a member
-covers only its direct path. The other possible scheme, in which the parent hash
-covers the tree hash, provides better group agreement properties, since a
-member's signature covers the entire membership of the trees it is in. Further
-discussion is needed to determine whether the benefits to deniability justify
-the harm to group agreement properties, or whether there are alternative
-approaches to deniability that could be compatible with the other approach. -->
+The `parent_hash` at the root is `0`. To compute the parent hash at a non-root
```
Bren2010, line 1128:
```suggestion
The `parent_hash` at the root is a zero-length bit string. To compute the parent hash at a non-root
```

draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
+node v with parent p and sibling u the `ParentHashInput` struct is used. It
```
Bren2010, line 1129:
```suggestion
node V with parent P and sibling S, the `ParentHashInput` struct is used. It
```

draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
+consists of three fields. The first contains the HPKE public key of p. The second
```
Bren2010, line 1130:
```suggestion
consists of three fields: The first contains the HPKE public key of P. The second
```

draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
+contains the `parent_hash` at p. The third contains the list of HPKE public
```
Bren2010, line 1131:
```suggestion
contains the `parent_hash` at P. The third contains the list of HPKE public
```

draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
+keys to which the HPKE secret key of v's parent was sent. That is, it consists of
```
Bren2010, line 1132:
```suggestion
keys to which the HPKE secret key of P was sent to. That is, it consists of
```

draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
+the array of `HPKEPublicKey` values of the nodes in the resolution of u but with
```
Bren2010, line 1133:
```suggestion
the array of `HPKEPublicKey` values of the nodes in the resolution of U, with
```

draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
+the keys of p's `unmerged_leaves` omitted.
```
Bren2010, line 1134:
```suggestion
the`unmerged_leaves` of P omitted. 
```

draft-ietf-mls-protocol.md
```diff
@@ -1148,39 +1166,42 @@ struct {
         case 1: T value;
     }
 } optional<T>;
-
-struct {
-    HPKEPublicKey public_key;
-    uint32 unmerged_leaves<0..2^32-1>;
-    opaque parent_hash<0..255>;
-} ParentNode;
```
Bren2010, line 1156: This struct got deleted but is still referenced in several places

Bren2010 (changes requested):
Please also remove trailing spaces from many lines

draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
+leaf for 6's parent node 5. Meanwhile, the `ParentHashInput` of node 1 has an
+array with the HPKE public key's of nodes 5 and 6.
+
+~~~~~
+struct {
+	HPKEPublicKey parents_public_key;
+	opaque parents_parent_hash<0..255>;
+	HPKEPublicKey original_sibling_resolution<0..2^32-1>;
+} ParentHashInput;
+~~~~~
```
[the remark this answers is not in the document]
bifurcation: It seems a little odd that we have a pointer "across" the tree here, from a node to its sibling.  Would it work to move the resolution to the parent?  So the ParentHashInput would say, "this public key under this parent, with the private key encrypted to this resolution".  Together with the above suggestions, you would end up with something like:

```
struct {
  HPKEPublicKey public_key;     // from parent
  opaque parent_hash<0..255>;   // from parent
  opaque resolution_hash<0..8>; // from non-updated child
} ParentHashInput;
```

Then your validity check for a non-blank parent node is that for one of its children:
```
one_child.parent_hash = H(ParentHashInput{public_key, parent_hash, other_child.resolution_hash})
```

draft-ietf-mls-protocol.md
```diff
@@ -1148,39 +1166,42 @@ struct {
...
-struct {
-    HPKEPublicKey public_key;
-    uint32 unmerged_leaves<0..2^32-1>;
-    opaque parent_hash<0..255>;
-} ParentNode;
 ~~~~~
 
-When computing the hash of a parent node, the `ParentNodeHashInput`
-structure is used:
+The tree hash a leaf node is the hash of leaf's `LeafNodeHashInput` object which
```
bifurcation, line 1171:
```suggestion
The tree hash of a leaf node is the hash of leaf's `LeafNodeHashInput` object which
```

draft-ietf-mls-protocol.md
Bren2010: I've always been trying to use zero-length octet strings instead of "0" because "0" is ambiguous to me. For `Nh`-long all-zero string, I don't think we've used that anywhere else. Probably just because it's wordier

psyoptix: My (weak) preference is to use an Nh long all-zero string. But only because that makes it the right data type for the output of a hash function (in contrast to the empty string).

draft-ietf-mls-protocol.md
psyoptix: I'm not a fan of "corresponding" subtree because its a bit imprecise since we dont define what the correspondence is. I find "subtree rooted at v" clearer. Alternatively we could define "corresponding subtree" somewhere but my preference is to not introduce new (e.g. graph theory) terminology whenever standard terms can be used to the same end.

draft-ietf-mls-protocol.md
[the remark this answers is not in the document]
psyoptix: Re: Validation. Yes, i think the same validation algo should work as before. And I agree text describing validation is necessary. Just not sure where to put that. Could put it in the Parent Hash section but maybe its more appropriate when describing how to process commit packets and how to join a group? (Group members should check that the parent_hash in the commitors key package checks out by recomputing it and comparing to whats in the key package. New members should check that the parent hash of each internal node is stored at one of its two children.)

draft-ietf-mls-protocol.md
```diff
@@ -1103,42 +1103,60 @@ opaque key_id<0..2^16-1>;
...
-discussion is needed to determine whether the benefits to deniability justify
-the harm to group agreement properties, or whether there are alternative
-approaches to deniability that could be compatible with the other approach. -->
+The `parent_hash` at the root is `0`. To compute the parent hash at a non-root
+node V with parent P and sibling S, the `ParentHashInput` struct is used. It
+consists of three fields: The first contains the HPKE public key of P. The second
+contains the `parent_hash` at P. The third contains the list of HPKE public
+keys to which the HPKE secret key of P was sent to. That is, it consists of
+the array of `HPKEPublicKey` values of the nodes in the resolution of S but with
+the`unmerged_leaves` of P omitted.
```
bifurcation, line 1134:
```suggestion
the `unmerged_leaves` of P omitted.
```

draft-ietf-mls-protocol.md
bifurcation: Yeah, I think we still need this one.  It goes in the syntax for how you send the tree to a new member.

draft-ietf-mls-protocol.md
```diff
@@ -1148,39 +1166,42 @@ struct {
...
+Now the tree hash of any non-leaf node is recursively defined to be the hash of
+its `ParentNodeTreeHashInput`. This includes an optional `ParentNodeTreeHashData`
+object depending on if the node is blank or not.
 
 ~~~~~
+struct {
+	HPKEPublicKey public_key;
+	opaque parent_hash<0..255>;
+    uint32 unmerged_leaves<0..2^32-1>;
+} ParentNodeTreeHashData;
```
bifurcation, line 1193:
```suggestion
struct {
    HPKEPublicKey public_key;
    opaque parent_hash<0..255>;
    uint32 unmerged_leaves<0..2^32-1>;
} ParentNode;
```

draft-ietf-mls-protocol.md
```diff
@@ -1148,39 +1166,42 @@ struct {
...
 ~~~~~
 
-The `left_hash` and `right_hash` fields hold the hashes of the node's
-left and right children, respectively. When computing the hash of
-a leaf node, the hash of a `LeafNodeHashInput` object is used:
+Note that the `node_index` field contains the index of the leaf among the nodes
+in the tree, not its index among the leaves; `node_index = 2 * leaf_index`.
+
+Now the tree hash of any non-leaf node is recursively defined to be the hash of
+its `ParentNodeTreeHashInput`. This includes an optional `ParentNodeTreeHashData`
```
bifurcation, line 1185:
```suggestion
its `ParentNodeTreeHashInput`. This includes an optional `ParentNode`
```

I prefer "ParentNode", since we use this for sending the tree as well as hashing.

bifurcation (commented):
@psyoptix I think you mean to change "ParentNode" to "ParentNodeTreeHashData", but didn't get all the way there.  I prefer it as "ParentNode" anyway :)  If you could fix that and the other comments noted, I think this will be ready to merge.

draft-ietf-mls-protocol.md
```diff
@@ -1101,44 +1102,78 @@ an explicit, application-defined identifier to a KeyPackage.
...
+
+### Verifying Parent Hashes
+
+To this end, when processing a Commit message clients MUST recompute the
+expected value of `parent_hash` for the committer's new leaf and verify that it
+matches the `parent_hash` value in the supplied `leaf_key_package`. Moreover, when
+joining a group, new members MUST authenticate each non-blank parent node P. A parent
+node P is authenticated by checking that it has children V and S such that V's
+`ParentNode` struct has its `parent_hash` field set to the Parent Hash of P with
+Co-Path Child S.
```
bifurcation, line 1165: This section should address the problem noted in #435.  Namely, `V` should either be `left(P)` or `left^k(right(P)`, for the smallest possible `k >= 0`. Suggested text:

```suggestion
To this end, when processing a Commit message clients MUST recompute the
expected value of `parent_hash` for the committer's new leaf and verify that it
matches the `parent_hash` value in the supplied `leaf_key_package`. Moreover, when
joining a group, new members MUST authenticate each non-blank parent node P. A parent
node P is authenticated by performing the following check: 

* Let L and R be the left and right children of P, respectively
* If L.parent_hash is equal to the Parent Hash of P with Co-Path Child R, the check passes 
* If R is blank, replace R with its left child until R is either non-blank or a leaf node
* If R is a leaf node, the check fails
* If R.parent_hash is equal to the Parent Hash of P with Co-Path Child L, the check passes
* Otherwise, the check fails

The left-child recursion under the right child of P is necessary because the expansion of 
the tree to the right due to Add proposals can cause blank nodes to be interposed 
between a parent node and its right child. 
```

cc @MartaMularczyk @dajost

bifurcation (commented):
One more small fix here, then I think this is ready to merge.

draft-ietf-mls-protocol.md
psyoptix: Nice! Looks good to me.

bifurcation (approved):

draft-ietf-mls-protocol.md
```diff
@@ -1101,52 +1102,87 @@ an explicit, application-defined identifier to a KeyPackage.
...
-discussion is needed to determine whether the benefits to deniability justify
-the harm to group agreement properties, or whether there are alternative
-approaches to deniability that could be compatible with the other approach. -->
+    HPKEPublicKey original_child_resolution<0..2^32-1>;
+} ParentHashInput;
+~~~~~
+
+The Parent Hash of P with Co-Path Child S is the hash of a `ParentHashInput` object
+populated as follows. The field `public_key` contains the HPKE public key of P. If P
+is the root then `parent_hash` is set to the all-zeroes vector of size `Nh`.
```
Bren2010, line 1127:
```suggestion
is the root, then `parent_hash` is set to a zero-length octet string.
```

draft-ietf-mls-protocol.md
```diff
@@ -1101,52 +1102,87 @@ an explicit, application-defined identifier to a KeyPackage.
...
+Otherwise `parent_hash` is the Parent Hash of P's parent with P's sibling as the
+co-path child.
+
+Finally, `original_child_resolution` is the array of `HPKEPublicKey` values of the
+nodes in the resolution of S but with the `unmerged_leaves` of P omitted. For
+example, in the ratchet tree depicted in {{resolution-example}} the
+`ParentHashInput` of node 5 with co-path child 4 would contain an empty
+`original_child_resolution` since 4's resolution includes only itself but 4 is also
+an unmerged leaf of 5. Meanwhile, the `ParentHashInput` of node 5 with co-path child
+6 has an array with one element in it; namely the HPKE public key of 6.
```
Bren2010, line 1137:
```suggestion
6 has an array with one element in it: the HPKE public key of 6.
```

draft-ietf-mls-protocol.md
```diff
@@ -1101,52 +1102,87 @@ an explicit, application-defined identifier to a KeyPackage.
...
+
+### Using Parent Hashes
+
+The Parent Hash of P appears in three types of structs. If V is itself a parent node
+then the P's Parent Hash is stored in the `parent_hash` fields of both V's
```
Bren2010, line 1142:
```suggestion
then P's Parent Hash is stored in the `parent_hash` fields of both V's
```

draft-ietf-mls-protocol.md
```diff
@@ -2815,7 +2860,9 @@ welcome_key = KDF.Expand(welcome_secret, "key", key_length)
     children are non-empty and have the hash of this node set as their
     `parent_hash` value (if the child is another parent) or has a `parent_hash`
     extension in the KeyPackage containing the same value (if the child is a
-    leaf).
+    leaf). If either of the node's children is empty, and in particular does not
```
Bren2010, line 2863:
```suggestion
    leaf). If either of the node's children are empty, and in particular does not
```

Bren2010 (commented):
Looks good, please fix conflicts

draft-ietf-mls-protocol.md
bifurcation: No, the subject which which the verb must agree is "either" (singular), not "children" (plural).

Bren2010 (approved):

PR #439, opened 2020-11-17
Identities SHOULD be unique per group

kkohbrok:
This PR changes the draft such that it mandates identities SHOULD be unique per group.

In particular, a committer SHOULD only include one Add per identity if multiple are present and SHOULD NOT include Adds for identities that are already members of the group.

I also made the requirement more precise, that there shouldn't be multiple Adds for the same "client", which I change "client" to "KeyPackage".

The rationale here being that this change allows us to address group members by their identity rather than their leaf index, which would allow us to finally hide the fact that we're using a tree underneath from users of MLS.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2185,10 +2185,11 @@ leaf in the tree, for the second Add, the next empty leaf to the right, etc.
   public key from the KeyPackage in the Add, as well as the credential under
   which the KeyPackage was signed
 
-The identity contained in the credential of the `key_package` should be unique
-in the context of the group.  While the proposer can not necessarily know if that
-is the case, the committer of the proposal SHOULD consider the proposal invalid
-if a credential with the same identity is already present in the group.
+The identity, as well as the signature key contained in the credential of the
+`key_package` MUST be unique in the context of the group. While the proposer can
+not necessarily know if that is the case, the committer of the proposal MUST
+consider the proposal invalid if a credential with the same identity is already
+present in the group.
 
 ### Update
 
@@ -2376,10 +2377,11 @@ Update for the leaf if there are no Removes. If there are multiple Add proposals
 referencing the same `key_package`, the committer again chooses one to include
 and considers the rest invalid.
 
-If there are multiple Add proposals of KeyPackages with the same identity, the
-committer SHOULD choose one and consider the rest invalid. Similarly, if there
-are Add proposals of KeyPackages with an identity that is already present as a
-member of the group, the committer SHOULD consider them invalid.
+If there are multiple Add proposals of KeyPackages with the same identity or
+signature key, the committer MUST choose one and consider the rest invalid.
+Similarly, if there are Add proposals of KeyPackages with an identity or
+signature key that is already present as a member of the group, the committer
+MUST consider them invalid.
 
 The Commit MUST NOT combine proposals sent within different epochs. In the event
 that a valid proposal is omitted from the next Commit, the sender of the
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2542,9 +2542,10 @@ A member of the group applies a Commit message by taking the following steps:
   any Update or Remove proposals, or if it's empty. Otherwise, the `path` value
   MAY be omitted.
 
-* If the `path` value is populated: Process the `path` value using the ratchet
-  tree the provisional GroupContext, to update the ratchet tree and generate the
-  `commit_secret`:
+* If the `path` value is populated: Verify that the identity and signature key
+  of the credential in the new leaf are unique for this group. Then process the
+  `path` value using the ratchet tree the provisional GroupContext, to update
+  the ratchet tree and generate the `commit_secret`:
 
   * Apply the UpdatePath to the tree, as described in
     {{synchronizing-views-of-the-tree}}, and store `key_package` at the
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2185,6 +2185,11 @@ leaf in the tree, for the second Add, the next empty leaf to the right, etc.
   public key from the KeyPackage in the Add, as well as the credential under
   which the KeyPackage was signed
 
+The identity contained in the credential of the `key_package` should be unique
+in the context of the group.  While the proposer can not necessarily know if that
+is the case, the committer of the proposal SHOULD consider the proposal invalid
+if a credential with the same identity is already present in the group.
+
 ### Update
 
 An Update proposal is a similar mechanism to Add with the distinction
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2185,6 +2185,11 @@ leaf in the tree, for the second Add, the next empty leaf to the right, etc.
   public key from the KeyPackage in the Add, as well as the credential under
   which the KeyPackage was signed
 
+The identity contained in the credential of the `key_package` should be unique
```
raphaelrobert, line 2188: should -> SHOULD

draft-ietf-mls-protocol.md
```diff
@@ -2368,8 +2373,14 @@ If there are multiple proposals that apply to the same leaf, the committer
 chooses one and includes only that one in the Commit, considering the rest
 invalid. The committer MUST prefer any Remove received, or the most recent
 Update for the leaf if there are no Removes. If there are multiple Add proposals
-for the same client, the committer again chooses one to include and considers
-the rest invalid.
+referencing the same `key_package`, the committer again chooses one to include
+and considers the rest invalid.
+
+If there are multiple Add proposals with KeyPackages belonging to the same
+identity, the committer SHOULD choose one and consider the rest invalid.
```
raphaelrobert, line 2380: "If there are multiple Add proposals of KeyPackages with the same identity"

draft-ietf-mls-protocol.md
```diff
@@ -2368,8 +2373,14 @@ If there are multiple proposals that apply to the same leaf, the committer
...
+Similarly, if there are Add proposals including KeyPackages belonging to an
+identity that is already present as a member of the group, the committer SHOULD
```
raphaelrobert, line 2382: "if there are Add proposals of KeyPackages with an identity that is already present"

raphaelrobert (changes requested):

draft-ietf-mls-protocol.md
kkohbrok: I thought those capitalized SHOULDs, MUSTs, etc are only used for actual instructions. I use it later on in cases where I describe what the committer SHOULD do. I'm happy to change it here, as well, though.

bifurcation:
This PR seems like all down-side to me.  On the one hand, it's not clear what benefit there is to maintaining distinct identities in the credential.  On the other hand, it seems like there are cases where it is desirable to have multiple appearances of the same identity, for example, multiple devices owned by the same user.  If the application wants to implement "remove a user" functionality, then they just need to iterate over all instances of that user in the tree.

So I would be inclined to close this PR with no action.

kkohbrok:
The benefit would be that group members are not solely identified by their leaf index. I understand if that's not motivation enough.

However, the spec currently says

> If there are multiple Add proposals for the same client, the committer again chooses one to include and considers the rest invalid.

"Client" is a bit tricky if we don't define it via the identity. It's defined as an entity that is defined by the key material it holds.  But how to detect if a client is added multiple times? We could try checking if it's the same key package, but if that party uses the same identity for all their devices/clients then it might well be the same, because it's down to the key package of last resort. Or are we assuming that clients will always use different key packages? Or is that something where we leave the check to the application layer? That point should be either a bit more specific or removed entirely.

>  If the application needs to individual device identifiers, it can choose to encode these in the credential or use an extension to provide a separate device identifier.

Why wouldn't that then be part of the `identity`? The only place we really do anything with the `identity` is currently the fact that it can't change with an update.

bifurcation:
> The benefit would be that group members are not solely identified by their leaf index. I understand if that's not motivation enough.

Yeah, I don't see why I would care about this.  The leaf index is clear and unambiguous for intra-MLS usage, and applications can use the rest of the KeyPackage to decide which leaves to remove.

Agree that we should fix that spec text.  IMO the right answer is to make clear that it's up to the application to decide when two KeyPackages represent the same client.

> Why wouldn't that then be part of the identity?

There are circumstances where `credential.identity` might be constrained so that it can't hold all the information you want.  For example, if you were using X509Credentials from a CA that certified email addresses, then all instances of a user with a given email address would have the same `credential.identity` (the email address).  You would need something else to distinguish different instances/devices of the same user, probably a KeyPackage extension.

kkohbrok:
I see your point. I'll close this, then.

kkohbrok:
In light of the discussion on the mailing list regarding the improvement of client authentication, I'm reopening this PR. I've updated this PR to include both identities and signature keys.

beurdouche (approved):

raphaelrobert:
This looks almost good! I think it's missing the rule that Commits MUST be rejected if they contain identities that are not unique.

seanturner:
Discussed at 20210526 Interim. Look to merge.

draft-ietf-mls-protocol.md
```diff
@@ -2368,8 +2374,14 @@ If there are multiple proposals that apply to the same leaf, the committer
...
-for the same client, the committer again chooses one to include and considers
-the rest invalid.
+referencing the same `key_package`, the committer again chooses one to include
+and considers the rest invalid.
+
+If there are multiple Add proposals of KeyPackages with the same identity or
+signature key, the committer MUST choose one and consider the rest invalid.
+Similarly, if there are Add proposals of KeyPackages with an identity or
+signature key that is already present as a member of the group, the committer
+MUST consider them invalid.
```
bifurcation, line 2384: I still object to this requirement, on the same grounds as before.  Namely, multi-device cases will result in multiple appearances of the same identity with high probability.

I thought we had agreed to make the prohibition on duplication of signature keys.  If that's the case, then let's make that change and merge.

bifurcation (changes requested):

draft-ietf-mls-protocol.md
kkohbrok: Is there a particular reason why you wouldn't want to use different identities for the devices? They will have to be distinguished at some point. Even if you use X509 certificates, you could still use the fingerprint as identity.

This change would allow us to address group members by their identity as opposed to their leaf index, which makes for a more intuitive API and allows us to hide the fact that we're using a tree underneath.

bifurcation: Well, if you define "identity" in the right way, sure :)  

Part of the challenge here is that we don't really have a well-defined notion of "identity" on which to hang requirements such as these.  And I would argue that we should not have such a notion at this level -- it's up to the application to define how the credentials in MLS are used to authenticate application-level identities.  "Identity" is also a bad API concept, because you're going to have different cases; think "remove this specific device" vs. "remove all participants with this phone number".

Your X.509 example seems like it points back to my suggestion of focusing this on signature keys.  Aside from serial numbers, public keys are the major reason that two certificates would be different.

kkohbrok: I'm less concerned about concrete API schemes, and more about the fact that we can't abstract the tree away.

I see that signature keys would allow us to hide the tree under an abstraction. However, signature keys of a given client, even if they are unique, (potentially) change over the time of that client's group membership, so if we used that to address individual group members, we would have to have an additional mapping that keeps track of changing signature keys, which is not ideal.

(I realize now that that means that my fingerprint example wouldn't work, either.)

You are right in that we should probably talk a bit about what "identity" really means in the context of MLS. As you said, it's the job of the application layer to verify credentials, so what is verified by the AS could be something more than the `identity` field that the spec currently requires. Looking at it from that angle: does the `identity` field currently serve any purpose in the spec other than that it's verified as part of the credential and that it doesn't change through updates?

If that is not the case, we could require the identity field to be something unique, while leaving other fields in the credential (that MLS doesn't have to know about) as the "user-facing" identity, which (also) gets verified by the AS as part of the credential and which the user can base their for authentication decision on. Thus, for MLS, the `identity` field, would be mostly administrative.

bifurcation:
Replaced by #476

PR #442, opened 2020-11-21
Add an "AppAck" proposal

bifurcation:
We had longstanding issue #160 to allow the participants in an MLS group to detect when the DS drops messages.  That issue was closed because it had been open for a long time without a concrete proposed solution.  More recently, however, @raphaelrobert and @beurdouche suggested using a Proposal for this, and it made a lot of sense to me when I looked into the details.  So this PR should be considere joint work between, @raphaelrobert, @beurdouche, and me.

This PR adds an `AppAck` Proposal that any member of the group can send to advertise the last generation it sent, and acknowledge the receipt of application messages from other members.  As a Proposal, it is included in the transcript via Commits, providing a pretty strong defence against message suppression -- the DS basically has to stop the group progressing at all if it wants to drop applications without the group being aware of it.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2134,6 +2134,7 @@ enum {
     psk(4),
     reinit(5),
     external_init(6),
+    app_ack(7),
     (255)
 } ProposalType;
 
@@ -2146,6 +2147,7 @@ struct {
         case psk:           PreSharedKey;
         case reinit:        ReInit;
         case external_init: ExternalInit;
+        case app_ack:       AppAck;
     };
 } Proposal;
 ~~~~~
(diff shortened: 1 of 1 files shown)


bifurcation:
BTW, there are a *bunch* of ways to go about [expressing what a member did and did not receive](https://en.wikipedia.org/wiki/Acknowledgement_(data_networks)).  Most of these schemes are designed for 1:1 connections like TCP or QUIC, but adapt in a straightforward way to groups.  Which one is most efficient depends on what you assume about losses.  Some examples:

1. List of received messages (the most naïve approach; efficient only if losses dominate)
2. List of received message ranges (implemented here, more efficient for infrequent losses)
3. Highest received generation, plus list of generations not received (more efficient for even more infrequent losses)

FWIW, scheme (2) is effectively [what QUIC does](https://tools.ietf.org/html/draft-ietf-quic-transport-32#section-19.3).  TCP does something like (3), but because it is synchronous, doesn't include generations not received.  The ["selective acknowledgement"](https://tools.ietf.org/html/rfc2018#section-3) option for TCP is basically the same as option (2), reporting ranges of received packets.

raphaelrobert (approved):

kkohbrok:
Great idea! Two comments:

To make sure that the DS doesn't drop any Commit that has an AppAck extension (to cover up their dropping of Application messages), either one has to encrypt Commit messages or AppAcks need to be mandatory for Commits by policy. Otherwise the DS can just drop any Commit that has an AppAck in it (without impeding the sending of messages in the group).

A cheaper, although functionally less powerful version would be to simply send a hash of the SecretTree. The other group members would then know if they received the same messages as the sender (but not more than that).

bifurcation:
@kkohbrok good point about addressing suppression of AppAck itself.  I added a paragraph. 

Sending a hash of the SecretTree seems both less powerful and way more complicated, since we would have to define a hashing scheme.  So I'm gonna pass :)

PR #446, opened 2020-11-28
Move joiner_secret to the end of the first KDF cycle in the Key Schedule

ericcornelissen:
This is a small suggestion to move the definition of the `joiner_secret` to the
end of the first Extract-Expand (KDF) cycle of the [key schedule]. This makes
more sense from a cryptographic perspective as we would not be using anything
from the middle of a KDF cycle.

A related suggestion is to remove the `member_secret` from the document. I
removed it from the key schedule, removed all the mentions to forget it as soon
as possible and updated the description of deriving the `welcome_key` and
`welcome_nonce` from the [welcoming new members] section.

[key schedule]: https://github.com/mlswg/mls-protocol/blob/master/draft-ietf-mls-protocol.md#key-schedule
[welcoming new members]: https://github.com/mlswg/mls-protocol/blob/master/draft-ietf-mls-protocol.md#welcoming-new-members

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1356,7 +1356,7 @@ proceeds as shown in the following diagram:
     commit_secret -> KDF.Extract
                          |
                          V
-                   DeriveSecret(., "member")
+                   DeriveSecret(., "joiner")
                          |
                          V
                     joiner_secret
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1362,7 +1362,7 @@ proceeds as shown in the following diagram:
                     joiner_secret
                          |
                          V
-psk_secret (or 0) -> KDF.Extract = member_secret
+psk_secret (or 0) -> KDF.Extract
                          |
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1353,13 +1353,16 @@ proceeds as shown in the following diagram:
...
                          V
-    commit_secret -> KDF.Extract = joiner_secret
+    commit_secret -> KDF.Extract
                          |
                          V
                    DeriveSecret(., "member")
                          |
                          V
-psk_secret (or 0) -> KDF.Extract = member_secret
+                    joiner_secret
```
bifurcation, line 1362: Nit: Either this should be `member` or the label above should be `joiner`

ericcornelissen: Changed it to `joiner` as I think calling this the _"joiner_secret"_ makes more sense, but would be happy to do the change the other way as well.

bifurcation:
Thanks for the quick fix, @ericcornelissen !

PR #449, opened 2020-12-16
Pin HPKE to -07.

chris-wood:
cc @bifurcation


beurdouche (approved):

chris-wood:
@beurdouche is the build expected to fail?

beurdouche:
No, but the problem seem unrelated to your change, so we are all good.

bifurcation:
Looks like it was a transient failure.  Re-running resulted in success.

PR #453, opened 2020-12-31
Use the GroupContext to derive the joiner_secret

ericcornelissen:
This is a proposal to change how the `joiner_secret` is derived. Before, it was derived simply by derivation with a label. This Pull Request updates this step to include the GroupContext, similar to how the `epoch_secret` is derived. Including the GroupContext will give stronger uniqueness guarantees on the `joiner_secret` cryptographically speaking. I think the GroupContext in the derivation of the `epoch_secret` should be kept as it proves that joiners have learned the GroupContext through the Welcome message as well.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1413,7 +1413,7 @@ proceeds as shown in the following diagram:
     commit_secret -> KDF.Extract
                          |
                          V
-                   DeriveSecret(., "joiner")
+                   ExpandWithLabel(., "joiner", GroupContext_[n], KDF.Nh)
                          |
                          V
                     joiner_secret


raphaelrobert (approved):

Bren2010 (approved):

bifurcation:
I have a couple of problems with this PR:

> I think the GroupContext in the derivation of the epoch_secret should be kept as it proves that joiners have learned the GroupContext through the Welcome message as well.

I don't see how this follows -- joiners are told the `joiner_secret` directly in the GroupSecrets message, they don't compute it themselves.  (It's also doesn't prove anything useful about the existing members, since their knowledge of the `init_secret` proves that they knew the GroupContext for the last epoch.)  If you wanted this property, you would have to instead send the output of `KDF.Extract` in GroupSecrets.

But there's a bigger problem here: This would create a circular dependency.  The joiner doesn't know the group context until they decrypt the Welcome message, which they do with the `welcome_secret`.  So you can't inject anything about the GroupContext in between the thing you hand to the joiner and the `welcome_secret`.

In other words, the injection of GroupContext is already done as early as it can be.  So we should close this PR

ericcornelissen:
> I don't see how this follows -- joiners are told the joiner_secret directly in the GroupSecrets message, they don't compute it themselves.

Apologies for the confusion but that is not quite what I intended to say. Yes, joiners are told the `joiner_secret`, so what I meant to say is this: as joiners provided with the `joiner_secret` directly they don't need to use the `GroupContext` in the first `ExpandWithLabel` call, hence the `GroupContext` should be **kept** in the second `ExpandWithLabel` call.


> It's also doesn't prove anything useful about the existing members, since their knowledge of the init_secret proves that they knew the GroupContext for the last epoch.

I (intuitively) agree with that, but as per the Pull Request description: I added the `GroupContext` in the first `ExpandWithLabel` call for stronger uniqueness guarantees on the `joiner_secret` only.



> But there's a bigger problem here: This would create a circular dependency. The joiner doesn't know the group context until they decrypt the Welcome message, which they do with the welcome_secret. So you can't inject anything about the GroupContext in between the thing you hand to the joiner and the welcome_secret.

I'm not sure if I follow, I'm not proposing we change anything about how joiners are added to the group, they would still receive the `joiner_secret` as they do now. Therefore I'm not injecting anything about the `GroupContext` in between the thing we hand to the joiner (i.e. the `joiner_secret`) and the `welcome_secret`.

Again, my apologies if the Pull Request description wasn't clear. The change I'm proposing is primarily for uniqueness guarantees on the `joiner_secret`. I included the last sentence of the Pull Request description only as a motivation for keeping the `GroupContext` in the second `ExpandWithLabel` call.

bifurcation:
Thanks for clarifying @ericcornelissen.  Good that we agree w.r.t. joiners.  If we're not aiming at that, then this seems pretty innocuous to me.  Would it be correct to say that this ensures that parties that don't agree on the GroupContext diverge as soon as possible?

I'm not sure I see a ton of value in that, since the `joiner_secret` is only used ephemerally by existing members -- for them, the meaningful transition is `init_secret + commit_secret + psk_secret -> epoch_secret`, and that already includes the GroupContext.

That said, as long as we're not trying to address joiners, this seems innocuous, so if others are keen, I can go along with it.

ericcornelissen:
> Would it be correct to say that this ensures that parties that don't agree on the GroupContext diverge as soon as possible?

That is correct, but not the only way of looking at it as per the below.


> I'm not sure I see a ton of value in that, since the `joiner_secret` is only used ephemerally by existing members -- for them, the meaningful transition is `init_secret + commit_secret + psk_secret -> epoch_secret`, and that already includes the GroupContext.

Practically speaking I would (intuitively) agree. That said, if a collision on the `joiner_secret` where to happen (and no PSK is used) everything up to and including the `welcome_secret` would collide and since the `GroupSecrets` are encrypted with keys derived (deterministically) from the `welcome_secret`, a collision _would_ be a problem.

Namely, in a cryptographic analysis this would mean that the adversary may know a `welcome_secret` that it is not supposed to know. In a game-based setting, the adversary can first derive a (malicious) `welcome_secret` that they know, and than a colliding one they're not supposed to know. If I recall correctly, this "problematic scenario" may result from the fact that the `commit_secret` is not chained between epochs (a feature like https://mailarchive.ietf.org/arch/msg/mls/ZR84smU5xeLrziNTk5W1P1Z1nQI/ would likely give the same level of uniqueness to the `joiner_secret`, but this Pull Request is a far simpler solution for this particular problem).

bifurcation:
Thanks for the explanation, @ericcornelissen.  Makes sense to me.

seanturner:
Discussed at 20210526 Interim. No downsides noted at this point. @kkohbrok agreed to review and provide any additional input.

bifurcation (approved):
One note I'll leave here for posterity: I briefly thought that because we are now injecting the GroupContext into the `joiner_secret`, we could remove the context from the `epoch_secret` derivation.  However, including the context there confirms that the joiner computed the same GroupContext as the Welcome sender, so it's important to keep it.

PR #454, opened 2021-01-21
more editorial changes

uhoreg:
- add the missing `authenticated_data` field to `MLSPlaintextCommitContent`
- the sections for the `Commit` sender and `Commit` receiver seem to have gotten out of sync.  I tried to sync them up (but let me know if I messed something up)
- fix a typo

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1289,6 +1289,7 @@ struct {
     opaque group_id<0..255>;
     uint64 epoch;
     Sender sender;
+    opaque authenticated_data<0..2^32-1>;
     ContentType content_type = commit;
     Commit commit;
     opaque signature<0..2^16-1>;
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2532,21 +2532,23 @@ uses:
    regard to any removed members and PCS for the committer and any updated
    members.
 
-When creating or processing a Commit, three different GroupContexts are used:
+When creating or processing a Commit, three different ratchet trees and
+their associated GroupContexts are used:
 
-1. The "old GroupContext" is the GroupContext for the epoch before the commit.
-   This is used when signing the MLSPlainText so that existing group members
-   can verify the signature before processing the commit.
-2. The "provisional GroupContext" is the GroupContext constructed after
-   applying the proposals that are referenced by the Commit.  This GroupContext
-   uses the epoch number for the new epoch, and the old confirmed transcript
-   hash.  This is used when creating the UpdatePath, if the UpdatePath is
-   needed.
-3. The "new GroupContext" is the GroupContext constructed after applying the
-   proposals and the UpdatePath (if any).  This GroupContext uses the epoch
-   number for the new epoch, and the new confirmed transcript hash.  This is
-   used when deriving the new epoch secrets, and is the only GroupContext that
-   newly-added members will have.
+1. "Old" refers to the ratchet tree and GroupContext for the epoch before the
+   commit.  The old GroupContext is used when signing the MLSPlainText so that
+   existing group members can verify the signature before processing the
+   commit.
+2. "Provisional" refers to the ratchet tree and GroupContext constructed after
+   applying the proposals that are referenced by the Commit.  The provisional
+   GroupContext uses the epoch number for the new epoch, and the old confirmed
+   transcript hash.  This is used when creating the UpdatePath, if the
+   UpdatePath is needed.
+3. "New" refers to the ratchet tree and GroupContext constructed after applying
+   the proposals and the UpdatePath (if any).  The new GroupContext uses the
+   epoch number for the new epoch, and the new confirmed transcript hash.  This
+   is used when deriving the new epoch secrets, and is the only GroupContext
+   that newly-added members will have.
 
 A member of the group creates a Commit message and the corresponding Welcome
 message at the same time, by taking the following steps:
@@ -2555,7 +2557,7 @@ message at the same time, by taking the following steps:
   field populated from Proposals received during the current epoch, and an empty
   `path` field.
 
-* Generate a provisional GroupContext object by applying the proposals
+* Generate the provisional ratchet tree and GroupContext by applying the proposals
   referenced in the initial Commit object, as described in {{proposals}}. Update
   proposals are applied first, followed by Remove proposals, and then finally
   Add proposals. Add proposals are applied in the order listed in the
@@ -2573,37 +2575,41 @@ message at the same time, by taking the following steps:
   based on the proposals that are in the commit (see above), then it MUST be
   populated.  Otherwise, the sender MAY omit the `path` field at its discretion.
 
-* If populating the `path` field: Create an UpdatePath using the new tree and
-  the provisional GroupContext. Any new member (from an add proposal) MUST be
+* If populating the `path` field: Create an UpdatePath using the provisional
+  ratchet tree and GroupContext. Any new member (from an add proposal) MUST be
   exluded from the resolution during the computation of the UpdatePath.  The
   `leaf_key_package` for this UpdatePath must have a `parent_hash` extension.
 
    * Assign this UpdatePath to the `path` field in the Commit.
 
    * Apply the UpdatePath to the tree, as described in
-     {{synchronizing-views-of-the-tree}}. Define `commit_secret` as the value
-     `path_secret[n+1]` derived from the `path_secret[n]` value assigned to
-     the root node.
+     {{synchronizing-views-of-the-tree}}, creating the new ratchet tree. Define
+     `commit_secret` as the value `path_secret[n+1]` derived from the
+     `path_secret[n]` value assigned to the root node.
 
 * If not populating the `path` field: Set the `path` field in the Commit to the
   null optional.  Define `commit_secret` as the all-zero vector of the same
-  length as a `path_secret` value would be.
+  length as a `path_secret` value would be.  In this case, the new ratchet tree
+  is the same as the provisional ratchet tree.
 
 * If one or more PreSharedKey proposals are part of the commit, derive the `psk_secret`
   as specified in {{pre-shared-keys}}, where the order of PSKs in the derivation
   corresponds to the order of PreSharedKey proposals in the `proposals` vector.
   Otherwise, set `psk_secret` to a zero-length octet string.
 
-* Use the `init_secret` from the previous epoch, the `commit_secret` and the
-  `psk_secret` as defined in the previous steps, and the new GroupContext to
-  compute the new `joiner_secret`, `welcome_secret`, `epoch_secret`, and
-  derived secrets for the new epoch.
-
 * Construct an MLSPlaintext object containing the Commit object. Sign the
-  MLSPlaintext using the old GroupContext as context. Use the
-  `confirmation_key` for the new epoch to compute the `confirmation_tag` value,
-  and the `membership_key` for the old epoch to compute the `membership_tag`
-  value in the MLSPlaintext.
+  MLSPlaintext using the old GroupContext as context.
+  * Use the MLSPlaintext to update the confirmed transcript hash and generate
+    the new GroupContext.
+  * Use the `init_secret` from the previous epoch, the `commit_secret` and the
+    `psk_secret` as defined in the previous steps, and the new GroupContext to
+    compute the new `joiner_secret`, `welcome_secret`, `epoch_secret`, and
+    derived secrets for the new epoch.
+  * Use the `confirmation_key` for the new epoch to compute the
+    `confirmation_tag` value, and the `membership_key` for the old epoch to
+    compute the `membership_tag` value in the MLSPlaintext.
+  * Calculate the interim transcript hash using the new confirmed transcript
+    hash and the `confirmation_tag` from the MLSPlaintext.
 
 * Construct a GroupInfo reflecting the new state:
   * Group ID, epoch, tree, confirmed transcript hash, and interim transcript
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2547,12 +2548,10 @@ message at the same time, by taking the following steps:
...
   populated.  Otherwise, the sender MAY omit the `path` field at its discretion.
 
-* If populating the `path` field: Create an UpdatePath using the new tree. Any
-  new member (from an add proposal) MUST be exluded from the resolution during
-  the computation of the UpdatePath. The GroupContext for this operation uses
-  the `group_id`, `epoch`, `tree_hash`, and `confirmed_transcript_hash` values
-  in the initial GroupContext object.  The `leaf_key_package` for this
-  UpdatePath must have a `parent_hash` extension.
+* If populating the `path` field: Create an UpdatePath using the new tree and
+  the provisional GroupContext. Any new member (from an add proposal) MUST be
```
Bren2010, line 2552: You say use the provisional GroupContext, but the original text says to use the original GroupContext

draft-ietf-mls-protocol.md
```diff
@@ -2570,10 +2569,14 @@ message at the same time, by taking the following steps:
...
-  `commit_secret` and the `psk_secret` to advance the key schedule and compute
-  the `confirmation_tag` value in the MLSPlaintext.
+* Use the `commit_secret`, the `psk_secret`, the provisional GroupContext, and
+  the init secret from the previous epoch to compute the epoch secret and
+  derived secrets for the new epoch.
+
+* Construct an MLSPlaintext object containing the Commit object. Sign the
+  MLSPlaintext using the current epoch's GroupContext as context. Use the
+  `confirmation_key` for the new epoch to compute the `confirmation_tag` value
+  in the MLSPlaintext.
```
Bren2010, line 2579: Again, mixing epochs seems bad

draft-ietf-mls-protocol.md
uhoreg: The "A member of the group applies a Commit message by taking the following steps:" section says

> * If the `path` value is populated: Process the `path` value using the new ratchet tree and the provisional GroupContext, to update the ratchet tree and generate the `commit_secret`:

So this change is to make the two sections consistent

draft-ietf-mls-protocol.md
uhoreg: This is how it works in the "A member of the group applies a Commit message by taking the following steps:" section.  It verifies the signature before doing any modifications, and uses the `confirmation_key` from the new epoch.

draft-ietf-mls-protocol.md
```diff
@@ -1289,6 +1289,7 @@ struct {
     opaque group_id<0..255>;
     uint64 epoch;
     Sender sender;
+    opaque authenticated_data<0..2^32-1>;
```
raphaelrobert, line 1292: The authenticated data is not part of the content of an `MLSPlaintext`, it only exists when `MLSCiphertext` is used

draft-ietf-mls-protocol.md
raphaelrobert: I agree it looks a bit inconsistent here indeed. I think this part is however the correct one and what you quote is confusing and should be changed instead.
The reason for my thinking is that the provisional group context is not known at this point in time and therefore the existing one should be used.

draft-ietf-mls-protocol.md
```diff
@@ -2570,10 +2569,14 @@ message at the same time, by taking the following steps:
   corresponds to the order of PreSharedKey proposals in the `proposals` vector.
   Otherwise, set `psk_secret` to a zero-length octet string.
 
-* Construct an MLSPlaintext object containing the Commit object. Sign the MLSPlaintext
-  using the current epoch's GroupContext as context. Use the signature, the
-  `commit_secret` and the `psk_secret` to advance the key schedule and compute
-  the `confirmation_tag` value in the MLSPlaintext.
+* Use the `commit_secret`, the `psk_secret`, the provisional GroupContext, and
+  the init secret from the previous epoch to compute the epoch secret and
+  derived secrets for the new epoch.
```
raphaelrobert, line 2574: This paragraph explains how the key schedule should be computed, but it omits quite a few steps (joiner secret, welcome secret, etc.) and the values are in the wrong chronologic order.

draft-ietf-mls-protocol.md
```diff
@@ -2632,8 +2635,8 @@ A member of the group applies a Commit message by taking the following steps:
   any Update or Remove proposals, or if it's empty. Otherwise, the `path` value
   MAY be omitted.
 
-* If the `path` value is populated: Process the `path` value using the ratchet
-  tree the provisional GroupContext, to update the ratchet tree and generate the
+* If the `path` value is populated: Process the `path` value using the new ratchet
+  tree and the provisional GroupContext, to update the ratchet tree and generate the
```
raphaelrobert, line 2639: See comment from above. I think this part is indeed wrong and the broken syntax might be a hint that it's the result of a copy & pasta job gone wrong.

raphaelrobert (changes requested):

draft-ietf-mls-protocol.md
uhoreg: I don't understand this comment.  `MLSPlaintext` does have an `authenticated_data` field, and the messages that are used here come from a Commit, which is encrypted, it would also use an `MLSCiphertext`.

Anyways, I had asked @bifurcation about this earlier, and he said that it was accidentally omitted and should be added, which is why I did this.  If it in fact should be omitted, then I can revert the change, but I think that we should either add the field, or else explain why it isn't there because every other field from the beginning of `MLSPlaintext` up to the `signature` field is included.

raphaelrobert: You are absolutely right, apologies for the confusion. The two should be aligned of course.

draft-ietf-mls-protocol.md
uhoreg: If I'm reading the mlspp sources right, it looks like it uses the provisional GroupContext for this: https://github.com/cisco/mlspp/blob/main/src/state.cpp#L347-L359

> The reason for my thinking is that the provisional group context is not known at this point in time and therefore the existing one should be used.

The provisional group context is created two steps before this step.

I personally don't care which one is used, but they should be consistent.  For now, I'm going to follow what mlspp does, but if the consensus is that the original text here is correct, then I'm happy to change it.

draft-ietf-mls-protocol.md
uhoreg: This was basically a copy-and-paste from the other section.  I can update both sections to be more accurate.

draft-ietf-mls-protocol.md
```diff
@@ -2572,12 +2587,16 @@ message at the same time, by taking the following steps:
...
   Otherwise, set `psk_secret` to a zero-length octet string.
 
-* Construct an MLSPlaintext object containing the Commit object. Sign the MLSPlaintext
-  using the current epoch's GroupContext as context. Use the signature, the
-  `commit_secret` and the `psk_secret` to advance the key schedule and compute
-  the `confirmation_tag` value in the MLSPlaintext.
+* Use the `init_secret` from the previous epoch, the `commit_secret` and the
+  `psk_secret` as defined in the previous steps, and the new GroupContext to
+  compute the new `joiner_secret`, `welcome_secret`, `epoch_secret`, and
+  derived secrets for the new epoch.
```
uhoreg, line 2593: Alternatively, I can say "Use the ... and the new GroupContext to advance the key schedule."  if we want to be less verbose.

draft-ietf-mls-protocol.md
```diff
@@ -2572,12 +2587,16 @@ message at the same time, by taking the following steps:
...
 
-* Update the tree in the provisional state by applying the direct path
```
uhoreg, line 2580: It looks like DirectPath was renamed to UpdatePath, and the UpdatePath is already applied in a previous step (as part of the step where it's actually created), so this looks to me like an obsolete leftover.

draft-ietf-mls-protocol.md
```diff
@@ -2572,12 +2587,16 @@ message at the same time, by taking the following steps:
...
+* Construct an MLSPlaintext object containing the Commit object. Sign the
+  MLSPlaintext using the old GroupContext as context. Use the
+  `confirmation_key` for the new epoch to compute the `confirmation_tag` value,
+  and the `membership_key` for the old epoch to compute the `membership_tag`
+  value in the MLSPlaintext.
```
uhoreg, line 2599: I'm not entirely sure why it takes the `confirmation_key` from the new epoch, but the `membership_key` from the old epoch, but that's what the current mlspp code does.

uhoreg:
I've tried to align with the current mlspp code for how to process commits.  If the current mlspp code is incorrect, or if I'm reading it incorrectly, I'm more than happy to fix my changes to reflect the correct behaviour.

draft-ietf-mls-protocol.md
uhoreg: Oh, actually, I do know, I think.  The `confirmation_tag` allows the other group members to check that they derived the same tree/secrets as the sender.  The `membership_tag` allows other group members to ensure that it was sent by a group member before processing the message.  I think this is mentioned in other parts of the document, but I can reiterate here if necessary.

draft-ietf-mls-protocol.md
seanturner: Discussed at 20210526 Interim. Will split out authenticated_data change to another PR so it can be merged. Editorial changes continue to be reviewed.

seanturner:
@bifurcation @raphaelrobert took action item to review editorial changes.

draft-ietf-mls-protocol.md
```diff
@@ -2532,6 +2532,22 @@ uses:
    regard to any removed members and PCS for the committer and any updated
    members.
 
+When processing a Commit, three different GroupContexts are used:
```
bifurcation, line 2535:
```suggestion
When creating or processing a Commit, three different GroupContexts are used:
```

draft-ietf-mls-protocol.md
bifurcation: I think "provisional" is the right answer here.  The idea of the GroupContext used with TreeKEM is to capture the tree to which the secrets are encapsulated.  That tree is the one in the provisional GroupContext, to which the proposals have been applied.

draft-ietf-mls-protocol.md
bifurcation: This should be "provisional" in both cases -- the tree after applying proposals but before applying the UpdatePath.

uhoreg: Uhm, yes.  Obviously it can't be the "new" ratchet tree, since this is the step that creates the "new" ratchet tree.  Not sure why I put "new" there.

draft-ietf-mls-protocol.md
uhoreg: I think I've addressed this comment

draft-ietf-mls-protocol.md
uhoreg: Oh, it's because we've only defined "old", "provisional", and "new" for GroupContext, and not for the ratchet tree.  I'll adjust the wording to fix that up.

uhoreg:
I think I've addressed all the comments.  I also realized that the calculation of the new GroupContext was in the wrong spot (it has to be done after the MLSPlaintext is created (sans confirmation_tag), and then the confirmed transcript hash is created, then the new GroupContext, then the secrets, then the confirmation tag -- that's what mlspp does AFAICT), so I've fixed that up as well.

bifurcation:
@raphaelrobert I think if you're OK with this, we're clear to merge.

bifurcation (approved):

draft-ietf-mls-protocol.md
```diff
@@ -2532,14 +2532,32 @@ uses:
...
+1. "Old" refers to the ratchet tree and GroupContext for the epoch before the
+   commit.  The old GroupContext is used when signing the MLSPlainText so that
+   existing group members can verify the signature before processing the
+   commit.
+2. "Provisional" refers to the ratchet tree and GroupContext constructed after
+   applying the proposals that are referenced by the Commit.  The provisional
+   GroupContext uses the epoch number for the new epoch, and the old confirmed
+   transcript hash.  This is used when creating the UpdatePath, if the
+   UpdatePath is needed.
+3. "New" refers to the ratchet tree and GroupContext constructed after applying
```
raphaelrobert, line 2547: I'm still struggling with the fact that there are 3 ratcheting trees. In my mind there are only 2:
 - The ratcheting tree *before* applying proposals and the (potential) UpdatePath
 - The ratcheting tree *after* applying proposals and the (potential) UpdatePath

The ratcheting trees described here as "provisional" and "new" are in fact identical. It's just that they exist at different times. I think it would be good to make that clear, i.e. a "provisional" ratcheting tree becomes a "new" ratcheting tree when a Commit is applied.

raphaelrobert (changes requested):
One more comment about the 3 ratcheting trees

draft-ietf-mls-protocol.md
uhoreg: The "provisional" ratchet tree is the tree after applying the proposals, but before applying the UpdatePath.  We could probably do without a "provisional" ratchet tree, but I think it helps with clarity to give it a name since applying the proposals and applying the UpdatePath are different steps.

seanturner: Discussed at 20211004 MLS interim.

seanturner:
As discussed at interim. @raphaelrobert review. If good, then merge.

draft-ietf-mls-protocol.md
raphaelrobert: Fine, that makes sense to me!

raphaelrobert (approved):
Ready to be merged from my side.

PR #455, opened 2021-02-11
Make PreSharedKeys non optional in GroupSecrets

raphaelrobert:
Currently the `psks` field in `GroupSecrets` is optional. `PreSharedKeys` only contains an array, which can be empty, even if the option is not empty. To avoid undefined behavior, I removed the option. In case no PSKs are used, the array should simply be empty.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2812,7 +2812,7 @@ struct {
 struct {
   opaque joiner_secret<1..255>;
   optional<PathSecret> path_secret;
-  optional<PreSharedKeys> psks;
+  PreSharedKeys psks;
 } GroupSecrets;
 
 struct {


Bren2010 (approved):

beurdouche:
I don't think the semantics of what you are suggesting is correct with respect to the TLS syntax. If you want to remove the option then the value has to be something, zeros for example.

beurdouche:
This is not a strong objection though...

raphaelrobert:
@beurdouche Can you point to the TLS syntax for that one? What I'm concerned about is specifically the UB, there has to be a better way to define this.

seanturner:
Discussed at 20210526 Interim: As long as there is some input you can have an empty list. No impact on security.

seanturner:
merge

PR #463, opened 2021-03-08
Remove AppAck proposal.

Bren2010:


seanturner:
As discussed at 20211004 interim, will retain text.

PR #464, opened 2021-03-09
Clarify which extensions should go into the group context

kkohbrok:
Just a small editorial change to clarify which extensions should be included in the `GroupContext`. There is a paragraph explaining that the `RatchetTreeExtension` should not go in, but I think the section on the `GroupContext` should be precise about what _does_ go in.

I did notice, though, that we currently don't have a way of actually changing extensions during the lifetime of the group. If someone wanted to change an extensions, they would currently have to re-init the whole group. Of course, this could be solved by an extension, but maybe it would be sensible to have an `Extension` proposal for that purpose?


kkohbrok:
We should decide if we want to tackle #473 before discussing this further.

kkohbrok:
Closing in favor of the approach decided upon in #473.

PR #466, opened 2021-03-19
Minor fixes to message formats

bifurcation:
This PR fixes a few small issues I noticed while reviewing mlspp for compliance with the spec:

* The size of the `Extension.extension_data` vector needs to expand `2^32-1` to support trees for large groups
* The content of the `parent_hash` extension is specified
* `authenticated_data` is included in `MLSPlaintextCommitContent` so that it goes into the transcript
* `MLSPlaintextAuthData` now carries an `optional<MAC>` even though the value is always present, so that the transcript matches the data on the wire

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -448,8 +448,8 @@ directly to the new member (there's no need to send it to the group). Only after
 A has received its Commit message back from the server does it update its state
 to reflect the new member's addition.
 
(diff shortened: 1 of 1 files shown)


Bren2010 (approved):

kkohbrok (approved):

seanturner:
Discussed at 20210526 Interim. Merge

PR #467, opened 2021-03-30
An entropy pool design for MLS

kkohbrok:
This PR mandates the use of an entropy pool for MLS and proposes a design that is modeled after the key schedule such that it
1) allows the extraction of entropy when needed,
2) allows the repeated injection of entropy originating from the system's RNG such that the entropy pool gets refreshed after a full state compromise,
3) gathers entropy over time while maintaining forward-secrecy,
4) allows the safe injection of entropy originating from other parties.

In particular, the design is such that we achieve what we believe are the best possible security guarantees from any injection operation:
* if the entropy from the system's RNG is good, the result is good independent of any other input to the pool
* if the entropy from one of the groups is good, the result is good independent of any other input to the pool

TODO:
* allow the injection of auxiliary randomness, for example a signature using a long term signature key as suggested by https://eprint.iacr.org/2018/1057

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3286,64 +3286,95 @@ copath node corresponding to the node in the UpdatePath.
 
 ## Improving Resiliance to Poor Random Number Generators
 
-The security of a group hinges on the fact that every member has access to a
-good source of entropy. To make the protocol more robust in environments with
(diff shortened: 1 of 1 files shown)


kkohbrok:
Closing this PR as the entropy pool specification is moved to an "implementation-considerations" document.

PR #471, opened 2021-05-18
Use HPKE draft-08

franziskuskiefer:
draft-08 is the latest HPKE draft, which should be stable. It is also what's being used for interop testing right now.


bifurcation (approved):

seanturner:
Discussed 20210526 Interim: Merge.

PR #476, opened 2021-08-13
Revisit the notion of identity in MLS groups

kkohbrok:
After a discussion with @bifurcation and @raphaelrobert, this PR is a new spin on and an extension of #439.

This PR adds the `endpoint_id` field to KeyPackages. The `endpoint_id` field serves to distinguish KeyPackages with Credentials that share the value of their `identity` field.

This PR also introduces a set of restrictions for KeyPackages that introduced to the group via Adds and Updates.

Generally, for both Adds and Updates, the committer has to ensure that
- signature keys are unique within a group (this is so that the sender of a message can be determined uniquely within a group),
- `(identity, endpoint_id)` tuples are unique in a group,
- `hpke_init_key`s are unique in a group.

Additionally, Updates have to ensure that
- the new and the old key package have the same `identity` (in the Credential), `endpoint_id`, `version` and `cipher_suite` as the key package they replace,
- the new and the old key package do not share the same `hpke_init_key`.

I didn't extend the section indicating that the processor of a commit has to check all these things in the same way that the committer has to. I think, here we should probably restructure the section, so that we can just reference the part where it is specified how to determine if a given proposal is valid in the context of a given group.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2492,8 +2492,13 @@ If there are multiple proposals that apply to the same leaf, the committer
 chooses one and includes only that one in the Commit, considering the rest
 invalid. The committer MUST prefer any Remove received, or the most recent
 Update for the leaf if there are no Removes. If there are multiple Add proposals
-for the same client, the committer again chooses one to include and considers
-the rest invalid.
+containing KeyPackages with the same tuple `(credential.identity, endpoint_id)`
+the committer again chooses one to include and considers the rest invalid. Add
+proposals that contain KeyPackages with an `(credential.identity, endpoint_id)`
+tuple that matches that of an existing KeyPackage in the group MUST be
+considered invalid. The comitter MUST consider invalid any Add or Update
+proposal if the Credential in the contained KeyPackage shares the same signature
+key with a Credential in any leaf of the group.
 
 The Commit MUST NOT combine proposals sent within different epochs. In the event
 that a valid proposal is omitted from the next Commit, the sender of the
@@ -2570,8 +2575,11 @@ message at the same time, by taking the following steps:
   new member (from an add proposal) MUST be exluded from the resolution during
   the computation of the UpdatePath. The GroupContext for this operation uses
   the `group_id`, `epoch`, `tree_hash`, and `confirmed_transcript_hash` values
-  in the initial GroupContext object.  The `leaf_key_package` for this
-  UpdatePath must have a `parent_hash` extension.
+  in the initial GroupContext object. The `leaf_key_package` for this UpdatePath
+  must have a `parent_hash` extension. Note, that the KeyPackage in the
+  `UpdatePath` effectively updates an existing KeyPackage in the group and thus
+  MUST adhere to the same restrictions as KeyPackages used in `Update`
+  proposals.
 
    * Assign this UpdatePath to the `path` field in the Commit.
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1007,13 +1007,11 @@ provide some public information about a user. KeyPackage
 structures provide information about a client that any existing
 member can use to add this client to the group asynchronously.
 
-A KeyPackage object specifies a ciphersuite that the client
-supports, as well as providing a public key that others can use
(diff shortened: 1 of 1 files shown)


br-hale:
Thank for proposing this PR, Konrad. The changes look beneficial.

I have one question about the requirement for updates that the ciphersuite matches the prior ciphersuite. Currently, all group members are required to use the same ciphersuite so this requirement does not seem necessary in most cases. There is one case worth considering though that this restricts - namely that it limits the option for the entire group to roll the ciphersuite simultaneously. It would be advantageous for us to support changing of the identity key ciphersuite for the entire group, if necessary, without a full tear-down of the state. I do not think we currently specify options for this in the spec, but a ciphersuite matching requirement would inhibit the possibility.

kkohbrok:
Thanks for the feedback @br-hale!

I agree that the requirement for all group members to use matching ciphersuites was there before, but I couldn't find it anywhere explicitly, so I added it just to be sure. If it turns out to be redundant, I'm happy to remove it again.

If I understand your proposal correctly, rolling the ciphersuite simultaneously would require a few more changes, for a example an additional proposal to coordinate the change. If that is the case, I'd say we leave it in and remove the requirement once we add the capability for the ciphersuite rolling. A problem I see with ciphersuite rolling is that it would require everyone to change their Credential simultaneously. Although maybe we could overcome that by coordinating Update proposals? Do you want to open a new GH issue/PR so you can outline your idea and we can continue the discussion there?

raphaelrobert:
On the topic of ciphersuite change: we have a mechanism for that in 10.1:

> Re-initializing an existing group may be used, for example, to restart the group with a different ciphersuite or protocol version.

I would assume that the binary tree wouldn't be re-usable anyway if different crypto primitives are used so it's only a matter of preserving existing entropy and carry over group secrets to the new group which can be done through the PSK mechanism. Am I missing something here?

br-hale:
What @raphaelrobert mentions is where my comment was vectored towards. The concern would be if locking in the ciphersuite per this PR blocks that option entirely. However, it is correct that the PSK mechanism can be used for creating a new group (with new ciphersuite) that ties to the prior group and therefore bypasses the need need to roll a ciphersuite within a group. It may be a heavier change mechanism than rolling a ciphersuite would be, but it works. 

Since this is specifically about rolling identity keys, the rest of the tree should still be usable following a roll of the ciphersuite (signatures properties active after the next update of leaf node keys of course). But @raphaelrobert you make a good point that any such handling would then need to prevent rolling of other ciphersuite elements aside from the identity. Such changes would need careful consideration - at this point in the process, it is best not to introduce something too nuanced. So, all that said, I think we can close the question on ciphersuite matching.

kkohbrok:
I just realized that one corollary of this change is that we can now use the `(identity, endpoint_id)` tuple in the message structs and thus further reduce our dependency on any specific binary tree implementation. Something like this:

```
struct {
    SenderType sender_type;
    opaque sender_identity;
    opaque sender_endpoint_id;
} Sender;
```
And similarly for the case of `MLSSenderData`, the `Remove` proposal, `MessageRange`, `PublicGroupState`, `GroupInfo` and anything else that's not immediately part of the tree primitives, at least for now.

draft-ietf-mls-protocol.md
```diff
@@ -2259,6 +2258,12 @@ struct {
 } Update;
 ~~~~~
 
+The KeyPackage contained in an `Update` proposal MUST share the values of the
```
bifurcation, line 2261: Rather than say "share" here, it would be clearer to say that the values MUST be equal.

bifurcation: Likewise below, instead of "MUST NOT share", "MUST be different".

bifurcation (approved):
Thanks, @kkohbrok this is a useful clarification.  It seems fine to handle the receiver processing in a separate PR; consolidating it as you suggest is a good idea.

draft-ietf-mls-protocol.md
raphaelrobert: Agreed

raphaelrobert (approved):
Looks good, modulo @bifurcation's remarks

bifurcation (approved):

PR #477, opened 2021-08-16
Signal the intended wire format for MLS messages

bifurcation:
@kkohbrok pointed out to me that there's currently no way for a DS to distinguish between MLSPlaintext and MLSCiphertext messages.  Obviously, an application could provide a flag on the outside of a message, but there are a few relatively minor benefits to including a flag in MLS:
 
 * Signing the flag authenticates the sender's intent
 * Including the flag in the transcript confirms that all group members got it in the same format
 * Including the flag in the struct provides a clear signal as to whether the `membership_tag` needs to be present/verified

Since these benefits are pretty minor, I could kind of go either way on whether to make this change.  But it seemed worth writing up a PR to see what the change would look like.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -448,8 +448,8 @@ directly to the new member (there's no need to send it to the group). Only after
 A has received its Commit message back from the server does it update its state
 to reflect the new member's addition.
 
-Upon receiving the Welcome message, the new member will be able to read and send 
-new messages to the group. Messages received before the client has joined the 
+Upon receiving the Welcome message, the new member will be able to read and send
+new messages to the group. Messages received before the client has joined the
 group are ignored.
 
 ~~~~~
@@ -1286,6 +1286,7 @@ The following general rules apply:
 
 ~~~~~
 struct {
+    WireFormat wire_format;
     opaque group_id<0..255>;
     uint64 epoch;
     Sender sender;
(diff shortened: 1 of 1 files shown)


kkohbrok:
My original intention was not for the DS to distinguish the message types, but also for a receiving client. Even if there's a group policy that dictates whether to encrypt handshake messages, the client won't know how to parse a message, because the `content_type` fields on plaintext and ciphertext are not aligned. So I think fixing that would be the low-hanging fruit.

That being said, I agree with @bifurcation that it would be neat to have a flag that signals the encryption status, but also communicates the sender's intent in a cryptographically authenticated manner.

raphaelrobert:
I agree with your reasoning and I think it's worthwhile doing.

The only part that is not straight forward for me is this:

> Including the flag in the struct provides a clear signal as to whether the membership_tag needs to be present/verified

If the sender is of type `external` the `membership_tag` also doesn't need to be present for an `MLSPlaintext`. This should either be clarified, or we could have a third wire format for external senders.

bifurcation:
Re: `external` -- there isn't an `external` sender type, but there are `new_member` and `preconfigured`.  You're correct that in both of those cases, the `membership_tag` shouldn't be populated.  The signal in this PR helps distinguish the two cases when `sender_type == member`.

kkohbrok (approved):

raphaelrobert (approved):

seanturner:
Discussed at 20211004 interim, merge.

PR #478, opened 2021-08-19
Inject GroupContext as HPKE info instead of AAD

### LINKED ISSUE #470: group_context is passed into HPKE's aad instead of info
This came up during a discussion with @raphaelrobert and @franziskuskiefer on an issue discovered by Théophile (https://github.com/openmls/openmls/pull/404).

I don't think this is a security issue necessarily, as both sender and receiver must agree on the group_context to successfully decrypt. However, I believe it's cleaner to have `group_context` go into the HPKE key derivation rather than the AEAD's aad.

bifurcation:
Fixes #470

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -448,8 +448,8 @@ directly to the new member (there's no need to send it to the group). Only after
 A has received its Commit message back from the server does it update its state
 to reflect the new member's addition.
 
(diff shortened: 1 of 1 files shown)


kkohbrok (approved):

raphaelrobert (approved):

seanturner:
Discussed at 20211004 interim, merge.

PR #479, opened 2021-08-19
Clarify extension handling and make extension updatable

### LINKED ISSUE #473: Extensions
After a thorough review of the current MLS extension system, there are a few questions and we also have a few suggestions for changes.

#### How we interpret the spec currently:
 - The total set of extensions is fixed and cannot be changed during the lifetime of a group.
 - The set of extensions that is associated with a group is decided by the group creator.
 - All extension types from the set have to be supported by all clients and must therefore be advertised in their capabilities extension.
 - The implementation needs to validate the support in 4 instances: When creating a new group, when joining a group, when creating an add proposal, when parsing someone else's add proposal as part of a Commit.
 - Key package extensions are simple and well-understood, in that they never leave the KeyPackage and there is group consensus through the tree hash.
 - The other kind of extension seems to be called a Group Info Extension (GIE).
 - GIEs can come in two flavours: a) they must be included in the group context, b) they must not be included in the group context.
 - The ratchet tree extensions is an example of type b).
 - GIEs of type a) are essentially just passed from the Adder to the New Joiner, using Welcome messages as an E2EE pipe.

#### Open questions:
 - The spec does not explicitly introduce the notion of a GIE, it is only mentioned in 16.2. Should the spec be more explicit here?
 - Should the flavour of a GIE (as described above) be reflected in the `Extension` struct in the spec? E.g. a message type that indicates whether a GIE should be included in the GroupContext (see proposal below).
 - Can the content of a GIE be dynamic (i.e. it can change during the lifetime of a group)? The ratchet tree extension seems to validate that assumption, it is a bit unclear if this is true for all kinds of GIE. We should be more explicit in the spec.
 - If the content of GIEs is indeed dynamic, should those GIEs always be excluded from the GroupContext? Or optionally? If they are to be included in the GroupContext, how do clients agree on the changes? (See proposal below)

We need more guidance regarding extensions in order to keep implementations aligned and capable of interop.

#### Concrete proposed improvements:

A first suggestion would be to allow for the creation of extensions that have dynamic content. This would allow them to use MLS to agree on arbitrary data.

The requirement would be to make GIEs with dynamic content more robust by introducing a new proposal type to update the content of extensions:

```c
struct {
  Extension extension;
} ExtensionUpdate;
```

Another improvement to the current extension system would be to be more clear about the kinds of extensions by introducing a dedicated `MessageType` field for every extension:

```c
enum {
    reserved(0),
    KeyPackage(1),
    GroupContext(2), // GIEs that are to be included in the group context
    GroupInfo(3), // GIEs that should not be included in the group context, like the ratchet tree extension
    (255)
} MessageType;

struct {
    ExtensionId extension_id;
    MessageType message_type;
    opaque extension_data<0..2^16-1>;
} Extension;
```

This would allow generic `GroupInfo` extensions, which in turn allow committers of Add proposals to communicate arbitrary data to new group members in a robust fashion.

bifurcation:
Fixes #473

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2391,6 +2391,26 @@ to recognize which messages conatain AppAcks.  The application can also have
 clients enforce an AppAck schedule, reporting loss if an AppAck is not received
 at the expected time.
 
+### GroupContextExtensions
+
+A GroupContextExtensions proposal is used to update the list of extensions in
+the GroupContext for the group.
+
+```
+struct {
+  Extension extensions<0..2^32-1>;
+} GroupContextExtensions;
+```
+
+A member of the group applies a GroupContextExtensions proposal by removing all
+of the existing extensions from the GroupContext object for the group and
+replacing them with the list of extensions in the proposal.  (This is a
+wholesale replacement, not a merge.  An extension is only carried over if the
+sender of the proposal includes it in the new list.)  Note that once the
+GroupContext is updated, its inclusion in the confirmation_tag by way of the key
+schedule will confirm that all members of the group agree on the extensions in
+use.
+
 ### External Proposals
 
 Add and Remove proposals can be constructed and sent to the group by a party
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -448,8 +448,8 @@ directly to the new member (there's no need to send it to the group). Only after
 A has received its Commit message back from the server does it update its state
 to reflect the new member's addition.
 
-Upon receiving the Welcome message, the new member will be able to read and send 
-new messages to the group. Messages received before the client has joined the 
+Upon receiving the Welcome message, the new member will be able to read and send
+new messages to the group. Messages received before the client has joined the
 group are ignored.
 
 ~~~~~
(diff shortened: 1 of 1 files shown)


bifurcation:
@kkohbrok r?

draft-ietf-mls-protocol.md
```diff
@@ -1312,8 +1312,8 @@ interim_transcript_hash_[n+1] =
 Thus the `confirmed_transcript_hash` field in a GroupContext object represents a
 transcript over the whole history of MLSPlaintext Commit messages, up to the
 confirmation tag field in the current MLSPlaintext message.  The confirmation tag
-is then included in the transcript for the next epoch.  The interim transcript
-hash is passed to new members in the GroupInfo struct, and enables existing
+is then included in the transcript for the next epoch.  The is computed new
```
kkohbrok, line 1315: Something seems to be missing here.

draft-ietf-mls-protocol.md
```diff
@@ -2886,9 +2910,11 @@ welcome_key = KDF.Expand(welcome_secret, "key", AEAD.Nk)
   divided by two.
 
 * Construct a new group state using the information in the GroupInfo object.
-  The new member's position in the tree is `index`, as defined above.  In
-  particular, the confirmed transcript hash for the new state is the
-  `prior_confirmed_transcript_hash` in the GroupInfo object.
+    * The GroupContext contains the `group_id`, `epoch`, `tree_hash`,
+      `confirmed`transcript_hash`, and `group_context_extensions` fields from
```
kkohbrok, line 2914:
```suggestion
      `confirmed_transcript_hash`, and `group_context_extensions` fields from
```

draft-ietf-mls-protocol.md
bifurcation: "interim transcript hash", apparently.

kkohbrok (approved):

raphaelrobert (approved):

seanturner:
Discussed at 20211004 interim, ready to merge after adding some additional detail about when to use one or the other.

PR #480, opened 2021-08-19
Improve extensibility of Proposals

bifurcation:
@psyoptix pointed out to me some improvements that could be made in the extensibility of proposals.  This PR makes a few changes to smooth out deployment of new proposals:

* Extends `ProposalType` to two bytes, creates an IANA registry, and reserves some space for private use
* Extends the `capabilities` extension so that clients can indicate which proposal types they support
* Forbids committers from including proposals that some people don't support


kkohbrok (approved):
Looks good to me!

raphaelrobert (approved):

raphaelrobert:
Just for context: This PR addresses the second part of #473.

PR #481, opened 2021-08-19
Constrain proposal in External Commit

### LINKED ISSUE #443: External commit for resync used with PSK
A good point was raised by Jonathon Hoyland during the MLS IETF 109 meeting regarding possible concerns in using external commits for resync, particularly in the case of Alice adding/removing herself. Richard noted that this is a feature in the case that Alice is no longer synchronized with the group and therefore can use an external commit to add herself back in, removing the previous version.
 
As opposed to any newcomer joining with an external commit, the case of Alice re-joining presents a potential security issue. Namely, as currently specified (in my reading of the draft), an existing group member, Bob, has no means to distinguish between the following cases:

1) Alice needs to resync and therefore performs an external commit and removes her prior version.
2) Alice’s signature keys are compromised (it is not necessary for the adversary to compromise any group state). The adversary performs an external commit in Alice’s name, and then removes her prior version and impersonates her to the group.
 
One might hope that Alice notices that she is removed and communicates this to the group members OOB, but it is also possible that that she assumes some other reason for the removal, is offline, or simply is not active enough to take action for a fairly long compromise window. Even if she tries to use an external commit to get back into the group and then removes the adversary-as-Alice, there is no means for other group members distinguish the real Alice from the adversary-as-Alice and the process could be circular (until new valid identity keys are issued).
 
While a newcomer is a fresh source to be trusted or not, Alice has been “healing” along with the group and the above option (2) allows the adversary to bypass all of that.
 
The source of the problem is that when Alice re-syncs, she is not providing any validation of being the same/previous identity, so it is easy for other group members to accept that nothing more than a resync has taken place. Thus, a fairly straightforward solution is to require PSK use in cases where an external commit is used for resync. By enabling a PSK derived from a previous epoch during which Alice was part of the group to be injected with the external commit, Alice provides some proof of prior group membership and we avoid the total reset.

This is not quite PCS in that the attacker is active following compromise, but it is a linked case. As such it is important that a conscious decision is made regarding this (either as a slight change before the feature-freeze in line with the other changes that the editors have proposed, or as a working group decision to close the issue as out-of-scope).

bifurcation:
Fixes #443

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -448,8 +448,8 @@ directly to the new member (there's no need to send it to the group). Only after
 A has received its Commit message back from the server does it update its state
 to reflect the new member's addition.
 
-Upon receiving the Welcome message, the new member will be able to read and send 
-new messages to the group. Messages received before the client has joined the 
+Upon receiving the Welcome message, the new member will be able to read and send
+new messages to the group. Messages received before the client has joined the
 group are ignored.
 
 ~~~~~
@@ -1539,6 +1539,7 @@ enum {
   external(1),
   reinit(2),
   branch(3),
+  resync(4)
   (255)
 } PSKType;
 
@@ -1554,6 +1555,9 @@ struct {
 
     case branch:
       opaque psk_group_id<0..255>;
+      uint64 psk_epoch;
+
+    case resync:
       uint64 psk_epoch;
   }
   opaque psk_nonce<0..255>;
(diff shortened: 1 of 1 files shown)


bifurcation:
@br-hale @kkohbrok r?

draft-ietf-mls-protocol.md
```diff
@@ -1539,6 +1539,7 @@ enum {
   external(1),
   reinit(2),
   branch(3),
+  resync(4)
```
br-hale, line 1542: I am not adverse to a resync option. However, this was in the original PR for PSKs, discussed by the group then removed. At that time it was called "recover" (original WG brief") and later "resumption". There was fairly extensive discussion on it, so we should probably double-check what was kept/removed and why in the mailinglist and meeting notes before re-introducing this.

draft-ietf-mls-protocol.md
```diff
@@ -2757,21 +2761,34 @@ External Commits.
...
 
-* External Commits MUST reference an Add Proposal that adds the issuing new
-  member to the group
+* The proposals included by value in an External Commit MUST meet the following
+  conditions:
+  * There MUST be a single Add proposal that adds the new issuing new member to
+    the group
+  * There MUST be a single ExternalInit proposal
+  * There MUST NOT be any Update proposals
+  * If a Remove proposal is present, then:
```
br-hale, line 2770: We would need to require that the prior identity (if there was prior membership) is removed. The proposer should do this - so maybe mandating a "must" for including a remove proposal if resyncing. Basically, the group should not end up with duplication of the identity (and potential backdoor). It is also inefficient for healing as Alice needs to replace keys at both nodes if both of the "Alice" identities are allowed to stay.

This is presents a question for handling sub-identities (if Alice has multiple devices and the identity is linked to Alice vs. a device). In that case an Add proposal for extra devices could include a PSK derived from the group, but the group will never know which device Alice is using (since the identity is the same). So what is the assumption here with respect to multi-device handling?

draft-ietf-mls-protocol.md
br-hale: Actually the "resumption" part was kept as a separate key from the key schedule, but downgraded from use as group "recovery" and applied strictly to branch/reinit. PR #421 shows the changed text.

draft-ietf-mls-protocol.md
kkohbrok: I believe PR #476 would allow us to check both user _and_ device identity. But of course this depends on how the application implements multi-device. They could user a different `identity` for each device, or a different `endpoint_id` or both.

draft-ietf-mls-protocol.md
kkohbrok: I'm afraid, I can't recall the reasons either. I still think this might be useful to have, though.

draft-ietf-mls-protocol.md
```diff
@@ -3219,6 +3229,24 @@ The second form of authentication is that group members can verify a message
...
+the group -- even using a "resync"-style external commit to replace the
+compromised member in the group.
+
+Applications can mitigate the risks of signature key compromise using pre-shared
+keys.  If a group requires joiners to know a PSK in addition to authenticating
+with a credential, then in order to mount an impersonation attack, the attacker
+would need to compromise the relevant PSK as well as the victim's signature key.
+The cost of this mitigation is that the application needs some external
+arrangement that ensures that the legitimate members of the group to have the
+required PSKs.
```
raphaelrobert, line 3248: Maybe worth noting that resumption secrets can be optionally used to mitigate the signature key compromise in the "resync" scenario.

raphaelrobert (changes requested):

seanturner:
Discussed at 20211004 interim, remove resynch with a single op (external join + remove in one op).

@bifurcation research why resynch was dropped.

PR #484, opened 2021-08-19
Clarify node vs. leaf indices

### LINKED ISSUE #432: Remove the concept of leaf indices
We currently distinguish between leaf indices and node indices, where node indices are a straight-forward index in the array representing a ratcheting tree and leaf indices index what (I think) used be called the roster, an array just containing the members. However, the roster was removed and instead, we always use the ratcheting tree array when referring to both nodes and leaves. As a result, whenever we use a leaf index, we have to translate it to a node index when actually indexing the array. I don't see a reason for this added complexity and propose only use node indices.

This was already discussed a bit in #431 , where @bifurcation pointed out that leaf indices make sure that a Remove only ever references a leaf. I think having a little check here that the index actually refers to a leaf is preferable over having to translate back and forth in many other places. Note, that we have to check anyway if the index in Remove is within the bounds of the tree, regardless of it being a leaf or a node index.

I'm happy to provide a PR if this is something we want to do.

bifurcation:
Fixes #432

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1569,8 +1569,8 @@ computed as follows:
 ~~~~~
 struct {
     PreSharedKeyID psk_id;
-    uint16 psk_index;       // i
-    uint16 count;           // n
+    uint16 psk_index;       // i in the below notation
+    uint16 count;           // n in the below notation
 } PSKLabel;
 
 psk_input_[i] = KDF.Extract(0, psk_[i])
@@ -2037,7 +2037,7 @@ encrypted, the sender data is encoded as an object of the following form:
 
 ~~~~~
 struct {
-    uint32 sender;
+    uint32 sender; // leaf index
     uint32 generation;
     opaque reuse_guard[4];
 } MLSSenderData;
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1573,9 +1568,9 @@ computed as follows:
 
 ~~~~~
 struct {
-    PreSharedKeyID id;
-    uint16 index;
-    uint16 count;
+    PreSharedKeyID psk_id;
+    uint16 psk_index;       // i
```
kkohbrok, line 1572: Could you elaborate what hat are "i" and "n" supposed to mean here? Or are those just typos?

draft-ietf-mls-protocol.md
```diff
@@ -2078,10 +2073,10 @@ struct {
...
 ~~~~~
 
-When parsing a SenderData struct as part of message decryption, the
-recipient MUST verify that the sender field represents an occupied
-leaf in the ratchet tree.  In particular, the sender index value
-MUST be less than the number of leaves in the tree.
+When parsing a SenderData struct as part of message decryption, if the
+SenderType is `member`,  the recipient MUST verify that the sender field
+represents an occupied leaf in the ratchet tree.  In particular, the sender
+index value MUST be less than the number of leaves in the tree.
```
kkohbrok, line 2079: Maybe mark the `sender` field in `MLSSenderData` as of type `// leaf index`?

kkohbrok (commented):
This certainly helps making thinks a bit clearer. I'd still feel better if we got rid of the notion of a "leaf index" entirely, though.

draft-ietf-mls-protocol.md
bifurcation: Good idea.  I think this paragraph is unnecessary anyway.

seanturner:
Discussed at 20211004 interim, will be reworked to remove leaf indices. And, be consistent through I-D.

bifurcation:
Replaced by #487

PR #487, opened 2021-10-04
Remove the notion of a 'leaf index'

### LINKED ISSUE #432: Remove the concept of leaf indices

bifurcation:
After discussion at the interim 2021-10-04, there seemed to be agreement to move from having two sets of indices (leaf and node indices) to just using node indices.  The main impact of this is that wherever there is a node index in the protocol that is only ever supposed to refer to a leaf node (`Remove.removed`, `GroupInfo.signer_index`, `MLSPlaintext.sender.sender`, `PublicGroupState.signer_index`), the recipient of the message needs to verify that the node index is actually a leaf.  However, implementations already needed to check that these fields were populated leaves, so the change should be minor.

Implementations that were using typed message structures to separate node and leaf indices can continue to use those types to assure that validation is done.  They'll just have to multiply by two on serialize and divide on deserialize if they use leaf indices as before internally.

There are also a couple of small, unrelated, "drive-by" fixes in here:
* Clean up of notation in the ratchet tree section
* Including PublicGroupState signer index in signature (in parallel with GroupInfo)

Replaces #484 
Fixes #432

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -617,7 +617,7 @@ A   B   C   D   E   F   G
 0 1 2 3 4 5 6 7 8 9 0 1 2
 ~~~~~
 
-Each node in the tree is assigned a _node index_, starting at zero and
+Each node in the tree is assigned an _index_, starting at zero and
 running from left to right.  A node is a leaf node if and only if it
 has an even index.  The node indices for the nodes in the above tree
 are as follows:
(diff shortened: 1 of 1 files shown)


kkohbrok (approved):

rohan-wire:
I think we should still use "node index" instead of just "index" for clarity.

rohan-wire:
I am concerned that concatenating the children to get the name of a node becomes awkward with blank nodes. As we already have the node index, why not just use it to identify non-leaf nodes?
~~~~~
           7 (root)
           /      \
          /        \
         /          \
      3              11
    /    \          /  \
   /      \        /    |
  1       5       9     |
 / \     / \     / \    |
A   B   _   D   E   F   G

                    1 1 1
0 1 2 3 4 5 6 7 8 9 0 1 2
~~~~~

This prevents node 5 from being D or _D

bifurcation:
Fair enough.  It also helps emphasize that parent nodes are of a different character than leaf nodes.  Just pushed a commit with numeric parent node labels and otherwise improved ASCII art.

PR #489, opened 2021-10-04
Add RequiredCapabilities extension

### LINKED ISSUE #485: Mandatory extension types?
Many protocols which have an extensibility mechanism have a way to indicate that understanding an extension is mandatory for certain behaviors. For MLS, this would mean having a way to require a member understand an MLS extension before it can join a group. The three extension types defined currently are: 
 MLS ciphersuites, 
 MLS extension types, and 
 MLS credential types. 

One logical way to add mandatory extensions would be to pick one of these types (the MLS extensions type) and use the high order bit or a specific range to indicate that a member must understand and implement the extension to join a group which uses that extension. The IANA values could be changed as follows:

Values from 0x8000-0xffff indicate understanding the extension is required
0x7f00 - 0x7fff  Reserved for Private Use (not required)
0xff00 - 0xfff   Reserved for Private Use (required)

If a required ciphersuite or credential becomes necessary, a new extension type could be created to add an extension in the KeyPackage or GroupInfo which would be required and would indicate support for the mandatory ciphersuites or credentials.

In order to prevent some other entity from adding an MLS client that does not support a mandatory extension, we already have the KeyPackage Client Capabilities, which lists which extensions are supported by the client.

bifurcation:
This PR adds a RequiredCapabilities extension that allows the members of a group to impose a requirement that new joiners support certain capabilities.  The extension is carried in the GroupContext and verified at Add time (including on external commits).

In the interim 2021-10-04, we briefly discussed whether this mechanism needed to be included in the core protocol.  Because of its interactions with Add and GroupContextExtensions proposals, it seems worth having in the main protocol.

Fixes #485

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2132,9 +2132,10 @@ The creator of a group MUST take the following steps to initialize the group:
   * Group ID: A value set by the creator
   * Epoch: 0
   * Tree hash: The root hash of the above ratchet tree
-  * Confirmed transcript hash: the zero-length octet string
-  * Interim transcript hash: the zero-length octet string
-  * Init secret: a fresh random value of size `KDF.Nh`
+  * Confirmed transcript hash: The zero-length octet string
+  * Interim transcript hash: The zero-length octet string
+  * Init secret: A fresh random value of size `KDF.Nh`
+  * Extensions: Any values of the creator's choosing
 
 * For each member, construct an Add proposal from the KeyPackage for that
   member (see {{add}})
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2158,6 +2159,28 @@ introduced.  The only choices the creator makes here are its own
...
+
+~~~~~
+struct {
+    ExtensionType extensions<0..255>;
+    ProposalType proposals<0..255>;
+} RequiredCapabilities;
+~~~~~
+
+This extension lists the extensions and proposal types that must be supported by
+all members of the group.  For new members, it is enforced during the
```
raphaelrobert, line 2177:
```suggestion
all members of the group.  For new members, it is enforced by existing members during the
```

PR #490, opened 2021-10-04
Use cascaded KDF instead of concatenation to consolidate PSKs

bifurcation:
In #483, @kkohbrok noted that the concatenation approach used to combine multiple PSKs is not well validated.  At the 2021-10-04 interim, we discussed moving from this approach to a "cascaded HKDF" approach, with the idea of doing the same thing here as the main key schedule.  This PR implements that suggestion.

One nice implication of the algorithm here is that when there are no PSKs, it produces the zero vector as a degenerate case.  So you could implement it pretty cleanly:

```
psk_secret = 0
for psk in psks:
    psk_secret = kdf.extract(psk_secret, extract_expand_psk(psk))

# Inject psk_secret into key schedule
```

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1602,9 +1602,10 @@ psk_secret_[i] = KDF.Extract(psk_input[i-1], psk_secret_[i-1])
 psk_secret     = psk_secret[n]
 ~~~~~
 
-The `index` field in `PSKLabel` corresponds to the index of the PSK in the `psk`
-array, while the `count` field contains the total number of PSKs.  In other
-words, the PSKs are chained together with KDF.Extract invocations, as follows:
+Here `0` represents the all-zero vector of length KDF.Nh. The `index` field in
+`PSKLabel` corresponds to the index of the PSK in the `psk` array, while the
+`count` field contains the total number of PSKs.  In other words, the PSKs are
+chained together with KDF.Extract invocations, as follows:
 
 ~~~~~
                 0                                   0       = psk_secret_[0]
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1594,13 +1594,34 @@ struct {
     uint16 count;
 } PSKLabel;
 
-psk_input_[i] = KDF.Extract(0, psk_[i])
-psk_secret_[i] = ExpandWithLabel(psk_input_[i], "derived psk", PSKLabel, KDF.Nh)
-psk_secret     = psk_secret_[0] || ... || psk_secret_[n-1]
+psk_extracted_[i] = KDF.Extract(0, psk_[i])
+psk_input_[i] = ExpandWithLabel(psk_extracted_[i], "derived psk", PSKLabel, KDF.Nh)
```
kkohbrok, line 1598:
```suggestion
psk_input_[i] = ExpandWithLabel(psk_extracted_[i], sprintf("derived psk %i", i), PSKLabel, KDF.Nh)
```
To err on the side of caution, I'd like to bake the structure of the cascade into the keys themselves. This ensures that you can't "extend" a cascade by taking the result of a cascade and use it to start a new one. Not something that would happen in the wild, I think, but it makes proofs easier and doesn't cost anything. The simple implementation you suggest would still work, you just have to add a counter.

(Not sure if the suggested line above is probably not what we want in the spec, but I think you understand what I mean.)

kkohbrok (changes requested):
Looks good with a small proposed addition.

draft-ietf-mls-protocol.md
kkohbrok: Ah, you are right, of course.

draft-ietf-mls-protocol.md
```diff
@@ -1623,6 +1624,9 @@ psk_[1] --> KDF.Extract --> ExpandWithLabel --> KDF.Extract = psk_secret_[1]
 psk_[n] --> KDF.Extract --> ExpandWithLabel --> KDF.Extract = psk_secret_[n]
 ~~~~~
 
+In particular, if there are no PreSharedKey proposals in a given Commit, then
+the resulting `psk_secret` is `psk_secret_[0]`, the all-zero vector.
```
kkohbrok, line 1628: ... of length `KDF.n`. (?)

bifurcation: We specify the length just a couple of paragraphs above ("Here `0` represents..."), so I thought it was clear enough to leave it out here.

PR #491, opened 2021-10-05
Use identity and endpoint_id to index clients in message structs

### LINKED ISSUE #486: Use `(identity, endpoint_id)` tuple instead of leaf indices in Message structs
As suggested in [here](https://github.com/mlswg/mls-protocol/pull/476#issuecomment-919746483), once #476 is merged, we can use `(identity, endpoint_id)` to uniquely address group members instead of leaf indices.

This slightly bloats messages, but it also helps abstract away the binary tree representation a little more.

kkohbrok:
This PR changes the message structs such that clients are now represented by the following struct instead of leaf indices:

```
struct {
    opaque sender_identity;
    opaque sender_endpoint_id;
} Client
```

I called the struct `Client`, as opposed to `Member` or `GroupMember`, since there are places where group membership first needs to be verified. Let me know what your thoughts are! 

Fixes #486.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1849,8 +1849,13 @@ enum {
 } SenderType;
 
 struct {
+    opaque sender_identity;
+    opaque sender_endpoint_id;
+} Client
+
+struct {
     SenderType sender_type;
-    uint32 sender;
+    Client sender;
 } Sender;
 
 struct {
@@ -2069,7 +2074,7 @@ encrypted, the sender data is encoded as an object of the following form:
 
 ~~~~~
 struct {
-    uint32 sender;
+    Client sender;
     uint32 generation;
     opaque reuse_guard[4];
 } MLSSenderData;
@@ -2105,10 +2110,9 @@ struct {
 } MLSSenderDataAAD;
 ~~~~~
 
-When parsing a SenderData struct as part of message decryption, the
-recipient MUST verify that the sender field represents an occupied
-leaf in the ratchet tree.  In particular, the sender index value
-MUST be less than the number of leaves in the tree.
+When parsing a SenderData struct as part of message decryption, the recipient
+MUST verify that the `Client` populating the `sender` field is a member of the
+group.
 
 # Group Creation
 
@@ -2282,20 +2286,20 @@ A member of the group applies an Update message by taking the following steps:
 
 ### Remove
 
-A Remove proposal requests that the client at a specified index in the tree be
-removed from the group.
+A Remove proposal requests that the `Client` `removed` be removed from the
+group.
 
 ~~~~~
 struct {
-    uint32 removed;
+    Client removed;
 } Remove;
 ~~~~~
 
 A member of the group applies a Remove message by taking the following steps:
 
-* Replace the leaf node at position `removed` with a blank node
+* Replace the leaf node of the group member `removed` with a blank node
 
-* Blank the intermediate nodes along the path from the removed leaf to the root
+* Blank the intermediate nodes along the path from that leaf to the root
 
 * Truncate the tree by reducing the size of tree until the rightmost non-blank leaf node
 
@@ -2371,7 +2375,7 @@ included in Commit messages.
 
 ~~~~~
 struct {
-    uint32 sender;
+    Client sender;
     uint32 first_generation;
     uint32 last_generation;
 } MessageRange;
@@ -2806,7 +2810,7 @@ struct {
     Extension group_context_extensions<0..2^32-1>;
     Extension other_extensions<0..2^32-1>;
     HPKEPublicKey external_pub;
-    uint32 signer_index;
+    Client signer;
     opaque signature<0..2^16-1>;
 } PublicGroupState;
 ~~~
@@ -2815,9 +2819,9 @@ Note that the `tree_hash` field is used the same way as in the Welcome message.
 The full tree can be included via the `ratchet_tree` extension
 {{ratchet-tree-extension}}.
 
-The signature MUST verify using the public key taken from the credential in the
-leaf node at position `signer_index`.  The signature covers the following
-structure, comprising all the fields in the PublicGroupState above `signer_index`:
+The signature MUST verify using the public key taken from the credential of the
+`signer`. The signature covers the following structure, comprising all the
+fields in the PublicGroupState above `signer`:
 
 ~~~~~
 struct {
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1848,9 +1848,14 @@ enum {
     (255)
 } SenderType;
 
+struct {
+    opaque sender_identity;
+    opaque sender_endpoint_id;
+} Client
```
bifurcation, line 1854: The bloat induced by this makes me pretty sad, and relying on just these two fields seems a little brittle.

As an alternative, what about using a hash of the leaf KeyPackage?:
* We only care about exact equality of these things, so we could use a hash of these values.  That also has the benefit of revealing less information on the wire.
* We only care about populated leaves, which contain a KeyPackage
* We already need `KeyPackage.hash()` for purposes of building/processing Welcome messages

That would get you down to the more compact:

```
uint8 KeyPackageID<0..255>;
```

I would be slightly happier if it were a fixed-length value and shorter than a hash output.  It's tempting to propose truncating the hash down to 128 bits, which seems unlikely to result in collisions in any practical scenario.  If it is fixed-length, you could have a nice compact value:

```
uint8 KeyPackageID[16];
```

draft-ietf-mls-protocol.md
```diff
@@ -1848,9 +1848,14 @@ enum {
...
+
 struct {
     SenderType sender_type;
-    uint32 sender;
+    Client sender;
```
bifurcation, line 1858: I don't think this quite works.  Remember that we have multiple sender types to accommodate: `member`, `preconfigured`, and `new_member`.  This covers the `member` case.  We should probably have a switch/variant struct here (probably always should have).

```
struct {
    SenderType sender_type;
    switch (sender_type) {
        case member:        Client client;
        case preconfigured: opaque key_id<0..255>;
        case new_member:    struct{};
    }
} Sender;
```

draft-ietf-mls-protocol.md
```diff
@@ -2105,10 +2110,9 @@ struct {
 } MLSSenderDataAAD;
 ~~~~~
 
-When parsing a SenderData struct as part of message decryption, the
-recipient MUST verify that the sender field represents an occupied
-leaf in the ratchet tree.  In particular, the sender index value
-MUST be less than the number of leaves in the tree.
+When parsing a SenderData struct as part of message decryption, the recipient
+MUST verify that the `Client` populating the `sender` field is a member of the
```
bifurcation, line 2114:
```suggestion
MUST verify that the `Client` indicated in the `sender` field is a member of the
```

draft-ietf-mls-protocol.md
bifurcation: Note that you can still use `Client` in MLSSenderData, because that always corresponds to the `member` case.

draft-ietf-mls-protocol.md
```diff
@@ -2815,9 +2819,9 @@ Note that the `tree_hash` field is used the same way as in the Welcome message.
 The full tree can be included via the `ratchet_tree` extension
 {{ratchet-tree-extension}}.
 
-The signature MUST verify using the public key taken from the credential in the
-leaf node at position `signer_index`.  The signature covers the following
-structure, comprising all the fields in the PublicGroupState above `signer_index`:
+The signature MUST verify using the public key taken from the credential of the
+`signer`. The signature covers the following structure, comprising all the
+fields in the PublicGroupState above `signer`:
```
bifurcation, line 2824: I actually think the old wording is clearer.  In general, I would still talk about these `Client` values as indices, since you need to look up leaves in the tree to get credentials, signature keys, etc.

draft-ietf-mls-protocol.md
bifurcation: Note also that if we make a smaller-than-hash-KeyPackage-ID thing here, we should also use it in Welcome.

kkohbrok: I like the idea, although I'm always a bit skeptical when it comes to shortening hashes. In this case it doesn't serve much of a security purpose, but I'd still prefer the full hash.

kkohbrok: Any suggestions for a name, though? The KeyPackage hash does not really represent the client anymore, but rather the client's presence in this particular group.

draft-ietf-mls-protocol.md
kkohbrok: Good point! I'll change it as suggested.

draft-ietf-mls-protocol.md
bifurcation: I kind of like `KeyPackageID`, as above.

draft-ietf-mls-protocol.md
```diff
@@ -1848,11 +1848,18 @@ enum {
     (255)
 } SenderType;
 
+struct {
+    uint8 KeyPackageID<0..255>;
+} MemberId
```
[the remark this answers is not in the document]
kkohbrok: Sounds like a good approach. I'll also note, that while we use a KeyPackage hash to identify members, these hashes change when the key package is updated, whereas the identity of the member does not. This could lead to errors, for example, when processing a `remove` right after an `update` for the same member was processed. I'm not entirely sure that's actually possible, but maybe you understand my concern. I don't want people to thing that KeyPackage hashes are static mappings to members.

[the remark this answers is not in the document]
bifurcation: Good point.  And to make it worse, we have the following:

> Update proposals are applied first, followed by Remove proposals

Maybe we should just declare that a single Commit containing an Update and a Remove for the same leaf is invalid.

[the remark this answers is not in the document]
kkohbrok: That would make sense, since updating and then immediately removing is rather pointless. Alternatively, we could make the KeyPackageID relative to the tree of the last epoch.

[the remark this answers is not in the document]
kkohbrok: > In any case, I don't think this extra layer of `struct` wrapping is needed.

The advantage of keeping the struct layer is that the type is more expressive. If we don't have the struct, we end up with, for example, 

```
~~~~~
struct {
     opaque removed<0..255>;
} Remove;
~~~~~
```
without any indication what `removed` actually is.

bifurcation (approved):
Thanks for the work on this @kkohbrok.  Looks good to me!

draft-ietf-mls-protocol.md
```diff
@@ -2983,9 +3012,9 @@ welcome_key = KDF.Expand(welcome_secret, "key", AEAD.Nk)
       public key in the node.
 
     * If the `path_secret` value is set in the GroupSecrets object: Identify the
-      lowest common ancestor of the leaves at `index` and at
-      `GroupInfo.signer_index`.  Set the private key for this node to the
-      private key derived from the `path_secret`.
+      lowest common ancestor of the leaf node index `index` and at the leaf
```
raphaelrobert, line 3015: This re-introduces leaf node indices.

PR #494, opened 2021-10-07
Require PublicGroupStateTBS for external init

kkohbrok:
There is really no benefit in requiring the whole PublicGroupState, inclucing signer and signature. Instead, this complicates things, since upon receiving a commit with an external init, the client has to fetch the PGS that was used for the external init, since they need it to compute the external init secret.

From what I can see, the only thing we really gain from using the signed PGS is a degree of attribution, where upon an external init, every group member knows who authored the PGS that was used for the external init. Feel free to disagree, but I don't think that's a property we really need here.

This PR makes it such that the `PublicGroupStateTBS` is used to compute the external init secret instead of the `PublicGroupState`.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1493,7 +1493,7 @@ information provided in the PublicGroupState and an external Commit to initializ
 their copy of the key schedule for the new epoch.
 
 ~~~~~
-kem_output, context = SetupBaseS(external_pub, PublicGroupState)
+kem_output, context = SetupBaseS(external_pub, PublicGroupStateTBS)
 init_secret = context.export("MLS 1.0 external init secret", KDF.Nh)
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1493,15 +1493,15 @@ information provided in the PublicGroupState and an external Commit to initializ
 their copy of the key schedule for the new epoch.
 
 ~~~~~
-kem_output, context = SetupBaseS(external_pub, PublicGroupState)
+kem_output, context = SetupBaseS(external_pub, PublicGroupStateTBS)
```
bifurcation, line 1496:
```suggestion
kem_output, context = SetupBaseS(external_pub, "")
```

raphaelrobert (approved):

PR #501, opened 2021-10-21
Add ciphersuite field to `PublicGroupStateTBS`

### LINKED ISSUE #498: `Ciphersuite` should be part of `PublicGroupStateTbs`
Am I missing something here, or should the ciphersuite really be part of the `PublicGroupStateTbs`? It's part of the `PublicGroupState`, but the way it is right now it doesn't get signed.

kkohbrok:
Fixes #498.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2965,6 +2965,7 @@ following structure, comprising all the fields in the PublicGroupState above
 
 ~~~~~
 struct {
+    ProtocolVersion version = mls10;
     CipherSuite cipher_suite;
     opaque group_id<0..255>;
     uint64 epoch;
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2965,6 +2965,7 @@ following structure, comprising all the fields in the PublicGroupState above
 
 ~~~~~
 struct {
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2965,6 +2965,7 @@ following structure, comprising all the fields in the PublicGroupState above
 
 ~~~~~
 struct {
+    CipherSuite cipher_suite;
```
raphaelrobert, line 2968:
```suggestion
    ProtocolVersion version = mls10;
    CipherSuite cipher_suite;
```

raphaelrobert (changes requested):
Good point! The protocol version was also missing. New joiners learn this from the `Welcome` message normally, that's why it wasn't included in the `GroupInfo` again.

The protocol version should also be repeated in the `PublicGroupState`.

draft-ietf-mls-protocol.md
kkohbrok: Can't the ProtocolVersion be inferred from the Ciphersuite?

bifurcation: No.  That's why they're separate fields in KeyPackage.

bifurcation (approved):

PR #507, opened 2021-11-15
Getting rid of the array representation

franziskuskiefer:
Filing this on behalf of @TWal after discussions with @kkohbrok and @karthikbhargavan.

The idea behind it is that the array representation is only one possible implementation for the tree. This PR removes the reliance on nodes indices to address nodes in the tree.

I'll leave it to @TWal to write a more detailed description.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1733,13 +1733,9 @@ in the ratchet.
 
 ~~~~
 DeriveTreeSecret(Secret, Label, Generation, Length) =
-    ExpandWithLabel(Secret, Label, RatchetContext, Length)
+    ExpandWithLabel(Secret, Label, Generation, Length)
 
-Where RatchetContext is specified as:
-
-struct {
-    uint32 generation = Generation;
-} RatchetContext;
+Where Generation is encoded as a uint32.
 ~~~~
 ~~~~~
 ratchet_secret_[N]_[j]
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1774,16 +1774,16 @@ forward secrecy for past messages. Members MAY keep unconsumed values around
 for some reasonable amount of time to handle out-of-order message delivery.
 
 For example, suppose a group member encrypts or (successfully) decrypts an
-application message using the j-th key and nonce in the ratchet of node
-N in some epoch n. Then, for that member, at least the following
+application message using the j-th key and nonce in the ratchet of leaf node
+L in some epoch n. Then, for that member, at least the following
 values have been consumed and MUST be deleted:
 
 * the `commit_secret`, `joiner_secret`, `epoch_secret`, `encryption_secret` of
   that epoch n as well as the `init_secret` of the previous epoch n-1,
 * all node secrets in the Secret Tree on the path from the root to the leaf with
-  node N,
-* the first j secrets in the application data ratchet of node N and
-* `application_ratchet_nonce_[N]_[j]` and `application_ratchet_key_[N]_[j]`.
+  node L,
+* the first j secrets in the application data ratchet of node L and
+* `application_ratchet_nonce_[L]_[j]` and `application_ratchet_key_[L]_[j]`.
 
 Concretely, suppose we have the following Secret Tree and ratchet for
 participant D:
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -648,43 +648,23 @@ For example, in the below tree:
 * The copath of C is (D, AB, EFG)
 
 ~~~~~
-              7 = root
+           ABCDEFG = root
         ______|______
        /             \
-      3              11
+     ABCD            EFG
     __|__           __|
    /     \         /   \
-  1       5       9     |
+  AB     CD       EF    |
  / \     / \     / \    |
 A   B   C   D   E   F   G
 
-                    1 1 1
-0 1 2 3 4 5 6 7 8 9 0 1 2
+0   1   2   3   4   5   6
 ~~~~~
 
-Each node in the tree is assigned an _index_, starting at zero and
-running from left to right.  A node is a leaf node if and only if it
-has an even index.  The node indices for the nodes in the above tree
-are as follows:
-
-* 0 = A
-* 1 = AB
-* 2 = B
-* 3 = ABCD
-* 4 = C
-* 5 = CD
-* 6 = D
-* 7 = ABCDEFG
-* 8 = E
-* 9 = EF
-* 10 = F
-* 11 = EFG
-* 12 = G
-
 A tree with `n` leaves has `2*n - 1` nodes.  For example, the above tree has 7
-leaves (A, B, C, D, E, F, G) and 13 nodes.  The root of a tree with `n` leaves
-is always the node with index `2^k - 1`, where `k` is the largest number such
-that `2^k < n`.
+leaves (A, B, C, D, E, F, G) and 13 nodes.
+
+Leaves are given an index, starting at 0 from the left to n-1 at the right.
 
 ## Ratchet Tree Nodes {#resolution-example}
 
@@ -700,7 +680,7 @@ Each node in a ratchet tree contains up to five values:
 
 * A private key (only within the member's direct path, see below)
 * A public key
-* An ordered list of node indices for "unmerged" leaves (see
+* An ordered list of leaf indices for "unmerged" leaves (see
   {{views}})
 * A credential (only for leaf nodes)
 * A hash of certain information about the node's parent, as of the last time the
@@ -729,18 +709,18 @@ brackets:
       _
     __|__
    /     \
-  _       5[C]
+  _      CD[C]
  / \     / \
 A   _   C   D
 
-0 1 2 3 4 5 6
+0   1   2   3
 ~~~~~
 
 In this tree, we can see all of the above rules in play:
 
-* The resolution of node 5 is the list \[CD, C\]
-* The resolution of node 2 is the empty list \[\]
-* The resolution of node 3 is the list \[A, CD, C\]
+* The resolution of node CD is the list \[CD, C\]
+* The resolution of leaf 1 is the empty list \[\]
+* The resolution of root node is the list \[A, CD, C\]
 
 Every node, regardless of whether the node is blank or populated, has
 a corresponding _hash_ that summarizes the contents of the subtree
@@ -1286,7 +1266,7 @@ might include a Key Package depending on whether or not it is blank.
 
 ~~~~~
 struct {
-    uint32 node_index;
+    uint32 leaf_index;
     optional<KeyPackage> key_package;
 } LeafNodeHashInput;
 ~~~~~
@@ -1303,7 +1283,6 @@ struct {
 } ParentNode;
 
 struct {
-    uint32 node_index;
     optional<ParentNode> parent_node;
     opaque left_hash<0..255>;
     opaque right_hash<0..255>;
@@ -1692,37 +1671,20 @@ For the generation of encryption keys and nonces, the key schedule begins with
 the `encryption_secret` at the root and derives a tree of secrets with the same
 structure as the group's ratchet tree. Each leaf in the Secret Tree is
 associated with the same group member as the corresponding leaf in the ratchet
-tree. Nodes are also assigned an index according to their position in the array
-representation of the tree (described in {{tree-math}}). If N is a node index in
-the Secret Tree then left(N) and right(N) denote the children of N (if they
-exist).
+tree.  If N is a node in the Secret Tree then left(N) and right(N) denote the
+children of N (if they exist).
 
-The secret of any other node in the tree is derived from its parent's secret
-using a call to DeriveTreeSecret:
-
-~~~~
-DeriveTreeSecret(Secret, Label, Node, Generation, Length) =
-    ExpandWithLabel(Secret, Label, TreeContext, Length)
-
-Where TreeContext is specified as:
-
-struct {
-    uint32 node = Node;
-    uint32 generation = Generation;
-} TreeContext;
-~~~~
-
-If N is a node index in the Secret Tree then the secrets of the children
+If N is a node in the Secret Tree then the secrets of the children
 of N are defined to be:
 
 ~~~~
 tree_node_[N]_secret
         |
         |
-        +--> DeriveTreeSecret(., "tree", left(N), 0, KDF.Nh)
+        +--> ExpandWithLabel(., "tree", "left", KDF.Nh)
         |    = tree_node_[left(N)]_secret
         |
-        +--> DeriveTreeSecret(., "tree", right(N), 0, KDF.Nh)
+        +--> ExpandWithLabel(., "tree", "right", KDF.Nh)
              = tree_node_[right(N)]_secret
 ~~~~
 
@@ -1734,10 +1696,10 @@ described in {{encryption-keys}}. The root of each ratchet is computed as:
 tree_node_[N]_secret
         |
         |
-        +--> DeriveTreeSecret(., "handshake", N, 0, KDF.Nh)
+        +--> ExpandWithLabel(., "handshake", "", KDF.Nh)
         |    = handshake_ratchet_secret_[N]_[0]
         |
-        +--> DeriveTreeSecret(., "application", N, 0, KDF.Nh)
+        +--> ExpandWithLabel(., "application", "", KDF.Nh)
              = application_ratchet_secret_[N]_[0]
 ~~~~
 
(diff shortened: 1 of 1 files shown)


TWal:
Thank you Franziskus!

The array representation is used in three places:
- in the tree hash (section 7.6),
- in the secret tree (section 8.3) and in the ratchet (section 8.4),
- in the ratchet_tree extension (section 11.3).

In the tree hash, we remove the node index in `ParentNodeTreeHashInput`, but keep a leaf index in `LeafNodeHashInput`. The leaf index is actually not needed if we only use the tree hash of the root, but if the protocol evolves later to use tree hashes of subtrees, it is actually important since they give information about where the subtree is located in the whole tree.

In the secret tree, we replace the node indices in the KDF by some "left" and "right" information which are sufficient.

For the ratchet_tree extension, we state that the order in which the nodes are serialized is the in-order traversal of the tree.

This allows us to remove the whole tree-math section.

draft-ietf-mls-protocol.md
```diff
@@ -3203,6 +3170,9 @@ struct {
 optional<Node> ratchet_tree<1..2^32-1>;
 ~~~~~
 
+The nodes are listed in the in-order trasversal of the ratchet tree: each node
+is listed between its left subtree and its right subtree.
```
bifurcation, line 3174: We probably need to provide more elaboration here to be explicit about the relations, and that probably requires re-importing a little bit of the tree math.  Suggested:

The nodes are listed in the order specified by a left-to-right in-order traversal of the rachet tree.  Each node is listed between its left subtree and its right subtree.  The leaves of the tree are stored in even-numbered entries in the array (the leaf with index `L` in array position `2*L`).  The root node of the tree is at position `2^k - 1` of the array, where `k` is the largest number such that `2^k` is smaller than the length of the array.  Intermediate parent nodes can be identified by performing the same calculation to the subarrays to the left and right of the root, following something like the following algorithm:

```
# Assuming a class Node that has left and right members
def subtree_root(nodes):
    # If there is only one node in the array return it
    if len(nodes) == 1:
        return Node(nodes[0])

    # Otherwise, the length of the array MUST be odd
    if len(nodes) % 2 == 0:
        raise Exception("Malformed node array {}", len(nodes))
    
    # Identify the root of the subtree
    k = 1
    while (1 << k) < len(nodes):
       k += 1
    R = (1 << (k-1)) - 1
    root = Node(nodes[R])
    root.left = subtree_root(nodes[:R])
    root.right = subtree_root(nodes[(R+1):])
    return root
```

The example tree in {{tree-computation-terminology}} would be represented as an array of nodes in the following form, where `R` represents the "subtree root" for a given subarray of the node array:

```
              7
        ______|______
       /             \
      3              11
    __|__           __|
   /     \         /   \
  1       5       9     |
 / \     / \     / \    |
A   B   C   D   E   F   G

                    1 1 1
0 1 2 3 4 5 6 7 8 9 0 1 2
<-----------> P <------->
<---> P <--->   <---> P -
- P -   - P -   - P -
```

draft-ietf-mls-protocol.md
```diff
@@ -3860,171 +3830,3 @@ MLS DE, that MLS DE SHOULD defer to the judgment of the other MLS DEs.
   thyla.van.der@merwe.tech
 
 --- back
-
-
-# Tree Math {#tree-math}
```
bifurcation, line 3865: I might keep this appendix, but change the emphasis:

# Array-Based Tree Representation

The ratchet tree extension defined in {{ratchet-tree-extension}} provides a way to transport a ratchet tree as an array of nodes, such that relationships among the nodes correspond to certain numerical relationships between the indices of the nodes in the array.  These numerical relationship can be used to enable an array-based in-memory representation (in addition to the on-the-wire representation) or to facilitate decoding of the wire-format array into another in-memory representation.  In this appendix, we provide provide Python example code for computing several common relationships among nodes, for example mapping between parents and left/right children.

bifurcation (commented):
I can understand the impulse here, and I think this is probably OK, especially for the ratchet tree parts.  The main concern I have is whether this leaves us with something clearly enough specified that the secret tree will work interoperably.  Maybe @psyoptix could opine here?

draft-ietf-mls-protocol.md
```diff
@@ -1734,10 +1696,10 @@ described in {{encryption-keys}}. The root of each ratchet is computed as:
 tree_node_[N]_secret
         |
         |
-        +--> DeriveTreeSecret(., "handshake", N, 0, KDF.Nh)
+        +--> ExpandWithLabel(., "handshake", "", KDF.Nh)
         |    = handshake_ratchet_secret_[N]_[0]
         |
-        +--> DeriveTreeSecret(., "application", N, 0, KDF.Nh)
+        +--> ExpandWithLabel(., "application", "", KDF.Nh)
```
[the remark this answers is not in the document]
TWal: That's right!

draft-ietf-mls-protocol.md
```diff
@@ -1810,14 +1779,14 @@ for some reasonable amount of time to handle out-of-order message delivery.
...
-index N in some epoch n. Then, for that member, at least the following
+N in some epoch n. Then, for that member, at least the following
 values have been consumed and MUST be deleted:
 
 * the `commit_secret`, `joiner_secret`, `epoch_secret`, `encryption_secret` of
   that epoch n as well as the `init_secret` of the previous epoch n-1,
 * all node secrets in the Secret Tree on the path from the root to the leaf with
-  node index N,
-* the first j secrets in the application data ratchet of node index N and
+  node N,
```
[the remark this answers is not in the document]
TWal: Fixed it in a new commit.

draft-ietf-mls-protocol.md
TWal: It looks good, however I think the `R` (that "represents the subtree root") and the `P` in the drawing should be the same letter?

TWal: Also I suggest the following change to the code, so that the `k` in the code and the `k` in the text above it are the same:
```
    k = 0
    while (1 << (k+1)) < len(nodes):
       k += 1
    R = (1 << k) - 1

File: draft-ietf-mls-protocol.md (hunk 2 above)
Comment (TWal) [reply draft-ietf-mls-protocol.md:3865]: Keeping it with a different emphasis seems fine, however I think it would better be located in a separate "implementation recommendation" document and not in the "protocol" document?

Comment (bifurcation) [reply draft-ietf-mls-protocol.md:3865]: I can live with that.  @kkohbrok I think you were going to start an "implementation considerations" document?

**Review by bifurcation — APPROVED** [review]
@TWal If you could add the ratchet tree description I provided, I'm OK to merge this.

**Comment by bifurcation** [comment]
One thing that I noticed working on #513 -- On Remove we truncate the tree to remove unused leaves from the right side.  That algorithm is clear when we're talking about an array representation; just find the rightmost non-blank leaf and remove every node (leaf or not) to the right of that index.  It seems like we should redefine that algorithm with this change.

File: draft-ietf-mls-protocol.md [hunk 5]
```diff
@@ -648,43 +648,23 @@ For example, in the below tree:
 * The copath of C is (D, AB, EFG)
 
 ~~~~~
-              7 = root
+           ABCDEFG = root
```
bifurcation, line 651: I would consider using single letters, from the end of the alphabet, for these.

bifurcation:
Suggested text for truncation:

* Truncate the tree:
    * Remove any blank leaf nodes to the right of the rightmost non-blank leaf
    * Perform the following steps until all parent nodes have two children:
        * If an intermediate node has no children, remove it
        * If an intermediate node N has one child C, remove the node and place C into the same child role with regard to N.
        * If the root node R of the tree has only one child C, remove R and assign C as the root of the tree

For example, suppose we had a tree with six leaves, of which (0, 1, 2, 5) were populated, (3, 4) blank, and in this situation a Remove was issued for the member at leaf 5.  Then the Remove would result in the following changes to the tree, resulting in a tree with three leaves, with root node `V`:

~~~~~
                         X                                    X
                   ______|___                           ______|___
                  /          \       Blank             /          \  
Start:           V           |       path             V           |  
               __|__         |       =====>         __|__         |  
              /     \        |                     /     \        | 
             U       W       Y                    U       W       Y  
            / \     / \     / \                  / \     / \     / \ 
           A   B   C   _   _   F                A   B   C   _   _   F


                         _                                    _
Remove             ______|___                           ______|   
blank             /          \       No                /             
leaves           V           |       children         V              
=====>         __|__         |       =====>         __|__            
              /     \        |                     /     \          
             U       W       _                    U       W          
            / \     /                            / \     / \         
           A   B   C                            A   B   C   _        


                                                               
One                                                               
child                                One                             
root             V                   child            V              
=====>         __|__                 =====>         __|_             
              /     \                              /    \           
             U       W                            U     |            
            / \     /                            / \    |            
           A   B   C                            A   B   C            
~~~~~

Note that if a batch of Removes is being processed, the truncation operation can be done once after all paths have been blanked.

bifurcation:
We may need a similar description of how you extend the tree to the right on Add.  I think the algorithm to add a leaf node `N` would be:

* We say that the subtree rooted at a node is "full" if the number of leaves under it is a power of 2.
* Let X = the parent of the rightmost leaf.  While X is full and X is not the root, let X = parent(X)
* If X is the root, add a new root node Y with left child X and right child N
* If X is not the root (and thus X is not full):
    * Let L = left(X) and R = right(X)
    * Set right(X) to a new node with left child R and right child N
 
It might be better to put these algorithms up in the tree section, as opposed to down here in the protocol.

bifurcation:
It makes me a little sad that these are so much more complicated than "truncate the array" and "add two slots to the array".  I guess I can understand the desire to make clear what happens to the logical tree, but it seems like we're introducing a lot of complexity.

TWal:
The Add algorithm doesn't work on the following tree:

          V  
        __|_ 
       /    \
      U     |
     / \    |
    A   B   C

After the While, we would have X = V and take the "If X is the root" step, which is wrong because X is not full.

Here is a corrected Add algorithm:
- Let X = the parent of the rightmost leaf. While X is full and X is not the root, let X = parent(X)
- If X is not full:
  - Let L = left(X) and R = right(X)
  - Set right(X) to a new node with left child R and right child N
- If X is full (and thus X is the root): add a new root node Y with left child X and right child N

I'm wondering if these algorithms are really necessary. It doesn't hurt to include them as an example on how to do things, but I think we can do simpler explanations using the fact that there is a unique left-balanced tree shape for every number of leaves.

Informally, we could say:
- for Add: "add enough blank parent node to have a valid left-balanced tree with n+1 leaves"
- for Remove: "remove all leaves to the right of the right-most non-blank leaf, and remove parents nodes to have a valid left-balanced tree".

I think explaining what is the expected output of the procedure, rather than explaining one way to do the procedure is easier to understand.

> It makes me a little sad that these are so much more complicated than "truncate the array" and "add two slots to the array". I guess I can understand the desire to make clear what happens to the logical tree, but it seems like we're introducing a lot of complexity.

It looks like added complexity, but for a new reader trying to understand MLS, I think it is easier to understand changes on the tree explained this way than understanding what happens to the tree when we do modifications on the underlying array representation.

draft-ietf-mls-protocol.md
TWal: I added the description you provided, and additionally replaced `1 << k` by `2**k` as I think it is more readable.

draft-ietf-mls-protocol.md
TWal: That's indeed better like this, I commited it.

bifurcation:
@TWal 

> I'm wondering if these algorithms are really necessary. It doesn't hurt to include them as an example on how to do things, but I think we can do simpler explanations using the fact that there is a unique left-balanced tree shape for every number of leaves.

That defines the *structure* of the tree, but not the *content* of the leaves.  We need to do both, so we need the transformation algorithms.  (In principle, you could write it starting from the structure, mapping old tree to new, but it would be an even more invasive rewrite.)

bifurcation (changes requested):
Marking this as "Request changes" until we get algorithms in place for truncate and add.

@TWal - Maybe as a compromise, we could do something like the following:

1. Present brief truncate and add algorithms of roughly the following form:
    * Truncate or extend the list of leaf nodes
    * Construct the unique left-balanced binary tree on the leaves, where each nodes' value is the value of the corresponding node in the old tree.
2. Present the full algorithms in an appendix

The key thing here is that truncate only deletes nodes and add only adds blank nodes, so there's not a ton of complexity in specifying the content of the nodes.  (Basically, we take the content of nodes from the old tree and the structure from the new.)  The thing that makes me queasy here is "corresponding node", since we don't have a definition for that.  Suggestions welcome.

bifurcation:
OK, to build some intuition here, I tried implementing a non-array-based variant tree structure that maintained the right structure.  [Quick draft in Go](https://go.dev/play/p/d6MW1vlc0oa).

The bad news is: I'm pretty convinced that there's no intuitive way to just say "make the unique left-balanced tree with these contents".  At best, we would need some way to define "Corresponding nodes" so that things could be copied over, which ultimately just turns into reinventing node indices.

The good news is: The above Go code validates two algorithms for extending and truncating the tree by one node, and they're pretty simple.  The add algorithm is effectively what @TWal proposed; the remove algorithm just excises and patches around the rightmost leaf's parent node.

It seems like we need to either add the algorithms or give up and close the PR.

@TWal @franziskuskiefer -- preferences between those two options?

franziskuskiefer:
> It seems like we need to either add the algorithms or give up and close the PR.

I'm not entirely decided on this. I agree with @TWal that the array representation isn't very intuitive for new readers and it's an implementation detail that's not good for the readability of the spec.
On the other hand do the algorithms get a pretty unwieldy and I'm not sure if they're really easier to understand than the array representation. And with the algorithms in the spec they aren't much better than the array representation in telling implementations what to do.

In an implementation of MLS that doesn't use the array representation underneath it is pretty easy to compute leaf indices for the few places they are needed.  So maybe another option is to leave the array representation in and only take the changes from this PR that remove the node index.
If the spec makes clear that the array representation is only one way of implementing the tree we could get the best of both worlds, a simple description of the left balanced tree without telling everyone what to implement as well as algorithms that only require leaf indices that can be easily retrieved from other implementations.

TWal:
I propose the following definition for "correspond":

    A node n1 in a tree t1 corresponds to a node n2 in a tree t2
    iff. the index of their leftmost leaf is the same, and they have the same height

This definition seems simple enough, while exactly capturing what nodes "are the same" in t1 and t2. It introduces the concept of "height" which is not currently defined in MLS, but I think it's fine.
Also, we can easily see that with this definition, two nodes "correspond" iff. they have the same node index in the array representation.

In fact, we can have the following alternate definition:

    A node n1 in a tree t1 corresponds to a node n2 in a tree t2
    iff. they have the same position in the in-order traversal

This is somewhat equivalent to re-introducing node-indexes, but in any case I find this definition less easy to understand than the first one.

My preferences are the following:
1. Only give a declarative description for Add and Remove, introducing the previous definition, and maybe put the algorithms in the implementation document.
2. Give the declarative description in a non-formal way (e.g. without defining "corresponds") and say afterward in the appendix "okay, here is the real algorithm which is the ground truth about what you should do".
3. Franziskus' proposition.

I prefer the first option, because then the intuition of the readers would actually be what is the ground truth. People working with trees are happy because they can work with a simple description on trees, and people working with arrays are happy because the "corresponds" definition is equivalent to "n1 corresponds to n2 iff. they have the same node index".

Then I prefer the second option, because I still think the tree algorithms are easier to understand than the bit-tricks used in tree math. With this option, the readers could gain an intuition about what happens to the tree using the declarative description, then read the "ground truth" algorithms and convince themselves that they indeed do what their intuition said. This option is great for people working with trees, but is a bit disappointing for people working with arrays.

Then the third option keep things as they are now, but implementations are not forced to implement tree math if they want to use another tree representation. This option is great for people working with arrays, but is a disappointing for people working with trees.

I'd like to go with the first option, since it is the only one where tree people and array people are happy. What do you think? @franziskuskiefer @bifurcation

bifurcation:
I think @franziskuskiefer is on the right track here -- let's phrase the structure of the tree in terms of node indices but be really clear that nodes don't have to use an array implementation.  So we would do something like:

* Keep node indices
* Truncate algorithm (in {{remove}}):
    * Identify the rightmost non-blank leaf node
    * Remove all nodes from the tree with an index higher than the rightmost non-blank leaf
* Extend algorithm (in {{add}}):
    * Let `n` be the index of the rightmost leaf node
    * Add a blank parent node with index `n+1` and a blank leaf node with index `n+2`
    * As a result, node `n+1` will have node `n+2` as its right child, and its left child and parent will be defined by the left-balanced binary tree structure.

Now that we have the tree-wise algorithms, it actually seems like a fine idea to add them in an appendix, parallel to the tree math appendix.  Then we could have a note in the Ratchet Tree section that says "if you want to use node indices to compute relations, see that appendix for some helpful detail; if you want to use pointers, see the other one".  That would make it abundantly clear that indices don't prescribe implementation.  I'll write a PR for that.

franziskuskiefer:
Getting the link based description in is a great first step.
The other change that's uncontroversial I think is to make clear throughout the spec that there are two different (obvious) ways of implementing the tree.

Then I we are left with one more question
> Keep node indices

With the two changes mentioned above I don't have strong feelings on this. Either way is fine with me. Only working with leaf indices instead of node indices is the more generic description, but also creates some churn in the algorithms that is also security relevant.

bifurcation:
Closing this since we seem to be converging on #522.

PR #509, opened 2021-11-16
Be explicit that Credentials can attest to multiple identities

bifurcation:
The specification currently assumes that a Credential provides a singular identity for a client, when in reality that is not the case.  For example, X509Credential can provide a whole collection of identities, and it's up to the application to decide if its preferred identity is among them.  Just like in TLS, a server can send a cert with 100 domains, but the TLS client knows which one it's looking for.

This PR refactors the parts of the spec that refer to identity to be consistent with this idea.  The main impacts are to processing of Update and Remove-within-External-Commit, both cases in which you want the new KeyPackage to logically represent the same identity as the old one.

This PR also removes the `endpoint_id` field from KeyPackage.  Since we now reference KeyPackages by KeyPackageID, there's no longer a need for a separate identifier.  (Unless we want an identifier that's stable across epochs, but (a) it's not clear to me that that is needed in general and (b) it can easily be added in an extension, as in fact we have done in Webex.)

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1283,9 +1283,34 @@ The signatures used in this document are encoded as specified in {{!RFC8446}}.
 In particular, ECDSA signatures are DER-encoded and EdDSA signatures are defined
 as the concatenation of `r` and `s` as specified in {{?RFC8032}}.
 
-Note that each new credential that has not already been validated
-by the application MUST be validated against the Authentication
-Service.
+Each new credential that has not already been validated by the application MUST
+be validated against the Authentication Service.
+
+### Uniquely Identifying Clients
+
+MLS implementations will presumably provide applications with a way to request
+protocol operations with regard to other clients (e.g., removing clients).  Such
+functions will need to refer to the other clients using some identifier.  MLS
+clients have a few types of identifiers, with different operational properties.
+
+The Credentials presented by the clients in a group authenticate
+application-level identifiers for the clients.  These identifiers may not
+uniquely identify clients.  For example, if a user has multiple devices that are
+all present in an MLS group, then those devices' clients will all present the
+user's application-layer identifiers.
+
+Internally to the the protocol, group members are uniquely identified by their
+leaves, expressed as KeyPackageRef objects.  These identifiers are unstable:
+They change whenever the member sends a Commit, or whenever an Update
+proposal from the member is committed.
+
+MLS provides two unique client identifiers that are stable across epochs:
+
+* The index of a client among the leaves of the tree
+* The `epoch_id` field in the key package
+
+The application may also provide application-specific unique identifiers in the
+`extensions` field of KeyPackage object.
 
 # Key Packages
 
@@ -2507,7 +2532,7 @@ Note that these three steps may be done by the same group member or different
 members.  For example, if a group member sends a commit with an inline ReInit
 proposal (steps 1 and 2), but then goes offline, another group member may send
 the corresponding Welcome.  This flexibility avoids situations where a group
-gets stuck between steps 2 and 3. 
+gets stuck between steps 2 and 3.
 
 Resumption PSKs with usage `reinit` MUST NOT be used in other contexts.  A
 PreSharedKey proposal with type `resumption` and usage `reinit` MUST be
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1392,6 +1392,39 @@ modification of its content MUST be reflected by a change in its
 signature. This allows other members to verify the validity of the KeyPackage
 at any time, particularly in the case of a newcomer joining the group.
 
+## KeyPackage Validation
+
+The validity of a KeyPackage needs to be verified at a few stages:
+
+* When a KeyPackage is downloaded by a group member, before it is used
+  to add the client to the group
+* When a KeyPackage is received by a group member in an Add, Update, or Commit
+  message
+* When a client joining a group receives KeyPackages for the other members of
+  the group in the group's ratchet tree
+
+The client verifies the validity of a KeyPackage using the following steps:
+
+* Verify that the credential in the KeyPackage is valid according to the
+  authentication service and the client's local policy.
+
+* Verify that the signature on the KeyPackage is valid using the public key
+  in the KeyPackage's credential
+
+* Verify that the KeyPackage is compatible with the group's parameters.  The
+  ciphersuite and protocol version of the KeyPackage must match those in
+  use in the group.  If the GroupContext has a `required_capabilities`
+  extension, then the required extensions and proposals MUST be listed in
+  the KeyPackage's `capabilities` extension.
+
+* Verify that the following fields in the KeyPackage are unique among the
+  members of the group (including any other members added in the same
+  Commit):
+
+    * `credential.signature_key`
+    * `hpke_init_key`
+    * `endpoint_id`
+
 ## Client Capabilities
 
 The `capabilities` extension indicates what protocol versions, ciphersuites,
@@ -2615,8 +2648,8 @@ retrieved by hash (as a ProposalOrRef object) in a later Commit message.
 ### Add
 
 An Add proposal requests that a client with a specified KeyPackage be added
-to the group.  The proposer of the Add MUST validate the KeyPackage in the same
-way as receipients are required to do below.
+to the group.  The proposer of the Add MUST verify the validity of the
+KeyPackage, as specified in {{keypackage-validation}}.
 
 ~~~~~
 struct {
@@ -2631,23 +2664,7 @@ placed in the leftmost empty leaf in the tree, for the second Add, the next
 empty leaf to the right, etc. If no empty leaf exists, the tree is extended to
 the right.
 
-* Validate the KeyPackage:
-
-    * Verify that the signature on the KeyPackage is valid using the public key
-      in the KeyPackage's credential
-
-    * Verify that the following fields in the KeyPackage are unique among the
-      members of the group (including any other members added in the same
-      Commit):
-
-        * `credential.signature_key`
-        * `hpke_init_key`
-
-    * Verify that the KeyPackage is compatible with the group's parameters.  The
-      ciphersuite and protocol version of the KeyPackage must match those in
-      use in the group.  If the GroupContext has a `required_capabilities`
-      extension, then the required extensions and proposals MUST be listed in
-      the KeyPackage's `capabilities` extension.
+* Validate the KeyPackage as specified in {{keypackage-validation}}
 
 * Identify the leaf L for the new member: if there are empty leaves in the tree,
   L is the leftmost empty leaf.  Otherwise, the tree is extended to the right
@@ -2674,32 +2691,19 @@ struct {
 
 A member of the group applies an Update message by taking the following steps:
 
-* Validate the KeyPackage:
+* Validate the KeyPackage as specified in {{keypackage-validation}}
 
-    * Verify that the signature on the KeyPackage is valid using the public key
-      in the KeyPackage's credential
+* Verify that the following fields in the new KeyPackage are the same as the
+  one being replaced:
 
-    * Verify that the KeyPackage's credential is valid according to the
-      Authentication Service
+    * `version`
+    * `cipher_suite`
 
-    * Verify that the following fields in the KeyPackage are unique among the
-      members of the group (including any other members added in the same
-      Commit):
+* Verify that the `hpke_init_key` value is different from the corresponding
+  field in the KeyPackage being replaced.
 
-        * `credential.signature_key`
-        * `hpke_init_key`
-
-    * Verify that the following fields in the new KeyPackage are the same as the
-      one being replaced:
-
-        * `version`
-        * `cipher_suite`
-
-    * Verify that the `hpke_init_key` value is different from the corresponding
-      field in the KeyPackage being replaced.
-
-    * Verify that the set of identities attested by the credential is acceptable
-      to the application for the participant being updated.
+* Verify that the set of identities attested by the credential is acceptable
+  to the application for the participant being updated.
 
 * Replace the sender's leaf KeyPackage with the one contained in
   the Update proposal
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1495,7 +1495,9 @@ The validity of a KeyPackage needs to be verified at a few stages:
 The client verifies the validity of a KeyPackage using the following steps:
 
 * Verify that the credential in the KeyPackage is valid according to the
-  authentication service and the client's local policy.
+  authentication service and the client's local policy.  These actions MUST be
+  the same regardless of at what point in the protocol the KeyPackage is being
+  verified.
 
 * Verify that the signature on the KeyPackage is valid using the public key
   in the KeyPackage's credential
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2679,6 +2679,9 @@ A member of the group applies an Update message by taking the following steps:
     * Verify that the signature on the KeyPackage is valid using the public key
       in the KeyPackage's credential
 
+    * Verify that the KeyPackage's credential is valid according to the
+      Authentication Service
+
     * Verify that the following fields in the KeyPackage are unique among the
       members of the group (including any other members added in the same
       Commit):
@@ -3162,6 +3165,9 @@ A member of the group applies a Commit message by taking the following steps:
 * If the `path` value is populated: Process the `path` value using the
   provisional ratchet tree and GroupContext, to generate the new ratchet tree
   and the `commit_secret`:
+
+  * Verify that the KeyPackage is acceptable according to the rules for Update
+    (see {{update}})
 
   * Apply the UpdatePath to the tree, as described in
     {{synchronizing-views-of-the-tree}}, and store `leaf_key_package` at the
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1284,7 +1284,9 @@ In particular, ECDSA signatures are DER-encoded and EdDSA signatures are defined
 as the concatenation of `r` and `s` as specified in {{?RFC8032}}.
 
 Each new credential that has not already been validated by the application MUST
-be validated against the Authentication Service.
+be validated against the Authentication Service.  Applications SHOULD require
+that a client present the same set of identifiers throughout its presence in
+the group, even if its Credential is changed in a Commit or Update.
 
 ### Uniquely Identifying Clients
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1286,7 +1286,11 @@ as the concatenation of `r` and `s` as specified in {{?RFC8032}}.
 Each new credential that has not already been validated by the application MUST
 be validated against the Authentication Service.  Applications SHOULD require
 that a client present the same set of identifiers throughout its presence in
-the group, even if its Credential is changed in a Commit or Update.
+the group, even if its Credential is changed in a Commit or Update.  If an
+application allows clients to change identifiers over time, then each time the
+client presents a new credential, the application MUST verify that the set
+of identifiers in the credential is acceptable to the application for this
+client.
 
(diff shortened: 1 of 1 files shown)


bifurcation:
How appropriate that this is issue (X.)#509 !

kkohbrok:
The advantage of keeping an `endpoint_id` is that we can mandate in the spec that it can't change across epochs. While if we have a proprietary extension that contains an identifier, every implementation would have to be aware of that extension and enforce that it doesn't change across updates. So I'd be in favor of keeping it.

raphaelrobert:
This softens the requirements quite a bit, namely:
 - There is no more requirement for identities to be unique across the group, e.g. Alice can double-join the tree by occupying two leaves, simply by issuing an Update.
 -  An identity can completely change in the course of an Update if the application allows it. 

I'm worried the proposed solution can lead to UB, when clients try to remove a certain member that potentially occupies more than one leaf.

bifurcation:
@kkohbrok - We already have an identifier that is stable across epochs -- the leaf index!  We don't have a use for it now (as evidenced by the `KeyPackageID` PR), but it's still there.

@raphaelrobert - I don't think those concerns hold up.  There's no new risk of double-join; a client can only present an identity for which it can obtain a Credential that is acceptable to other clients.  If a client presents a legitimate identity twice, that's not harm; it might be multiple devices, which we can still uniquely reference by KeyPackageID.  If a client can get Credentials for identities they don't own, that's a bug in the authentication system, not MLS.

As far as identities completely changing -- one could consider that a feature, not a bug.  TLS has had to go to [a fair bit of trouble](https://datatracker.ietf.org/doc/draft-ietf-tls-exported-authenticator/) to allow servers to assert additional identities over the course of a connection.  It seems like we have an elegant solution to the problem here.  Applications can of course always be stricter.

kkohbrok:
It's stable across epochs until you remove the party in that leaf and add another one into it. 

The options that have been suggested are the following:
- leaf index: is not tied to any particular identity/key package
- key package (id/hash): potentially changes across epochs
- identity in the credential: not unique per group 

None of these variables/indices/identifiers really does the trick of uniquely and consistently (across epochs) identifying a client in a group. The endpoint id/identity tuple does all of the above. We don't have to use it in message structs, where the key package ids are just fine, as messages are tied to an epoch anyway. I'm just in favor of keeping the endpoint id so it can serve as a consistent identifier for group members for use by the application.

bifurcation:
@kkohbrok - I can live with re-adding `endpoint_id`.  I can imagine scenarios where it would be troublesome, but in the worst case, you just fill in some long random junk there.  I have re-added.

Any other issues, or are we good to merge?

kkohbrok:
I think I'm ok with this approach, but I don't think we should merge this without previous discussion either in a meeting or on the mailing list. Is there an interim planned before the WGLC?

bifurcation:
There is no interim planned.  I'll send a ping to the mailing list.

raphaelrobert:
I thought about this some more and I managed to convince myself that the `endpoint_id` is not strictly needed. Not having it however puts a higher burden on either the implementation or on the AS to track changes across epochs in order to ensure continuity. 
In practical terms, the only constant thing about member X is that it occupies leaf node i between the time it was added and removed (regardless of updates during that time). If however X gets removed and a new member Y gets added at the same leaf node, i now references Y and not X, which makes leaf node indices useless across epochs.

For me this means one of the following two things must be true:

 a) Implementations emulate the concept of an `endpoint_id`. They assign it to a member at the point in time when the member gets added and delete it when the member gets removed. In between the two events, this `endpoint_id` is used as the only unique identifier for a member.
 
 b) We put more load on the AS and mandate the AS must be able to correlate two KeyPackages from two different epochs to ascertain the two KeyPackages belong to the same user. For example, member X gets added in epoch A, and updates its KeyPackage (including its credential, HPKE init key, and signature key) in epoch B. The AS must then be able to tell in epoch C whether it's still the same member X or a new member Y, just by looking at the two KeyPackages.

bifurcation:
@raphaelrobert Yeah, I think I'm about in the same place as you.  As I think I've said before, to use MLS in Webex, we needed to add a KeyPackage extension for the application unique identifer anyway.  With `endpoint_id`, we could have just used that field.

But note that the PR has already been updated to re-add `endpoint_id`.  So are you good with the PR in its current state?

br-hale:
Zooming out on this problem, I will make a couple remarks about options for identifying parties. First, it seems that the above alternatives (and adding some not discussed) can generally be grouped into one of the following categories (for these, 'persistence' implies over the duration of the session/group, and 'directly authenticated' implies either as a credential or an identifier signed through use of a credential):

1. endpoint identifiers change across epochs and directly authenticated along with that change
2. endpoint identifiers are persistent across epochs and directly authenticated once
3. endpoint identifiers are not persistent across epochs, but directly authenticated only once 
4. endpoint identifiers are persistent across epochs, but unauthenticated

I note the above differentiators, since the role of endpoint identifiers begins to factor in when assessing E2EE. @chelseakomlo and I are working on some of the 'ends' aspect of E2EE and I think the current PR change could prove non-trivial with respect to justifying E2EE in MLS. Option 3 is what Signal does and, throughout the development of MLS, Signal as been used as a guidepost for 'at least equivalent' security. Naturally, if credentials are used directly, we meet Option 2 or 1, dependent on handling. Please correct me if I am wrong, but it appears that the current PR would lead us into Option 4.

I support keeping credentials, at least in the sense of keys. Actually, in this respect, it is a little confusing with regard the issue with x509. Since early on, "credential" in MLS has been used as denoting a signature key, not a certificate that may contain more. Maybe I am not quite seeing the issue as to why the application cannot handle interpreting the certificate as separate to the credential in the spec.

bifurcation:
@br-hale Strongly disagree with 'Since early on, "credential" in MLS has been used as denoting a signature key, not a certificate that may contain more.'  The notion of a Credential has always included at least an identifier as well as the signature key, in order to address the Unknown Key Share / Identity Misbinding attacks that occur in Signal and elsewhere.  If you put that together with the idea that Credentials are verifiable using some AS, you basically arrive at certificates.

In your taxonomy, only option (1) actually works for MLS, assuming that by "endpoint identifiers", you mean the authenticated ones, and not just the opaque `endpoint_id`, which just has to be unique.  Option (4) is clearly unacceptable, because the whole point of these identifiers is that they're authenticated by the AS.  Options (2) and (3) don't make sense because members can be added to the group and need to be able to authenticate the other members when they join (so "directly authenticate once" doesn't work).  Note that MLS is unlike Signal in that the membership is multi-party and dynamic, and that we seek to address UKS issues that Signal explicitly disclaims.

> @chelseakomlo and I are working on some of the 'ends' aspect of E2EE and I think the current PR change could prove non-trivial with respect to justifying E2EE in MLS.

Could you describe more what the issue is here?  Just because the endpoints can change the set of identities they express doesn't seem like it undermines the notion of E2EE.

br-hale:
@bifurcation I agree that option (1) is what is needed for MLS. The discussion above seems to advocate for a variant of option (4) though. This variant question also seems to be the crux of the matter @kkohbrok was pointing out above, so if something is being miscommunicated here it would help if you could clarify that, i.e. that we are at least on the same page regarding (1) being the goal. The 'Signal alignment' only holds as a baseline, i.e. if you are advocating for (4) then I would say that we can and should do better. The claim about options (2) and (3) is incorrect though, since direct authentication occurs at time of group join. 

To avoid diverging down side discussion points and to keep this PR decision on track, it is essential to establish clarity on the claim that MLS (with the current PR) would be able to offer option (1). Copying from @kkohbrok notes above, we have that (quote):
- leaf index: is not tied to any particular identity/key package
- key package (id/hash): potentially changes across epochs
- identity in the credential: not unique per group
(end quote)

and as @bifurcation you say, the opaque endpoint_id is by no means sufficient. By elimination, this does not leave many options. So what aspect are you referring ref. "the authenticated ones"? I.e. what persistent or rotated endpoint identifier is there which consistently links authenticity throughout the lifespan of the group? Anything that is not persistent is insufficient as is anything that is authenticated in a disconnected manner (e.g. can be replaced and authenticated with a new key that is not linked to the prior one).

bifurcation:
@br-hale What I mean by "the authenticated ones" is the identifier(s) in the Credential, in the KeyPackage, which are supposed to be authenticated with the AS.  To be specific: For BasicCredential, the `identity` field, and for X509Credential, identifiers in the `subject` or `subjectAltName` fields.

So yes, we have a number of identifiers for endpoints:

1. Identifiers in credentials (as above)
2. Leaf index: Unique within a group; constant for a member's lifetime in the group, by virtue of tree math
3. `endpoint_id`: Unique within a group; constant for a member's lifetime in the group, by fiat
4. KeyPackageID (==hash): Unique within a group; changes on Commit/Update by the member

The point of this PR is just to clarify that the identifiers authenticated in credentials are neither unique within a group nor constant across Commit/Update by the member.  (And to clarify what must remain constant across Commit/Update.)  Does that seem problematic to you?

br-hale:
The issue of credentials not being unique or constant is at the heart of this. Other items here are not authenticated or are reliant on those credentials and therefore the consistency of e.g. the leaf index does not actually imply consistency of identity. 

If credentials were not unique/constant in the sense that a prior one signs off on the next one, then we would inherit consistency from that (which is different from the case here). Similarly, if a single root identity ("identity") were to forward to other sub-identities, then there is consistency in the meaning of identity. 

What we have here is a backdoor on switching out identities which would be allowed by the spec - yes, they are subcomponents of a credential, but any sender is not actually assured of which identity they are conversing with. For example, Alice may trust a domain name / email address from Bob, but when that switches part-way through the conversation, there is no opt-out point or notification to Alice. There is considerable drift available to Bob (either unintentionally or maliciously) that could occur without knowledge from other group members. This is noticeably different from the case that Bob simply maliciously forwards messages externally, as the drift occurs internally to the group.

draft-ietf-mls-protocol.md
```diff
@@ -3051,10 +3068,11 @@ External Commits work like regular Commits, with a few differences:
...
   * There MUST be a single ExternalInit proposal
-  * There MUST NOT be any Add or Update proposals
-  * If a Remove proposal is present, then the `credential` and `endpoint_id` of
-    the removed leaf MUST be the same as the corresponding values in the Add
-    KeyPackage.
+  * There MUST NOT be any Update proposals
+  * If a Remove proposal is present, then the `credential` of the Add KeyPackage
+    MUST present a set of identifiers that is acceptable to the application for
+    the removed participant (as if this were an Update for that participant).
+    The Add KeyPackage and the removed leaf MUST have the same `endpoint_id`.
```
kkohbrok, line 3075: In other words, this gives us a new requirement for the AS, namely that it has to be able to tell us if a given credential is the valid successor of another credential. With fixed identities, this would not be necessary. Of course one way of implementing this check for the AS is to check if some notion of identity that arises from a given credential type is indeed the same for both certificates.

bifurcation: I wouldn't frame this as an AS requirement, but regardless, you have to do *something* here: The application has to either (1) verify that the "identity" is the same (for some definition thereof), or (2) verify that the identity is acceptable. 

(2) is the state of the art basically everywhere right now.  See, e.g., [RFC 6125](https://datatracker.ietf.org/doc/html/rfc6125).  Doing (1) for an X.509 certificate is actually pretty complex -- or even defining it -- given all the notions of "identity" lurking around in there.  For example, we could enumerate the subjectAltName, but then someone defines [a new way of embedding telephone numbers](https://datatracker.ietf.org/doc/html/rfc8226).  The only way to be sure is to require that the whole certificate be the same, but you probably also want to allow signature key rotation, so really you want the whole cert to be the same except for the public key.  This is not functionality provided by any X.509 library that I'm aware of.

I think the underlying confusion here is around what layers are responsible for what.  Compare to HTTPS:

* The application layer provides the identifier you're supposed to be authenticating (host name in the URL)
* TLS only provides a *set of identifiers* that are authenticated
* It's up the the application layer to verify that the authenticated identifiers are acceptable

(This is true even with extensions like [exported authenticators](https://datatracker.ietf.org/doc/draft-ietf-tls-exported-authenticator/) that allow the set of identifiers to change.) All this PR is proposing is a similar division of labor in the MLS case.

bifurcation:
You might think the simplest thing we could do here is just require that a Credential be stable for the lifetime of a member's presence in the group.  If you want to change the Credential, you have to be removed and re-added.  But (a) that seems like throwing out a bunch of useful functionality, and (b) it doesn't actually remove the risk, just moves it over to the remove/re-add case, making it harder for the application to track continuity.

bifurcation:
(Force-pushed to address failed merge)

tomleavy:
+1 to this change. I think this clarifies how to handle application layer identity, and that is particularly helpful when using X509 as the credential where "identifier" is a bit complicated.

br-hale:
@bifurcation shifted this discussion offline and it has been discussed extensively there with some of the issues more clearly outlined. To avoid copying it all into this PR, I suggest discussion at the next WG informal interim on 27 Jan., or as seen suitable by the authors.

bifurcation:
Thanks @br-hale for some offline discussion here to help me understand your concerns.

IIUC the core of your concern is that the spec doesn't clearly require that new credentials be validated with the AS whenever they appear.  I can see how if you believe credentials are only validated on Add, then you need everything to chain back to the Add, and changes are problematic.

I don't think the spec is actually broken in this way, since it says in `# Credentials` that "each new credential that has not already been validated by the application MUST be validated against the Authentication Service" (and has for 2 years according to `git blame`).  But I agree the spec should be clearer and more consistent on this point, and would be happy to make the edits required.

bifurcation:
Latest commit attempts to address @br-hale's concerns by explicitly requiring credential validation on Update/full-Commit.

br-hale:
Given the push to shift the key points of discussion here, I am copying them avoid paraphrasing confusions: 

[Responding to @bifurcation regarding the sufficiency of text that suggests developers determine whether identifiers should be constant based on whether they feel they have knowledge to make changes]

A recommendation for consistency of identifiers will likely be insufficient; I suspect that this decision point underpins being able to show this as an AKE at all. It would be like making certs optional and changeable during TLS’s handshake. What TLS does once, MLS does continuously (the handshake). It is not a ‘different’ handshake per epoch, but all links through all back to the initialization. Proving security of an AKE is in part conditioned on the identities (and commitment to those identities) of the entities involved (‘identity’ being generic, not a specific term in MLS here). So if that changes part way through the handshake (= session in MLS), then one of the key properties of an AKE is likely going to break down. Relying on the group key state to fill the consistency check gap that is created would be circular logic.
 
Some options are: 1) have a consistent identifier (regardless of whether that is a cert, endpoint_id, or other), 2) have a logically linked chain of identifiers (e.g. the last one signs the next as some type of cert chain if the identifier is taken to be a cert, or hash the last identifier to make the next in the chain, etc.), or 3) remove the member and add a ‘new’ member whenever the identifier changes. This really is not an optional piece of a protocol or something that can passably be left up to how the developer is feeling – we are really dealing with the fundamental question of whether MLS will be an AKE or not.
 
I do not really see an advantage with us being loosie and shifting MLS to a KE vs an AKE at this point. It may be worthwhile instead to clarify the use case so we can see what the need is and then we can see how that may be solved. For example, if you want multiple identities/credentials for an entity, then we could require that those all be present somehow at the start of the protocol. Note that switching among a pre-announced set of identities for a member or removing an old identifier if the next one is already in the set could potentially be alright – it is the introduction of some new identifier or credential that carries no explicit, persistent link from prior ones that is a problem. Unsurprisingly, the consequence of fixing side case uses is that the protocol gets more complex, but if it is meaningful for some development case then that is the reality.

br-hale:
Also:

[In response to @bifurcation regarding the assertion that MLS authenticates in each epoch with credentials and via AS, making it a fresh AKE per epoch.]

The analogy of TLS authenticators is misconstruing the spec, since that is a single session – you see, using alternative identifiers for any given session is not a problem, but switching them out mid-session is. While it is true that ratcheted key exchange is normally looked at as epoch-level mini-sessions as far as key establishment goes, the ‘A’ part stretches the length of the entire session and ties to the very start (‘Add’). This is intuitively noticeable in that the spec requires that the key package be validated and the credential verified only at an Add (pg 54) and not every epoch. The credential (issuer, keys, or identity) could effectively change without group members even verifying the new credential (and that is without touching on the legitimacy of the ‘chain’ between the two). That is why Updating/Committing cannot be regarded as a remove and Add: the required operations for each differ.
 
(example on E2EE identities changing maliciously given that the verification requirements at the Add step differ from at the Update step when the cert changes)
 
In short, no, the identifiers are not authenticated in each epoch, only at the start. Confidentiality ratchets per epoch, but authenticity does not. That said, if you can require that every time there is a change in a credential then all of the prescribed steps in an Add are also followed (with respect to verifying identities) then my concerns are addressed and I would also agree that we have at least authentication per change (although not per epoch).
 
 
 ----
 
As an aside note, the source of this challenge for MLS is that, unlike TLS, we do not prescribe credentials from a certifying authority. In TLS the certificate authority and therefore means of verifying authenticity of identities is very explicit – no matter when those are verified, a certificate authority will always be listed. Thus TLS can easily require checks. In MLS (section 6.2) we only say ‘MAY verify identities’ (MLS requires any credential not validated by members to be validated by the AS by an unspecified method). If I recall correctly this choice was made early on to capture the loophole for ‘trust on first use’ type applications. For that case the game changes entirely though since both the initial authentication and any subsequent authentication both suffer the same challenge of needing an OOB solution for authenticating identity (in other words, cert changes are no worse off than at the start).
However, for MLS using normal certificates, the fact that this loophole exists creates a challenge on how to ensure that the same authentication functionality is being used when certs change (the wording is simply harder to formulate while still allowing for the loophole). TLS in contrast is very explicit about certificate verification, but it can be because 1) there is only one point were this is done in the protocol as a single session and 2) there is really only one main method of handling cert authorities with x509, with explicit exceptions given by e.g. the export exemption you name.
 
In short, if you authenticate OOB the first time, OOB after certificate change is persistent. If you authenticate with cert authorities the first time, then changing the certificate should imply a re-validation of the new cert’s chain. What we cannot do is allow a functionality method that is not persistent (i.e. the current problem) where we verify once but allow changes later without equivalent oversight.

br-hale:
With the provided background context for the WG above, I will address the latest question from @bifurcation and the changes.

First - thank you for the edits to the PR. 

The proposed changes do not fully address the above however. The key gap is on what is being required to be done for verification of identity at the point of change. We need to explicitly enforce that the same process at the Add needs to be done at the change. Right now there is still a bailout option, where credentials can be verified directly at the Add, but are slipped into a background/AS option. This is mixing and matching.

bifurcation:
DIscussion on working call:

* Need to make sure we have consistent validation logic across Add/Update/Commit

bifurcation:
@br-hale The latest commit adds a consolidated "KeyPackage validation" section that describes the common rules that are applied across Add, Update, and Commit, including credential validation in particular.

The only difference between those cases now is that only Update and Commit have a requirement to check that the new set of identifiers is OK.  It seems like that check doesn't make sense for Add.

Let me know if that looks good to you!

draft-ietf-mls-protocol.md
```diff
@@ -1439,6 +1479,39 @@ modification of its content MUST be reflected by a change in its
...
+  to add the client to the group
+* When a KeyPackage is received by a group member in an Add, Update, or Commit
+  message
+* When a client joining a group receives KeyPackages for the other members of
+  the group in the group's ratchet tree
+
+The client verifies the validity of a KeyPackage using the following steps:
+
+* Verify that the credential in the KeyPackage is valid according to the
+  authentication service and the client's local policy.
```
br-hale, line 1496: I suggest adding:
The authentication service validity check steps and client's local policy MUST be the same for all verification stages (KeyPackage download, Add, Update, Commit, or KeyPackage receipt).

br-hale: @bifurcation the updated changes look very good. I made a suggestion above. With that addition (or equivalent), it looks good to me.

kkohbrok: I would suggest do go slightly further and require that on top of the verification you do in case of an Add (i.e. asking the AS "Is this credential Ok?"), the AS should also answer the question "Is this credential OK as the successor to this previous credential?".

bifurcation: Thanks @br-hale.  @kkohbrok I think that's covered down in the Update section.

PR #510, opened 2021-11-16
Use smaller, fixed-size hash-based identifiers

### LINKED ISSUE #504: KeyPackage ID vs KeyPackage Identifier
There's a KeyPackageID ( 7.1. Key Package IDs ) and a KeyPackage Identifiers ( 7.4. KeyPackage Identifiers ) right now.
This is pretty confusing and probably not necessary. 

As far as I can tell the only difference is that the KeyPackageID has to be unique within a group. If we add this restriction to the KeyPacakge Identifier, for example by adding the key package hash, we could get rid of the KeyPackageID and the confusion.

### LINKED ISSUE #506: Use smaller, fixed-size hash-based identifiers
Right now, we use hash-based identifiers to reference KeyPackages (e.g., in KeyPackageID) and Proposals (ProposalOrRef).  To avoid truncating hashes, we use full-sized hashes, which are (a) much larger than necessary and (b) vary in size depend on the ciphersuite.

We should consider moving toward hash-identifiers that are of a fixed size just long enough to make collisions negligible (say 128 bits, like a UUID).  A rough proposal:

```
opaque HashIdentifier[16];

MakeHash(value) = KDF.expand(KDF.extract("", value), "MLS 1.0 id", 16)
```

bifurcation:
Fixes #506 
Fixes #504

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -952,6 +952,7 @@ following primitives to be used in group key computations:
   * A Key Derivation Function (KDF)
   * An AEAD encryption algorithm
 * A hash algorithm
+* A MAC algorithm
 * A signature algorithm
 
 MLS uses draft-08 of HPKE {{I-D.irtf-cfrg-hpke}} for public-key encryption.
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -975,6 +976,26 @@ used to add new members).
...
+
+MakeHashRef(value) = KDF.expand(KDF.extract("", value), "MLS 1.0 ref", 16)
+
+HashReference KeyPackageRef;
+HashReference ProposalRef;
+~~~~~
+
+For a KeyPackageRef, the `value` input is the encoded KeyPackage, and the
+ciphersuite specified in the KeyPackage determines the KDF used.  For a
+ProposalRef, the `value` input is the MLSPlaintext carrying the proposal, and
```
franziskuskiefer, line 996: Should the `Commit` section then use `ProposalRef` rathen than `opaque hash<0..255>` for the `reference`?

franziskuskiefer:
Fixes #504

draft-ietf-mls-protocol.md
bifurcation: This is done via the changes to ProposalOrRef further down.

franziskuskiefer: Ah missed that, never mind 🤦🏻

franziskuskiefer (approved):

raphaelrobert (approved):

PR #511, opened 2021-11-16
Remove the requirement for Add in external commit

### LINKED ISSUE #505: Add is unnecessary in external Commit
When a new joiner sends an external Commit, within the current structure, they have to generate and send **two** KeyPackages:  First in an Add proposal, and second in `Commit.path.leaf_key_package`.  Obviously, this is wasteful, since the second immediately overwrites the first.  It also complicates receiver processing, since the receiver has to find where the appropriate Add happened so that it knows the sender's index in the new tree, so that it knows how to apply the update path in the Commit.

We should remove the requirement for an Add commit in an external Commit, and instead assign the new member the leftmost free slot after all proposals have been applied, then apply the commit from there.

bifurcation:
Fixes #505

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2990,10 +2990,8 @@ External Commits work like regular Commits, with a few differences:
 
 * The proposals included by value in an External Commit MUST meet the following
   conditions:
-  * There MUST be a single Add proposal that adds the new issuing new member to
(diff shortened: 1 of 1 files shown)


raphaelrobert (approved):
Approved, but see my comment in https://github.com/mlswg/mls-protocol/issues/505#issuecomment-974637916

PR #513, opened 2021-11-19
Two editorial issues

### LINKED ISSUE #447: Motivate Ratchet Trees
The top of Sec. 5 has exactly one sentence before diving into tree terminology. Please explain at a high level what is the motivation for this structure and maybe how it fits into the rest of the protocol.

### LINKED ISSUE #468: Clarify relationship between Ratchet Tree Evolution, Group Creation, and Key Schedule
The terminology used in these sections is different, so that it's unclear if they're descriptions of the same operation with different levels of detail or entirely different operations performed at different times. I come in having read most of the original white paper, so that my intuition is that it's one operation performed against data stored in the tree structure. As such, I'm unclear how the `init_secret`->`epoch_secret` relates to the `leaf_secret` via the "number of secrets [that] are derived from the epoch secret for different purposes".

Also, The Group Creation section makes no reference to any of these secrets by name. I would kind of expect something called, `init_secret`, to be involved during group creation, but it doesn't seem like that's necessary. 

is `authenticated_data` defined anywhere? Is it application level data to be included as desired?

bifurcation:
Fixes #447 
Fixes #468

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -613,8 +613,8 @@ member of the group requires only log(N) encryptions, instead of the N-1
 encryptions that would be needed to encrypt to each participant individually
 (where N is the number of members in the group).
 
-This efficient remove operation allows MLS to efficiently achieve
-post-compromise security.  In an Update or Commit message, an old, possibly
+This remove operation allows MLS to efficiently achieve
+post-compromise security.  In an Update proposal or a full Commit message, an old, possibly
 compromised representation of a member is effeciently removed from the group and
 replaced with a freshly generated instance.
 
@@ -798,7 +798,7 @@ they receive the private keys for nodes, as described in
 For example, consider a four-member group (A, B, C, D) where the node above the
 right two members is blank.  (This is what it would look like if A created a
 group with B, C, and D.)  Then the public state of the tree and the views of the
-private keys of the tree held be each participant would be as follows, where `_`
+private keys of the tree held by each participant would be as follows, where `_`
 represents a blank node, `?` represents an unknown private key, and `pk(X)`
 represents the public key corresponding to the private key `X`:
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -600,8 +600,8 @@ A              B     ...      Z          Directory       Channel
 
 # Ratchet Trees
 
-The protocol uses "ratchet trees" for deriving shared secrets among
-a group of clients.  A ratchet tree is a distribution of secrets among the
+The protocol uses "ratchet trees" for deriving shared secrets among a group of
+clients.  A ratchet tree is an arrangement of secrets and key pairs among the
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -600,8 +600,23 @@ A              B     ...      Z          Directory       Channel
...
+
+The unique power of a ratchet tree is that it allows a group to efficiently
+remove a member, because it allows a group member to efficiently encrypt new
+entropy to a subset of the group.  A ratchet tree assigns shared keys to
+subgroups of the overall group, so that, for example, encrypting to all but one
+member of the group requires only log(N) encryptions, instead of the N-1
+encryptions that would be needed to encrypt to each participant individually
+(where N is the number of members in the group).
+
+This efficient remove operation allows MLS to efficiently achieve
```
raphaelrobert, line 616:
```suggestion
This remove operation allows MLS to efficiently achieve
```

draft-ietf-mls-protocol.md
```diff
@@ -600,8 +600,23 @@ A              B     ...      Z          Directory       Channel
...
+post-compromise security.  In an Update or Commit message, an old, possibly
```
raphaelrobert, line 617:
```suggestion
post-compromise security.  In an Update proposal or a full Commit message, an old, possibly
```

draft-ietf-mls-protocol.md
```diff
@@ -779,6 +795,35 @@ give it access to all of the nodes above it in the tree.  Leaves are "merged" as
 they receive the private keys for nodes, as described in
 {{ratchet-tree-evolution}}.
 
+For example, consider a four-member group (A, B, C, D) where the node above the
+right two members is blank.  (This is what it would look like if A created a
+group with B, C, and D.)  Then the public state of the tree and the views of the
+private keys of the tree held be each participant would be as follows, where `_`
```
raphaelrobert, line 801:
```suggestion
private keys of the tree held by each participant would be as follows, where `_`
```

draft-ietf-mls-protocol.md
```diff
@@ -954,9 +1003,11 @@ following primitives to be used in group key computations:
 * A hash algorithm
 * A signature algorithm
 
-MLS uses draft-08 of HPKE {{I-D.irtf-cfrg-hpke}} for public-key encryption.
-The `DeriveKeyPair` function associated to the KEM for the ciphersuite maps
-octet strings to HPKE key pairs.
+MLS uses draft-08 of HPKE {{I-D.irtf-cfrg-hpke}} for public-key encryption.  The
```
raphaelrobert, line 1006: Thi reintroduces HPKE draft-08. I guess #512 should be merged after this PR.

raphaelrobert (changes requested):

draft-ietf-mls-protocol.md
bifurcation: I merged #512 so this got fixed as a conflict resolution.

raphaelrobert (approved):

PR #515, opened 2021-11-20
Add protocol version to PGS

raphaelrobert:
This is a follow-up to #501 (which was merged before all comments were addressed).


bifurcation (approved):

kkohbrok:
Sorry for chiming in so late, but isn't this redundant, as the version is included in the ciphersuite? Same for the `Welcome` message.

PR #518, opened 2021-11-30
Consolidate resumption PSK cases

### LINKED ISSUE #502: Consolidate resumption PSK definitions
Right now, we treat the re-init and branch PSK cases as distinct cases, as opposed to just having a general "export PSK from one group+epoch, insert it into another" function.  What I would call a "resumption PSK", following TLS.   Even though we have the distinction, the content of the PSK itself is the same: (group ID, epoch, resumption secret).  So the only difference is that the PreSharedKeyID has a bit that signals the intent of injecting this resumption PSK.

Do we really need that bit?  If there's not a compelling need, it would streamline the API and allow for more flexibility if we just had the notion of a resumption PSK.

bifurcation:
Fixes #502


raphaelrobert (approved):

br-hale:
Comments in associated #502 discussion.

PR #521, opened 2021-12-03
Add algorithms for link-based trees

bifurcation:
As discussed in #507, having algorithms for a "link-based" representation of the tree would help emphasize that the array-based tree representation is not required.  This PR adds them in an appendix.

PR #522, opened 2021-12-06
Generalize tree extend/truncate algorithms

bifurcation:
This PR builds on #507 and #521.  It adds the following:

* An explicit note in `Ratchet Tree Terminology` that implementations may use any representation
* A general description of how nodes are added to / removed from the tree
* Specific algorithms for "array-based trees" and "link-based trees"

(I also made a few minor editorial revisions.)

@TWal @franziskuskiefer - I think this hits the right balance of specificity and generality.  Interested in your thoughts here.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -859,10 +859,10 @@ For example, suppose there is a group with four members, with C an unmerged leaf
 at node 5:
 
 ~~~~~
-      3
+      Y
     __|__
    /     \
-  1       5[C]
+  X       Z[C]
  / \     / \
 A   B   C   D
 
@@ -889,10 +889,10 @@ After applying the UpdatePath, the tree will have the following structure, where
 described above:
 
 ~~~~~
-    np[1] -> 3
+    np[1] -> Y'
            __|__
           /     \
-np[0] -> 1       5[C]
+np[0] -> X'      Z[C]
         / \     / \
        A   B   C   D
            ^
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -747,7 +747,7 @@ In this tree, we can see all of the above rules in play:
 
 * The resolution of node Z is the list \[Z, C\]
 * The resolution of leaf 1 is the empty list \[\]
-* The resolution of root node is the list \[A, CD, C\]
+* The resolution of root node is the list \[A, Z, C\]
 
 Every node, regardless of whether the node is blank or populated, has
 a corresponding _hash_ that summarizes the contents of the subtree
@@ -908,18 +908,19 @@ delete the leaf_secret.
 ## Adding and Removing Leaves
 
 In addition to the path-based updates to the tree described above, it is also
-necessary to add and remove leaves of the tree in order to reflect members being
-added to or removed from the group.  Leaves are always added and removed at the
+necessary to add and remove leaves of the tree in order to reflect changes to the
+membership of the group (see {{add}} and {{remove}}).  Leaves are always added and removed at the
 right edge of the tree: Either a new rightmost leaf is added, or the rightmost
 leaf is removed.  Nodes' parent/child node relationships are then updated to
 maintain the tree's left-balanced structure.  These operations are also known as
 _extending_ and _truncating_ the tree.
 
-To add a new leaf: Add leaf as the new rightmost leaf of the tree.  Add a blank
-parent node whose right child is the new leaf.  The parent is attached to the
-tree as the right child of the appropriate node to make the updated tree
-left-balanced (or set as a new root).  The former right child of the parent's
-parent becomes the parent's left child (or the old root becomes the left child).
+To add a new leaf: Add leaf L as the new rightmost leaf of the tree.  Add
+a blank parent node P whose right child is L.  P is attached to the
+tree as the right child of the only appropriate node to make the updated tree
+left-balanced (or set it as a new root).  The former right child of the P's
+parent becomes P's left child (or the old root becomes the P's left child if
+P is the new root).
 
 ~~~~~
                    _ <-- new parent              _
@@ -930,10 +931,10 @@ parent becomes the parent's left child (or the old root becomes the left child).
 A   B        A   B   C <-- new leaf        A   B   C   D <-- new leaf
 ~~~~~
 
-To remove the rightmost leaf: Remove the rightmost leaf node and its parent
-node.  If the parent node was the root of the tree, the parent node's left child
-is now the root of the tree.  Otherwise, set the right child of the parent's parent
-(formerly the parent itself) to be the parent's left child.
+To remove the rightmost leaf: Remove the rightmost leaf node L and its parent
+node P.  If P was the root of the tree, P's left child
+is now the root of the tree.  Otherwise, set the right child of P's parent
+to be the P's left child.
 
 ~~~~~
       Y                                    Y
@@ -2432,7 +2433,7 @@ An Add is applied after being included in a Commit message.  The position of the
 Add in the list of proposals determines the leaf node where the new member will
 be added.  For the first Add in the Commit, the corresponding new member will be
 placed in the leftmost empty leaf in the tree, for the second Add, the next
-empty leaf to the right, etc. If not empty leaf exists, the tree is extended to
+empty leaf to the right, etc. If no empty leaf exists, the tree is extended to
 the right.
 
 * Validate the KeyPackage:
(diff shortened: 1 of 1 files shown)


kkohbrok (approved):

draft-ietf-mls-protocol.md
```diff
@@ -3196,22 +3222,22 @@ welcome_key = KDF.Expand(welcome_secret, "key", AEAD.Nk)
...
 
-    * Update the leaf at index `index` with the private key corresponding to the
+    * Update the leaf `my_leaf` with the private key corresponding to the
       public key in the node.
 
     * If the `path_secret` value is set in the GroupSecrets object: Identify the
-      lowest common ancestor of the node index `index` and of the node index of
-      the member with KeyPackageRef `GroupInfo.signer`. Set the private key for
+      lowest common ancestor of the leaf node `my_leaf` and of the node of
+      the member with KeyPackageID `GroupInfo.signer`. Set the private key for
```
franziskuskiefer, line 3240: KeyPackageRef

draft-ietf-mls-protocol.md
```diff
@@ -2400,15 +2428,12 @@ struct {
...
-
 An Add is applied after being included in a Commit message.  The position of the
-Add in the list of proposals determines the node index `index` of the leaf node
-where the new member will be added.  For the first Add in the Commit, `index` is
-the leftmost empty leaf in the tree, for the second Add, the next empty leaf to
-the right, etc.
+Add in the list of proposals determines the leaf node where the new member will
+be added.  For the first Add in the Commit, the corresponding new member will be
+placed in the leftmost empty leaf in the tree, for the second Add, the next
+empty leaf to the right, etc. If not empty leaf exists, the tree is extended to
```
TWal, line 2435:
```suggestion
empty leaf to the right, etc. If no empty leaf exists, the tree is extended to
```

TWal:
This looks good!

If I am understanding everything well, the normative description for extention / truncation is the section "Adding and Removing Leaves", which is a tree-based description?

draft-ietf-mls-protocol.md
```diff
@@ -914,6 +905,58 @@ np[0] -> 1       5[C]
...
+right edge of the tree: Either a new rightmost leaf is added, or the rightmost
+leaf is removed.  Nodes' parent/child node relationships are then updated to
+maintain the tree's left-balanced structure.  These operations are also known as
+_extending_ and _truncating_ the tree.
+
+To add a new leaf: Add leaf as the new rightmost leaf of the tree.  Add a blank
+parent node whose right child is the new leaf.  The parent is attached to the
+tree as the right child of the appropriate node to make the updated tree
+left-balanced (or set as a new root).  The former right child of the parent's
+parent becomes the parent's left child (or the old root becomes the left child).
```
TWal, line 922: I suggest giving names to make it easier to read. Also add emphasis that there is a unique way to incorporate the new parent in the tree that keeps it left-balanced (-> " […] as the only appropriate node […]").

```suggestion
To add a new leaf: Add leaf L as the new rightmost leaf of the tree.  Add
a blank parent node P whose right child is L.  P is attached to the
tree as the right child of the only appropriate node to make the updated tree
left-balanced (or set it as a new root).  The former right child of the P's
parent becomes P's left child (or the old root becomes the P's left child if
P is the new root).
```

draft-ietf-mls-protocol.md
```diff
@@ -914,6 +905,58 @@ np[0] -> 1       5[C]
...
+                /    \                        /     \
+  X    ===>    X     |               ===>    X       _ <-- new parent
+ / \          / \    |                      / \     / \
+A   B        A   B   C <-- new leaf        A   B   C   D <-- new leaf
+~~~~~
+
+To remove the rightmost leaf: Remove the rightmost leaf node and its parent
+node.  If the parent node was the root of the tree, the parent node's left child
+is now the root of the tree.  Otherwise, set the right child of the parent's parent
+(formerly the parent itself) to be the parent's left child.
```
TWal, line 936: I also suggest giving names to nodes. Also removing "(formerly the parent itself)" because I don't see what that means.
In the first example, the leaf is D, the parent is Z, the parent's parent is Y, so the sentence says that Y's right child is now Z's left child (so C). I don't get what "former parent" means here.
```suggestion
To remove the rightmost leaf: Remove the rightmost leaf node L and its parent
node P.  If P was the root of the tree, P's left child
is now the root of the tree.  Otherwise, set the right child of P's parent
to be the P's left child.
```

draft-ietf-mls-protocol.md
```diff
@@ -745,18 +736,18 @@ brackets:
...
 ~~~~~
 
 In this tree, we can see all of the above rules in play:
 
-* The resolution of node 5 is the list \[CD, C\]
-* The resolution of node 2 is the empty list \[\]
-* The resolution of node 3 is the list \[A, CD, C\]
+* The resolution of node Z is the list \[Z, C\]
+* The resolution of leaf 1 is the empty list \[\]
+* The resolution of root node is the list \[A, CD, C\]
```
bifurcation, line 750:
```suggestion
* The resolution of root node is the list \[A, Z, C\]
```

draft-ietf-mls-protocol.md
```diff
@@ -914,6 +905,58 @@ np[0] -> 1       5[C]
 After performing these operations, the generator of the UpdatePath MUST
 delete the leaf_secret.
 
+## Adding and Removing Leaves
+
+In addition to the path-based updates to the tree described above, it is also
+necessary to add and remove leaves of the tree in order to reflect members being
+added to or removed from the group.  Leaves are always added and removed at the
```
bifurcation, line 912:
```suggestion
necessary to add and remove leaves of the tree in order to reflect changes to the
membership of the group (see {{add}} and {{remove}}).  Leaves are always added and removed at the
```

draft-ietf-mls-protocol.md
```diff
@@ -914,6 +905,58 @@ np[0] -> 1       5[C]
...
+is now the root of the tree.  Otherwise, set the right child of the parent's parent
+(formerly the parent itself) to be the parent's left child.
+
+~~~~~
+      Y                                    Y
+    __|__                                __|_
+   /     \                              /    \
+  X       Z <-- remove parent  ===>    X     | <-- reassign child
+ / \     / \                          / \    |
+A   B   C   D <-- remove leaf        A   B   C
```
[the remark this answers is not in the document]
bifurcation: `D` isn't necessarily a blank leaf.   In principle, one can remove the rightmost leaf even if it's populated.  This section is just describing the structure modifications; the later sections deal with what's in the tree when it gets update.

draft-ietf-mls-protocol.md
bifurcation:
```suggestion
      the member with KeyPackageRef `GroupInfo.signer`. Set the private key for
```

draft-ietf-mls-protocol.md
```diff
@@ -3263,6 +3289,62 @@ struct {
...
+    k = 0
+    while (2**(k+1)) < len(nodes):
+       k += 1
+    R = 2**k - 1
+    root = Node(nodes[R])
+    root.left = subtree_root(nodes[:R])
+    root.right = subtree_root(nodes[(R+1):])
+    return root
+~~~~~
+
```
bifurcation, line 3325:
```suggestion

(Note that this is the same ordering of nodes as in the array-based tree representation
described in {{array-based-trees}}.  The algorithms in that section may be used to 
simplify decoding this extension into other representations.)

```

bifurcation, line 3325:
```suggestion

(Note that this is the same ordering of nodes as in the array-based tree representation
described in {{array-based-trees}}.  The algorithms in that section may be used to 
simplify decoding this extension into other representations.)

```

draft-ietf-mls-protocol.md
```diff
@@ -4092,3 +4183,94 @@ def common_ancestor_direct(x, y, _):
...
+            return
+
+        # If X is full, then X is the root, so P replaces the root
+        P.left = self.root
+        self.root.parent = P
+        self.root = P
+        return
+
+    def truncate(self):
+        X = self.root.rightmost_leaf()
```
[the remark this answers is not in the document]
bifurcation: As above, not true that we can only truncate blank leaves.

draft-ietf-mls-protocol.md
```diff
@@ -914,6 +905,59 @@ np[0] -> 1       5[C]
...
+
+
+      Y <-- remove parent
+    __|_
+   /    \
+  X     |                  ===>    X <-- reassign root
+ / \    |                         / \
+A   B   C <-- remove leaf        A   B
+~~~~~
+
```
TWal, line 955: Even if this only describe the structure modification, I think it is valuable to give hints to the reader that in practice only blank leaves will be removed. If here the reader thinks that we can truncate any leaf and only afterward the reader discovers that only blank leaves can be truncated, I think it can be confusing.
Also it might look like a violation of the tree invariant, since in the second tree, D knows Y's secret without being its descendant.

```suggestion

Note: in the rest of the protocol, the rightmost leaf will be removed only when it is blank.

```

bifurcation: Fair enough.  I think I will word this slightly differently, but I can grant that it makes sense to have a note here.

franziskuskiefer (approved):
Thanks lgtm now

PR #524, opened 2021-12-14
Careful truncation

bifurcation:
As @kkohbrok noted [on the mailing list](https://mailarchive.ietf.org/arch/msg/mls/P8CiOeQy3lev7s5vbnaRkPwRa1Q/), the current truncation algorithm breaks parent hashes.  This PR implements the "careful truncation" algorithm he proposes.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2489,10 +2489,12 @@ A member of the group applies a Remove message by taking the following steps:
 
 * Blank the intermediate nodes along the path from `removed_index` to the root
 
-* Truncate the tree by removing leaves from the right side of the tree until
-  either the rightmost leaf node is not blank or the parent of the rightmost
-  leaf node is not blank.  (If the parent node is not blank, then it is needed
-  to verify parent hashes in the truncated tree.)
+* Truncate the tree by removing leaves from the right side of the tree as long
+  as the following conditions hold (since non-blank intermediate nodes hold
+  information that is necessary for verifying parent hashes):
+
+  * The rightmost leaf is blank
+  * The parent of the rightmost leaf is either blank or the root of the tree
 
 ### PreSharedKey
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2489,8 +2489,10 @@ A member of the group applies a Remove message by taking the following steps:
 
 * Blank the intermediate nodes along the path from `removed_index` to the root
 
-* Truncate the tree by removing leaves from the right side of the tree until the
-  rightmost leaf node is not blank.
+* Truncate the tree by removing leaves from the right side of the tree until
+  either the rightmost leaf node is not blank or the parent of the rightmost
+  leaf node is not blank.  (If the parent node is not blank, then it is needed
+  to verify parent hashes in the truncated tree.)
```
kkohbrok, line 2495: If the parent node is the root node, it can be removed even if it is non-blank, since removing the root node doesn't destroy any valid parent hash chains.

bifurcation: Good point.  Updated to reflect this.

kkohbrok (approved):

draft-ietf-mls-protocol.md
```diff
@@ -2489,8 +2489,12 @@ A member of the group applies a Remove message by taking the following steps:
 
 * Blank the intermediate nodes along the path from `removed_index` to the root
 
-* Truncate the tree by removing leaves from the right side of the tree until the
-  rightmost leaf node is not blank.
+* Truncate the tree by removing leaves from the right side of the tree as long
+  as the following conditions hold (since non-blank intermediate nodes hold
```
raphaelrobert, line 2493:
```suggestion
  as all of the following conditions hold (since non-blank intermediate nodes hold
```

raphaelrobert:
I think this should be good to go?

Bren2010 (approved):

bifurcation:
Merging this for now to un-break the spec.  I will work with Marta to see if we can make a follow-up that removes redundant nodes, so that we can get back to more aggressive truncation.

PR #525, opened 2021-12-15
Tighten up branch and reinit; define ResumptionPSKUsage

bifurcation:
This PR reflects further discussion on #502 and replaces #518.  The high-level impacts are:

* Add clearer, more prescriptive definitions of the re-init and branch operations
* Use a single syntax for all resumption PSKs, but also have a `usage` parameter to distinguish cases
* Allow an `application` usage to cover any other cases defined by applications

Once #520 lands, we should also add some text to the protocol overview to describe these operations.  Pasting some text here for usage in that future PR...

```
## Relationships Between Epochs

A group comprises a single linear sequence of epochs. However, it can sometimes
be useful to express relationships between epochs, either within a group or
across groups.  MLS derives a resumption pre-shared key (PSK) from each epoch to
allow a entropy extracted from one epoch to be injected into a future epoch,
effectively creating a causal link between the two epochs.

MLS supports two ways to tie a new group to an existing group. Re-initialization
closes one group and creates a new group comprising the same members with
different parameters. Branching starts a new group with a subset of the original
group's participants (with no effect on the original group).  In both cases,
the new group is linked to the old group via a resumption PSK.

~~~~~
epoch_A_[n-1]
     |
     |
     |<-- ReInit
     |
     V
epoch_A_[n]           epoch_B_[0]
     .                     |
     .  PSK(usage=reinit)  |
     .....................>|
                           |
                           V
                      epoch_B_[1]
~~~~~
{: title="Reinitializing a group" }


~~~~~
epoch_A_[n-1]
     |
     |
     |<-- ReInit
     |
     V
epoch_A_[n]           epoch_B_[0]
     |                     |
     |  PSK(usage=branch)  |
     |....................>|
     |                     |
     V                     V
epoch_A_[n+1]         epoch_B_[1]
~~~~~
{: title="Branching a group" }

Applications may also choose to use resumption PSKs to link epochs in other
ways.  For example, the following figure shows a case where a resumption PSK
from epoch `n` is injected into epoch `n+k`.  This demonstrates that the members
of the group at epoch `n+k` were also members at epoch `n` (irrespective of any
changes to these members' keys due to Updates or Commits.

~~~~~
epoch_A_[n-1]
     |
     |
     |<-- ReInit
     |
     V
epoch_A_[n]
     |                     
     |  PSK(usage=application)  
     |.....................
     |                    . 
     |                    . 
    ...                  ...
     |                    . 
     |                    . 
     V                    .
epoch_A_[n+k-1]           .
     |                    . 
     |                    . 
     |<.................... 
     |                      
     V                     
epoch_A_[n+k]       
~~~~~
{: title="Reinjecting entropy from an earlier epoch" }
```

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1643,33 +1643,31 @@ compromise.  These factors are outside of the scope of this document, but should
 be considered by application designers relying on PSKs.
 
 Each PSK in MLS has a type that designates how it was provisioned.
-External PSKs are provided by the application, while recovery and re-init PSKs
+External PSKs are provided by the application, while resumption PSKs
 are derived from the MLS key schedule and used in cases where it is
-necessary to authenticate a member's participation in a prior group state.
-In particular, in addition to external PSK types, a PSK derived from within MLS
-may be used in the following cases:
-
-  - Re-Initialization: If during the lifetime of a group, the group members
-    decide to switch to a more secure ciphersuite or newer protocol version,
-    a PSK can be used to carry entropy from the old group forward into a new
-    group with the desired parameters.
-
-  - Branching: A PSK may be used to bootstrap a subset of current group
-    members into a new group. This applies if a subset of current group
-    members wish to branch based on the current group state.
+necessary to authenticate a member's participation in a prior epoch.
 
 The injection of one or more PSKs into the key schedule is signaled in two ways:
-1) as a `PreSharedKey` proposal, and 2) in the `GroupSecrets` object of a
-Welcome message sent to new members added in that epoch.
+Existing members are informed via PreSharedKey proposals covered by a Commit,
+and new members added in the Commit are informed via GroupSecrets object in the
+Welcome message corresponding to the Commit.  To ensure that existing and new
+members compute the same PSK input to the key schedule, the Commit and
+GroupSecrets objects MUST indicate the same set of PSKs, in the same order.
 
 ~~~~~
 enum {
   reserved(0),
   external(1),
-  reinit(2),
-  branch(3)
+  resumption(2),
   (255)
 } PSKType;
+
+enum {
+  reserved(0),
+  application(1),
+  reinit(2),
+  branch(3),
+} ResumptionPSKUsage;
 
 struct {
   PSKType psktype;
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2323,25 +2314,52 @@ extensions can be updated, a GroupContextExtensions proposal is invalid if it
...
+in the Welcome message that creates the branched subgroup.
+
+A client receiving a Welcome including a PreSharedKey of type `resumption` with
+usage `branch` MUST verify that the new group reflects a subgroup branched from
+the referenced group.
+
+* The `version` and `ciphersuite` values in the Welcome MUST be the same as
+  those used by the old group.
+* Each KeyPackage in a leaf node of the new group's tree MUST be a leaf in the
+  old group's tree at the epoch indicated in the PreSharedKey.
```
kkohbrok, line 2355: Do we really want the KeyPackages to be identical? This seems to constitute key-reuse across groups. This is where a (identity,endpoint_id) tuple would come in handy ;-) But I think going with the new multi-identity credentials, we could mandate here that the application has to ensure that the participants in the new group are a subset of the participants in the old group

kkohbrok (commented):
Looks good! Although I'm not sure in the branching keys we want to have the exact same key packages as in the main group.

kkohbrok:
One question regarding the diagram in the PR description. In the re-init case, shouldn't the "B" group start at epoch `n+1`? At least that's what the PR specifies in the spec.

TWal:
I think the re-init diagram is fine because the spec says

    The `epoch` in the Welcome message MUST be 1

However in the branch case (and application case) diagram I don't think there should be a ReInit proposal? (At least it is not said in the spec).

Also, it is not entirely clear how to construct a `PreSharedKeyID` in the branch case (or application case): what values should have `psk_group_id` and `psk_epoch`?

kkohbrok:
Ah, sorry I misread. You are right and the re-init diagram is fine the way it is.

raphaelrobert (commented):
I think this looks fine generally (just one comment)

br-hale:
Minor typo in the text above: "allow a entropy" -> "allow entropy".

PR #526, opened 2021-12-16
Unambiguous signatures

TWal:
Signatures as specified in the MLS protocol document are ambiguous, in the sense that signatures produced for one purpose in the protocol can be (maliciously) used in a different place of the protocol.
Indeed, we found that for example, a `KeyPackage` can have the same serialization as `MLSPlaintextTBS`, which proves that one signature could be accepted for different structures.

The collisions we have found so far are quite artificial, so it is not super clear how to exploit this, but at least it makes it impossible to have provable security.

The solution is pretty simple, we simply prefix each signed value with a label, in a fashion similar to `ExpandWithLabel`.
That way, when calling `SignWithLabel` you can give the structure name in the `Label` argument.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3348,7 +3348,6 @@ welcome_nonce = KDF.Expand(welcome_secret, "nonce", AEAD.Nn)
 welcome_key = KDF.Expand(welcome_secret, "key", AEAD.Nk)
 ~~~~~
 
-
 * Verify the signature on the GroupInfo object. The signature input comprises
   all of the fields in the GroupInfo object except the signature field. The
   public key and algorithm are taken from the credential in the leaf node of the
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1025,6 +1025,27 @@ the same as the signature algorithm specified in the credential field of the
 KeyPackage objects in the leaves of the tree (including the InitKeys
 used to add new members).
 
+To disambiguate different signatures used in MLS, each signed value is prefixed
+by a label as shown below:
+
+~~~~~
+SignWithLabel(SignatureKey, Label, Content) =
+    Signature.Sign(SignatureKey, SignContent)
+
+VerifyWithLabel(VerificationKey, Label, Content) =
+    Signature.Verify(VerificationKey, SignContent)
+
+Where SignContent is specified as:
+
+struct {
+    opaque label<7..255> = "mls10 " + Label;
+    opaque content<0..2^32-1> = Content;
+} SignContent;
+~~~~~
+
+Here, the functions `Signature.Sign` and `Signature.Verify` are defined
+by the signature algorithm.
+
 The ciphersuites are defined in section {{mls-ciphersuites}}.
 
 ## Hash-Based Identifiers
@@ -1145,8 +1166,10 @@ The value for hpke\_init\_key MUST be a public key for the asymmetric
 encryption scheme defined by cipher\_suite. The whole structure
 is signed using the client's signature key. A KeyPackage object
 with an invalid signature field MUST be considered malformed.
-The input to the signature computation comprises all of the fields
-except for the signature field.
+
+The signature is computed by the function `SignWithLabel` with a label
+`KeyPackage` and a content comprising of all of the fields except for the
+signature field.
 
 ~~~~~
 enum {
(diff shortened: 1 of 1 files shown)


kkohbrok (approved):
Great catch and a neat solution!

TWal:
Something that is still bothering me is the `MLSPlaintextTBS` structure:

    struct {
        select (MLSPlaintextTBS.sender.sender_type) {
            case member:
                GroupContext context;
    
            case preconfigured:
            case new_member:
                struct{};
        }
    
        WireFormat wire_format;
        opaque group_id<0..255>;
        uint64 epoch;
        Sender sender;
    
        ContentType content_type;
        select (MLSPlaintextTBS.content_type) {
            case application:
    
            case proposal:
              Proposal proposal;
    
            case commit:
              Commit commit;
        }
    } MLSPlaintextTBS;

Since the first `select` is done on a field defined after it, this structure is not parsable. We could say that's not a problem since we only need to serialize it, but for security purposes I would recommend to put it at the end, like this:

    struct {
        WireFormat wire_format;
        opaque group_id<0..255>;
        uint64 epoch;
        Sender sender;
    
        ContentType content_type;
        select (MLSPlaintextTBS.content_type) {
            case application:
    
            case proposal:
              Proposal proposal;
    
            case commit:
              Commit commit;
        }

        select (MLSPlaintextTBS.sender.sender_type) {
            case member:
                GroupContext context;
    
            case preconfigured:
            case new_member:
                struct{};
        }
    } MLSPlaintextTBS;

What do you think about it?

raphaelrobert:
Nice one! Crafting collisions is probably not straightforward, but eliminating the the risk altogether is the way to go.

TWal:
It is actually quite easy to craft collisions, I made a script do find them: https://github.com/TWal/TLSCollisionFinder
(I wanted to check whether there was accidentally a way do disambiguate the structures)

draft-ietf-mls-protocol.md
```diff
@@ -2233,6 +2248,15 @@ struct {
...
     }
+
+    select (MLSPlaintextTBS.sender.sender_type) {
+        case member:
+            GroupContext context;
+
+        case preconfigured:
+        case new_member:
+            struct{};
+    }
```
bifurcation, line 2259: I don't love the aesthetics here.  Maybe we could solve this a little more elegantly by having a `context` input to `SignWithLabel`/`VerifyWithLabel`?

Alternatively: Do we even need this field any more?  This field was added before `membership_tag`, and now the `membership_tag` already (a) demonstrates knowledge of the GroupContext and (b) prevents a non-member from synthesizing a valid MLSPlaintext.

draft-ietf-mls-protocol.md
```diff
@@ -2536,7 +2561,7 @@ the right.
 * Validate the KeyPackage:
 
     * Verify that the signature on the KeyPackage is valid using the public key
-      in the KeyPackage's credential
+      in the KeyPackage's credential and the label `"KeyPackage"`
```
bifurcation, line 2564: I would prefer that we put these descriptions with the structs they belong to, as opposed to down here in the instructions.  We already have prose that says what the signature covers; that should just specify the label as well.  You could also include it as a comment in the syntax specification:

```
struct {
    ...
    // SignWithLabel(., "MLSPlaintextTBS", MLSPlaintextTBS)
    opaque signature<0..2^16-1>;
} KeyPackage;

struct {
    ...
    // SignWithLabel(., "GroupInfoTBS", GroupInfoTBS)
    opaque signature<0..2^16-1>;
} GroupInfo;

struct {
    ...
    // SignWithLabel(., "PublicGroupStateTBS", PublicGroupStateTBS)
    opaque signature<0..2^16-1>;
} PublicGroupState;

struct {
    ...
    // SignWithLabel(., "MLSPlaintextTBS", MLSPlaintextTBS)
    opaque signature<0..2^16-1>;

    // MAC(confirmation_key, GroupContext.confirmed_transcript_hash)
    optional<MAC> confirmation_tag;

    // MAC(membership_key, MLSPlaintextTBM);
    optional<MAC> membership_tag;
} MLSPlaintext;
```

(Note: I don't think GroupInfoTBS exists, but it probably should.)

bifurcation (changes requested):

draft-ietf-mls-protocol.md
TWal: I not sure the alternative is a good idea for several reasons, for example the `membership_tag` is present only in `MLSPlaintext` and not in `MLSCiphertext`, also I'm not confident in removing things from the signature without a solid proof that it is not necessary (for example, signing the tree hash seems like a reasonable thing to do?).

I feel like adding a `context` input is a bit weird. What do you think about this third option: adding an `option<GroupContext> group_context;` and say something like "group_context MUST contain a value if sender_type = ..." below? This pattern already exists for `confirmation_tag` for example.

draft-ietf-mls-protocol.md
kkohbrok: (#529 adds a `GroupInfoTBS`)

TWal: If every structure has a `TBS` associated with it, this seems like a nice formatting of how you must compute the signatures, however for `KeyPackage` there is no `TBS` (yet), it is said in the text above that the TBS contains "all of the fields except for the signature field".

Should we add a `TBS` for every structure that we sign? This is for example done for `PublicGroupState` where a `TBS` is explicitly described even if it contains all fields except the signature field in `PublicGroupState`.

bifurcation (approved):

PR #527, opened 2021-12-16
Stronger parent hashes for authenticated identities

TWal:
Our proposal
============

The parent hash construction currently authenticates only the public keys in the ratchet tree and does not protect other tree contents, notably the credentials at the leaves.
Consequently, a malicious member can tamper with the credentials before sending a Welcome package to a new member, even if it cannot tamper with the public keys.

Instead of using the tree resolution in the parent hash computation, we propose to use the tree hash.

The last time this was discussed on the list, it was agreed that it would be better to use tree hashes from a security point of view, but it was felt that complexity-wise that the recipient of a Welcome message would have to do O(n^2) work.
A more thorough analysis shows that processing a Welcome message can be done in O(n*log(log(n))).
The result is that each signature captures the full tree contents including leaf credentials, which provides better authentication guarantees for member identities.

Efficient algorithm to process a Commit
=======================================

When processing Commits, the implementation can store the tree hashes of every subtree.
That way, with n being the number of leaves, you can process each proposal in O(log(n)) and each UpdatePath in O(log(n)).
Then you can check the parent hash when processing a Commit in O(p*log(n)) where p is the number of proposals.

Efficient (enough) algorithm to process a Welcome
=================================================

When processing a Welcome message, we need to do the parent hash verification procedure.
To do it with the new parent hash, we need to compute tree hashes of subtrees at the last time they were updated.
This can be done using the following modified tree hash:

    OldTreeHash(leaf, excluded_leaves) = Hash(leaf) if leaf not in excluded_leaves
    OldTreeHash(leaf, excluded_leaves) = Hash(blank leaf) if leaf in excluded_leaves
    OldTreeHash(node, excluded_leaves) = Hash(node.content | OldTreeHash(node.left, left_excluded_leaves) | OldTreeHash(node.right, right_excluded_leaves))
    where left_excluded_leaves = leaves that are in excluded_leaves and in node.left (same for right_excluded_leaves).
    and `excluded_leaves` are removed from `node.content.unmerged_leaves`.

Then we can compute the parent hash for a node `N` with child `C` and sibling `S` using `OldTreeHash(S, N.content.unmerged_leaves)`: this old tree hash don't change when Adds are applied to the tree.

In the next paragraph, we will see that we can compute the parent hash for every node in `O(n*log(n))`.

Let `h` be the height of the tree (so `h = log(n)`). There are `O(2^{h-i})` nodes at height `i`, and computing the parent hash for a node at height `i` is done in `O(2^i)`.
Hence, the complexity for the parent hash verification procedure is:

      \sum_{i=1}^h O(2^i 2^{h-i})
    = \sum_{i=1}^h O(2^h)
    = O(h 2^h)
And since `h = log(n)`, the complexity is `O(n*log(n))`.

Efficient algorithm to process a Welcome
========================================

Imagine we have the following tree:

                  W
            ______|______
           /             \
          U               Y [E, G]
        __|__           __|__
       /     \         /     \
      T       V       X [E]   Z [G, H]
     / \     / \     / \     / \
    A   B   C   D   E   F   G   H

For each node `N`, `OldTreeHash(N, excluded_leaves)` will be called with different values for `excluded_leaves`.

For example, if the sibling is always the right child, then:
- `OldTreeHash(H, .)` will be called with 1 time with `. = [H]` (when computing the parent hash of `Z`) and 2 times with `. = []` (when computing the parent hash of `Y` and `W`)
- `OldTreeHash(Z, .)` will be called with 1 time with `. = [G]` (when computing the parent hash of `Y`) and 1 time with `. = []` (when computing the parent hash of `W`)

If the sibling is always the left child, then:
- `OldTreeHash(A, .)` will be called 3 times with `. = []` (when computing the parent hash of `T`, `U`, `W`)
- `OldTreeHash(T, .)` will be called 2 times with `. = []` (when computing the parent hash of `U`, `W`)

We can see that some values are computed several times, which means that the whole computation could be sped up using memoization.

To estimate the speedup, we will estimate with how many values for `excluded_leaves` is called `OldTreeHash(N, .)` for each node `N`. Summing up these values will give a complexity, which will be `O(n*log(log(n)))`.

Let `node.parent[k]` be the `k`th parent of `node`, and `node.leaves` be the set of leaves in the subtree rooted at `node`.
Then, each call of `OldTreeHash(node, excluded_leaves)` will have `excluded_leaves = Intersection(node.parent[k].unmerged_leaves, node.leaves)` for some `k`.
Let's note this set `excluded_leaves_k`.
The tree has the invariant that if `i <= j` then `excluded_leaves_j` is a subset of `excluded_leaves_i` (i.e. the sets `excluded_leaves_k` are decreasing).

Let d be the depth of `node` (with the root being at depth 1). Then, the set `excluded_leaves_set := {excluded_leaves_1, ..., excluded_leaves_{d-1}}` corresponds to the values for `excluded_leaves` in the calls to `OldTreeHash(N, .)`.

First bound: `|excluded_leaves_set| <= d-1 <= d` because it is defined using `d-1` values.
Second bound: `|excluded_leaves_set| <= 2^{h-d}+1` because `|excluded_leaves_0| <= 2^{h-d}` and the sets `excluded_leaves_k` are decreasing.
Therefore: for a node at depth d, `|excluded_leaves_set| <= min(d, 2^{h-d}+1)`

The complexity to do the parent hash verification procedure is therefore in `O(sum_{i=1}^h 2^i min(i, 2^{h-i}+1))`.

Let's estimate this complexity more explicitly:

       sum_{i=1}^h 2^i min(i, 2^{h-i}+1)
    <= (sum_{i=1}^t i 2^i) + (sum_{i=t+1}^h 2^i (2^{h-i}+1)) [for all t]
    =  (2 (t-1) 2^t + 2) + ((h-t) 2^h + 2^{h+1} - 2^{t+1}) [for all t]

For the first inequality, we split the sum in two parts and use `min(i, 2^{h-i}+1) <= i` in the left part and `min(i, 2^{h-i}+1) <= 2^{h-i}+1` in the right part.
For the second equality, it can be derived using the formula for geometric sum.

If we take `t = h - log(h)`, we can compute that the left part is `O(2^h)` and the right part is `O(log(h)*2^h)`.

Therefore, the parent hash verification procedure complexity is `O(log(h)*2^h) = O(n*log(log(n)))`.

One might say that adding memoization has a cost and isn't in O(1), but here it is: using the fact that `excluded_leaves_k` are decreasing, if you process the tree in a bottom-up or top-down way, you only have to compare the length of the current `excluded_leaves` to the length of the previous one, and it tells you if you need to recompute the tree hash.

The current parent hash is not super fast
=========================================

In the worst case, if most internal nodes are blank, then computing the parent hash for a node is in O(size of its subtree), which give an O(n*log(n)) complexity as we saw in the section about the naive algorithm.

When processing a Commit, if most of the internal nodes are blank, then computing the new parent hashes is in O(n) which is quite slow.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1205,7 +1205,7 @@ to which PK's secret key was encrypted in the commit packet that anounced the
 struct {
     HPKEPublicKey public_key;
     opaque parent_hash<0..255>;
-    HPKEPublicKey original_child_resolution<0..2^32-1>;
+    opaque original_sibling_tree_hash<0..255>;
 } ParentHashInput;
 ~~~~~
 
@@ -1215,13 +1215,34 @@ is the root, then `parent_hash` is set to a zero-length octet string.
 Otherwise `parent_hash` is the Parent Hash of P's parent with P's sibling as the
 co-path child.
 
-Finally, `original_child_resolution` is the array of `HPKEPublicKey` values of the
-nodes in the resolution of S but with the `unmerged_leaves` of P omitted. For
-example, in the ratchet tree depicted in {{resolution-example}} the
-`ParentHashInput` of node 5 with co-path child 4 would contain an empty
-`original_child_resolution` since 4's resolution includes only itself but 4 is also
-an unmerged leaf of 5. Meanwhile, the `ParentHashInput` of node 5 with co-path child
-6 has an array with one element in it: the HPKE public key of 6.
+Finally, `original_sibling_tree_hash` is the original tree hash of S. The
+original tree hash corresponds to the tree hash of S the last time P was
+updated. It can be computed as the tree hash of S modified the following way:
+
+* reset the leaves in P.unmerged_leaves to blanks
+* remove P.unmerged_leaves from all unmerged_leaves lists
+
+For example, in the following tree:
+
+~~~~~
+    ABCD[C]
+    __|__
+   /     \
+  AB      CD[C]
+ / \     / \
+A   B   C   D
+~~~~~
+
+With P = ABCD and S = CD, `original_sibling_tree_hash` is the tree hash of the
+following tree:
+
+~~~~~
+  CD
+ / \
+_   D
+~~~~~
+
+Because ABCD.unmerged_leaves = [C], C is removed from CD.unmerged_leaves and C is replaced with a blank leaf.
 
 ### Using Parent Hashes
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1414,6 +1414,7 @@ updated. It can be computed as the tree hash of S modified the following way:
 
 * reset the leaves in P.unmerged_leaves to blanks
 * remove P.unmerged_leaves from all unmerged_leaves lists
+* truncate the tree as described in {{remove}}
 
 For example, in the following tree:
 
@@ -1424,7 +1425,7 @@ For example, in the following tree:
       U               Y [F, H]
     __|__           __|__
    /     \         /     \
-  T       V       X [F]   Z [H]
+  T       V       X [F]   _
  / \     / \     / \     / \
 A   B   C   D   E   F   G   H
 ~~~~~
@@ -1436,12 +1437,14 @@ following tree:
       Y [F]
     __|__
    /     \
-  X [F]   Z
- / \     / \
-E   F   G   _
+  X [F]  |
+ / \     |
+E   F    G
 ~~~~~
 
-Because W.unmerged_leaves = [H], H is removed from Y.unmerged_leaves and Z.unmerged_leaves, then H is replaced with a blank leaf.
+Because `W.unmerged_leaves = [H]`, H is removed from `Y.unmerged_leaves`,
+then H is replaced with a blank leaf, then the tree is truncated removing the
+last two nodes.
 
 ### Using Parent Hashes
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2037,7 +2037,20 @@ the following way:
 * reset the leaves in P.unmerged_leaves to blanks
 * remove P.unmerged_leaves from all unmerged_leaves lists
 * truncate the ratchet tree as described in {{remove}}
+Note that no recomputation is needed if the tree hash of S is unchanged since
+the last time P was updated.  This is the case for computing or processing a
+Commit whose UpdatePath traverses P, since the Commit itself resets P.  (In
+other words, it is only necessary to recompute the original sibling tree hash
+when validating group's tree on joining.) More generally, if none of the entries
+in `P.unmerged_leaves` is in the subtree under S (and thus no nodes were truncated),
+then the original tree hash at S is the tree hash of S in the current tree.
 
+If it is necessary to recompute the original tree hash of a node, the efficiency
+of recomputation can be improved by caching intermediate tree hashes, to avoid
+recomputing over the subtree when the subtree is included in multiple parent
+hashes.  A subtree hash can be reused as long as the intersection of the
+parent's unmerged leaves with the subtree is the same as in the earlier
+computation.
 For example, in the following tree:
 
 ~~~~~
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2037,20 +2037,7 @@ the following way:
 * reset the leaves in P.unmerged_leaves to blanks
 * remove P.unmerged_leaves from all unmerged_leaves lists
 * truncate the ratchet tree as described in {{remove}}
-Note that no recomputation is needed if the tree hash of S is unchanged since
-the last time P was updated.  This is the case for computing or processing a
-Commit whose UpdatePath traverses P, since the Commit itself resets P.  (In
-other words, it is only necessary to recompute the original sibling tree hash
-when validating group's tree on joining.) More generally, if none of the entries
-in `P.unmerged_leaves` is in the subtree under S (and thus no nodes were truncated),
-then the original tree hash at S is the tree hash of S in the current tree.
 
-If it is necessary to recompute the original tree hash of a node, the efficiency
-of recomputation can be improved by caching intermediate tree hashes, to avoid
-recomputing over the subtree when the subtree is included in multiple parent
-hashes.  A subtree hash can be reused as long as the intersection of the
-parent's unmerged leaves with the subtree is the same as in the earlier
-computation.
 For example, in the following tree:
 
 ~~~~~
@@ -2095,6 +2082,21 @@ A   B   C   _
 
 This time we have 4 leaf nodes because the truncation of the ratchet tree didn't
 remove the last leaf.
+
+Note that no recomputation is needed if the tree hash of S is unchanged since
+the last time P was updated.  This is the case for computing or processing a
+Commit whose UpdatePath traverses P, since the Commit itself resets P.  (In
+other words, it is only necessary to recompute the original sibling tree hash
+when validating group's tree on joining.) More generally, if none of the entries
+in `P.unmerged_leaves` is in the subtree under S (and thus no nodes were truncated),
+then the original tree hash at S is the tree hash of S in the current tree.
+
+If it is necessary to recompute the original tree hash of a node, the efficiency
+of recomputation can be improved by caching intermediate tree hashes, to avoid
+recomputing over the subtree when the subtree is included in multiple parent
+hashes.  A subtree hash can be reused as long as the intersection of the
+parent's unmerged leaves with the subtree is the same as in the earlier
+computation.
 
 ### Using Parent Hashes
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2031,22 +2031,23 @@ co-path child.
 
 Finally, `original_sibling_tree_hash` is the original tree hash of S. The
 original tree hash corresponds to the tree hash of S the last time P was
-updated. It can be computed as the tree hash of S modified the following way:
+updated. It can be computed as the tree hash of S in the ratchet tree modified
+the following way:
 
 * reset the leaves in P.unmerged_leaves to blanks
 * remove P.unmerged_leaves from all unmerged_leaves lists
-* truncate the tree as described in {{remove}}
+* truncate the ratchet tree as described in {{remove}}
 
 For example, in the following tree:
 
 ~~~~~
-              W [H]
+              W [D, H]
         ______|_____
        /             \
-      U               Y [F, H]
+      U [D]           Y [F, H]
     __|__           __|__
    /     \         /     \
-  T       V       X [F]   _
+  T       _       X [F]   _
  / \     / \     / \     / \
 A   B   C   D   E   F   G   H
 ~~~~~
@@ -2066,6 +2067,21 @@ E   F    G
 Because `W.unmerged_leaves = [H]`, H is removed from `Y.unmerged_leaves`,
 then H is replaced with a blank leaf, then the tree is truncated removing the
 last two nodes.
+
+With P = W and S = U, `original_sibling_tree_hash` is the tree hash of the
+following tree:
+
+~~~~~
+      U
+    __|__
+   /     \
+  T       _
+ / \     / \
+A   B   C   _
+~~~~~
+
+This time we have 4 leaf nodes because the truncation of the ratchet tree didn't
+remove the last leaf.
 
 ### Using Parent Hashes
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1225,24 +1225,30 @@ updated. It can be computed as the tree hash of S modified the following way:
 For example, in the following tree:
 
 ~~~~~
-    ABCD[C]
-    __|__
-   /     \
-  AB      CD[C]
- / \     / \
-A   B   C   D
+              W [H]
+        ______|_____
+       /             \
+      U               Y [F, H]
+    __|__           __|__
+   /     \         /     \
+  T       V       X [F]   Z [H]
+ / \     / \     / \     / \
+A   B   C   D   E   F   G   H
 ~~~~~
 
-With P = ABCD and S = CD, `original_sibling_tree_hash` is the tree hash of the
+With P = W and S = Y, `original_sibling_tree_hash` is the tree hash of the
 following tree:
 
 ~~~~~
-  CD
- / \
-_   D
+      Y [F]
+    __|__
+   /     \
+  X [F]   Z
+ / \     / \
+E   F   G   _
 ~~~~~
 
-Because ABCD.unmerged_leaves = [C], C is removed from CD.unmerged_leaves and C is replaced with a blank leaf.
+Because W.unmerged_leaves = [H], H is removed from Y.unmerged_leaves and Z.unmerged_leaves, then H is replaced with a blank leaf.
 
 ### Using Parent Hashes
 


raphaelrobert:
@psyoptix Since you worked on this last, it would be great if you could have a look at this.

bifurcation:
I agree that this scheme seems strictly stronger than the current scheme, and it
probably has the same operational properties given that the original sibling
tree hash changes when the original sibling resolution changes, and vice
versa.

The main thing that seems trickier about this version is the "reversing"
operation, i.e., the process of computing the earlier state from the current
state.  With the resolution approach, it's just a question of removing some
values from a list of public keys.  With this approach, you need to replicate
the actual structure of the tree.

As usual, the trouble is at the right edge of the tree.  For a parent node along
the right edge of the tree with unmerged leaves added to the right edge of the
tree, you can't just replace the blanks with blanks, you need to remove the
intermediate nodes that were added as well.  

Looking at your example case, I think you've actually drawn an impossible tree!
For the node Z to be populated, it has to have been set by either G or H.  If H
had done it, then obviously H wouldn't be unmerged.  If G did it, then it would
have encrypted Z to H if H were present, or if it were absent, then G's parent
would be Y and Z would still be blank from H being added.

So the more realistic case is something like:

~~~~~
              W [H]
        ______|_____
       /             \
      U               Y [F, H]
    __|__           __|__
   /     \         /     \
  T       V       X [F]   _
 / \     / \     / \     / \
A   B   C   D   E   F   G   H
~~~~~

Just to be confirm this is realistic: This could arise via the following sequence of operations:

1. Start with A, B, C, D
2. A adds E, F', G; commit-path
3. A removes F'; commit+path
4. E commits without proposal => path sets X, Y, W
5. A adds F; commit+path
6. A adds H; commit-path

Now suppose H as a new joiner wants to verify this tree.  The parent hash of U was last set at 5, at which point the tree looked like this:

~~~~~
              W
        ______|_____
       /             \
      U               Y [F]
    __|__           __|_
   /     \         /    \
  T       V       X [F] |
 / \     / \     / \    |
A   B   C   D   E   F   G 
~~~~~

So in your case `P=W`, `S=Y`, the "original tree hash" for Y needs to be the hash of the tree *without the parent node above G and H*.

In other words, the original tree hash reconstruction algorithm needs to truncate the tree in addition to replacing leaves with blanks.  From an implementation POV, that starts to sound quite a bit more challenging, since I'm now doing computing tree hashes over a tree I don't otherwise have.

It seems like there are a few ways forward here:

1. Do nothing; use the original sibling resolution.
2. Use the original tree hash, but truncate the tree for rightmost-added unmerged leaves.
3. Use the original tree hash, but update so that the tree always expands by doubling, instead of adding one leaf at a time.  This allows the leaf-blanking algorithm to work, at the expense of some other inefficiencies.
4. Use the tree hash, but remove the notion of unmerged leaves altogether (so there's no "original" tree hash).

Among those, my favorite option might actually be (4).  At this point, it kind of seems like unmerged leaves are more trouble than they're worth.  Their best-case efficiencies arise only in kinda weird cases, and they're making it hard to get good guarantees.

@raphaelrobert @kkohbrok thoughts?

kkohbrok:
I agree that an "upgrade" to using the tree hash rather than the public key would be a good idea. I'm also leaning towards (4), but I'm not convinced that unmerged leaves are more trouble than they're worth. But at this point, I don't have any data on how unmerged leaves impact performance and I also don't have a better idea, so I'd be Ok with (4) (although it would also be interesting to see what the truncation algorithm for (2) would look like).

raphaelrobert:
Thanks @TWal! I've been trying to wrap my head around this, because it is a major change with a lot of subtle nuances.
I think it's important to differentiate the concerns:

*Security* 

You mention that "a malicious member can tamper with the credentials before sending a Welcome package to a new member, even if it cannot tamper with the public keys". This makes it sound like a bigger problem than it is (unless I missed something). A malicious member can only tamper with credentials if it also has the corresponding private key material to do so. If that's the case, a wealth of new attacks become possible because authentication is obviously broken. There might be some more subtle attacks where a malicious member could do cross-group attacks without actually changing the credential. For example, the malicious member could pretend that member X is part of the new group, when member X is really only part of a totally different group. This would trick the new joiner Y into believing two things:

a) Member X is a member of the new group (this attack has always been possible with empty trees)
b) Member X has previously sent a full commit and therefore proven it's aware that it is in the group (this attack is currently possible)

Attack b) would become impossible by using the tree hash as you propose, and therefore it's obviously strictly more secure. If you thought of other types of attacks (that are maybe not so niche), it would be interesting to hear about it.

If the definition you propose is also motivated by better probability, this would be another reason to go for it.

*Efficiency*

This is harder to reasons about. As always, one has to make a number of assumptions about concrete scenarios. Dropping unmerged leaves would certainly make a lot of things easier and cleaner, but so far I've always assumed that they make things more efficient. My reasoning was the following: Without unmerged leaves, adding multiple new members in different locations of the tree (wherever the leftmost blanks are) would quickly blank most parent nodes in the tree and we'd go back to a linear effort. It's been mentioned that unmerged leaves can also be slower when there is a lot of them, however I find that an unlikely scenario to occur. You only get a lot of unmerged leaves if the tree is densely populated, which in turn means that members issue full commits frequently. Because of the latter, it is unlikely that the new members (that caused the unmerged leaves to appear) would not update just as frequently. I know this is a bit hand wavy, I guess running MC simulations in different scenarios would help.

With all that, I'm wondering if the current parent hash definition couldn't be extended by covering the whole key packages at the leaf level instead of just the keys (by modifying the `ParentHashInput`)? Maybe I'm missing something, but I think it would just be a small change in the spec that should address the security concern and not affect the efficiency at all.

TWal:
@bifurcation Thanks for noticing the bug, indeed since `Add`s can make the tree grow, we need to truncate it when checking the parent hash.
There is two things to notice for option (2):
- we only have to do these complex computations when receiving a Welcome message, not when processing a commit
- truncating the tree doesn't affect the O(n log log n) complexity, since computing the tree hash of a subtree can still be done in linear time
Option (4) is appealing because reasoning about unmerged leaves is surprisingly hard. I would be interested in an analysis on the performance impact of removing them (we could easily do an empirical analysis by doing random changes to a group and compare the size of the resolution in the trees with and without unmerged leaves, however I don't know what realistic changes to a group would look like).
If we expect new members to quickly commit once they arrived in a group, this should not cause too many performance issues I guess.

@raphaelrobert 
To be honest, I didn't find of highly convincing attacks. However what motivates me for this change is that with the current parent hash, there is a some room for tree modification which makes my gut feeling tell me that if someone think hard enough about this a convincing attack can be discovered. If we include the sibling tree hash in the parent hash then the parent hash makes everything super rigid and I don't see any starting point for an attack where the parent hash would be the weak link.
(In fact we can tamper the tree when the last tree modification is an Add-only commit, because in every parent hash the added leaf is removed from the tree hash, so we can tamper its KeyPackage)

I'm not sure to understand your last paragraph, do you suggest to keep the resolution but include the whole `ParentNode` and `KeyPackage` instead of only their `hpke_public_key`?

About cross-group attacks: we can easily avoid such attacks by including the group id in the signature of key packages. In the current design, this means adding a mandatory-when-comitting-with-path extension (like the ParentHash extension).

raphaelrobert:
> @raphaelrobert To be honest, I didn't find of highly convincing attacks. However what motivates me for this change is that with the current parent hash, there is a some room for tree modification which makes my gut feeling tell me that if someone think hard enough about this a convincing attack can be discovered. If we include the sibling tree hash in the parent hash then the parent hash makes everything super rigid and I don't see any starting point for an attack where the parent hash would be the weak link. (In fact we can tamper the tree when the last tree modification is an Add-only commit, because in every parent hash the added leaf is removed from the tree hash, so we can tamper its KeyPackage)
> 
> I'm not sure to understand your last paragraph, do you suggest to keep the resolution but include the whole `ParentNode` and `KeyPackage` instead of only their `hpke_public_key`?

Thinking about it again, it would only partially solve the problem. The tree hash would be stronger indeed.

> About cross-group attacks: we can easily avoid such attacks by including the group id in the signature of key packages. In the current design, this means adding a mandatory-when-comitting-with-path extension (like the ParentHash extension).

I think that's a good idea! Given the group ID is immutable, new joiners should not have issues verifying the signatures.

TWal:
I fixed the "right edge to the tree problems", I believe the procedure is now correct. The naive O(n log n) algorithm to process a Welcome message still works, I believe the O(n log log n) algorithm can be adapted but that's not 100% trivial.

As asked yesterday, here is in more details the algorithm to process a Commit

# Complexity to process a Commit

I believe we can expect from any efficient implementation to cache the tree hash of every subtree of the ratchet tree, because if this is not cached, then computing the tree hash at each epoch is in O(n), which is slow.

We have to recompute some of these cached values when processing a Proposal or an UpdatePath. When doing so, O(log n) of these values change, and by computing them bottom-up, we can compute each of them in O(1) giving a O(log n) complexity in total.
Therefore, updating these cached values when processing a Commit with p proposals is in O(p log n) which is fast.

The parent hash of a node has to be updated when processing an UpdatePath. We notice that doing so, every node that need parent hash recomputation has no unmerged leaves. Therefore every parent hash can be computed in O(1) by using the cached sibling tree hash, and updating the parent hashes is therefore in O(log n).

kkohbrok:
Since this also came up, here's a situation that showcases the benefits from including credentials in the parent hash.

Alice is in a group with Bob and some others. Bob has already made a commit in the course of the group's history, thus and signing a parent hash extension in this leaf KeyPackage.

Alice now adds Charly to the group.

With the parent hash mechanism as-is, Charly can see that Bob has at some point made a commit and with his parent hash attested to the fact that a certain set of public keys were in the tree when Bob made the commit. However, Charly can not rule out the possibility that Alice has generated a bunch of credentials, and used them to sign KeyPackages that include the public keys that Bob attested to (although of course she does not own the corresponding private keys).

With the proposed parent hash mechanism, Charly can rule out this possibility, since Bob attested not only to the public keys, but also to the credentials in the leaf KeyPackages.

Some notes on this scenario (and parent hashes in general):
* Alice's ability to forge meaningful KeyPackages should be limited, since Charly can verify them with the AS. So while Alice might not be able to make it look like Bob is in a group of clients of Alice's choosing, Charly still has no way of asserting who Bob thinks is actually in the group.
* Including Credentials in the parent hash could limit the amount deniability that can be achieved by vanilla MLS, although, for example, an extension could in theory change the way the parent hash works.
* In both scenarios, Charly's ability to figure out what Bob thinks the group looks like is best-effort, since Bob's direct path might have been overwritten to any degree by others since his last (full) commit.
* Ways to get more attestation, disregarding deniability completely:
  * have Bob (in addition to parent hashes) put the group id, epoch and some form of the roster (e.g. full credentials or just hashes) in a KeyPackage extension upon commit
  * cut the time until a new group member learns about who is actually in the group on a best-effort basis: Demand that after processing a commit that contains an Add, a group member must send an (empty) application message as a way of attesting to the current state of the group

MartaMularczyk:
I was going through some examples to verify that the procedure is correct. I think I found two cases where it may be incorrect. @TWal could you check the examples?
1. What happens if some nodes in the subtree of the sibling (Y in your examples) have been truncated since the commit?
2.  See this file [parent_hash.txt](https://github.com/mlswg/mls-protocol/files/7927786/parent_hash.txt) (the trees get large)

TWal:
Hi Marta, thank you for checking this! Indeed there are correctness problems that I completely overlooked…

1. I'm not sure to understand the problem? The tree is truncated when there is a Remove, which will blank the nodes so the parent hash won't be checked?
2. This is a good counter-example.

I see several solutions:
- remove unmerged leaves as suggested by @bifurcation (doing some analysis beforehand would be useful, but my guess is that it's not too bad if we say that new members SHOULD do a full commit as soon as possible)
- next to the parent hash, store how many leaves contained the sibling tree when the tree hash was computed, like this you know which nodes were truncated or not (this option is the least invasive)
- modify the tree hash in a way that would correspond to doing the current tree hash in a complete tree, which would solve the problem since we have the property Complete(Truncate(tree)) == Complete(tree)

kkohbrok:
Yes, I think you got it right. Alice can make Charly think that they are in a group with Bob and because Bob only attested to the HPKE public keys, Alice can potentially mess around with the credentials and Charly has only the AS to protect her from that. Depending on the AS that can be more or less helpful. If you have a weak AS, Charly cannot really know if Bob attested to being in a group with these group members. With the tree hash variant, Charly could do that.

I think @psyoptix got it right by saying that the current parent hash covers the confidentiality guarantees, whereas the tree hash approach gets some authentication guarantees as well.

I don't have a strong opinion either way, I just wanted to make sure we agree on the gap we're discussing.

MartaMularczyk:
Let me add a bit to the discussion about security. I hope it clarifies a bit parent hashes in general.

In our security analysis with @psyoptix we used the parent hash mechanism (only) to prove that MLS provides the following security property. 

> In any epoch where the credentials (i.e. signature keys) of all group members are “good”, i.e., not leaked to or chosen by the adversary, then the messages sent in this epoch are confidential and authentic.

An example scenario: A malicious insider Alice invites honest Charlie to a fake group with honest Bob, then Charlie adds Dave and removes Alice. Since all members use good credentials, the messages they sends to  each other are confidential and authentic. Note that Bob may not be aware of the fake group with Charlie and Dave. 

This guarantee is provably achieved both by the current parent hash and by the tree hash. So to motivate the tree hash, we need to find a different security property, outside our model, which it adds.

@kkohbrok proposed such a guarantee. The way I understand it is that a member Charlie can deduce additional facts about the group. For instance, if Bob attested to the credential of Emma via his parent hash, then Charlie knows that Bob was in a group with Emma. (This doesn’t mean that Emma actually was in a group with Bob; she may have been added but never joined.)

I agree that this is an additional goal one can aim for. But a couple of comments:
- Such guarantees are relatively hard to explain to the end user. So do we actually want them?
- The current parent hash also allows to deduce such statements, but fewer of them. For example, if Emma attested to an HPKE key _pk_ she chose during her commit, _pk_ happens to be on Bob's copath and Bob attested to _pk_ via the parent hash, then Charlie can deduce that Bob was in a group with Emma (just like in the example above).
- MLS doesn’t expose to higher level apps any interface that allows them to check the additional facts deduced from the parent hash. The only to get these facts is to dig into the details of the ratchet tree. This is not at the abstraction level apps typically work on. So if we introduce tree hash, we should think of exposing an interface that allows to take advantage of it.

bifurcation:
Elaborating on the idea I had on the call today:

It seems like @TWal's idea of using the tree hash as the input to the parent hash gets complicated in the following situation:

* Let P be an intermediate node
* Let L and R be the left and right children of P, respectively
* Consider the case where we are computing the parent hash that will be stored at L
* `parent_hash = H(P, original_tree_hash(R))`
* If R has no `unmerged_leaves`, then `original_tree_hash(R) == tree_hash(R)`
* If R has `unmerged_leaves`, but the addition of the leaves did not change the structure of the tree (i.e., they went into existing blank leaves), `original_tree_hash(R)` can be recomputed just by blanking out those leaves and following the current tree structure.
* If R has `unmerged_leaves`, and some of those leaves changed the tree structure, you end up in the hard situation that @TWal has been trying to address

Removing `unmerged_leaves` entirely would eliminate the latter two cases.  What I'm wondering is whether we can eliminate only the last, challenging case.  I think the change would look something like the following:

* When adding a new member:
    * If the member is occupying an existing slot, proceed as currently specified
    * If the member occupies a new slot added to the right of the tree, blank their direct path instead of adding `unmerged_leaves`.
* When computing parent hash, instead of the original resolution of the non-updated child, include the "original tree hash" == tree hash with unmerged leaves set to blanks.

That way, you never end up in the difficult situation above, because you never have unmerged leaves when the tree structure below you has changed since the last parent_hash -- in fact, you don't have a parent node at all when the subtree structure has changed.

(I think there are some elegant algorithms for computing original tree hash at the same time, but no need to get into that optimization now.)

This will remove some of the efficiency of `unmerged_leaves`, but it affects at worst a log-size number of parent nodes per Commit.  That could still cause some big jumps in Commit size, if one of these "leading-edge" parent nodes has a mostly-empty subtree.  But maybe the circumstances here are constrained enough that it's worth the extra assurance.

kkohbrok:
@MartaMularczyk thanks for the example on the guarantees we get from the current parent hash scheme. I think those are already great and I'm not convinced we should jump through a lot of hoops to close what seems to be a relatively slight gap. But again, not a strong opinion.

One thought regarding the surfacing of any sort of security guarantee to the user, though. While it would be nice to communicate these guarantees to the user, not being able to do that doesn't mean the guarantees are not worth having in the first place. Whether the user realizes it or not, they are protected against certain types of attacks. If the parent hash doesn't check out, the Add will fail due to an invalid group state. The way I understand it, both parent hash approaches give use best-effort guarantees that may only hold for parts of the tree in any case, so we have that problem with the current parent hash version already. The user will always have to assume that the tree in a group they are joining is as trustworthy as the client that added them to the group.

psyoptix:
TL;DR : To justify adopting a new ("tree-hash" or "leaf-hash" based)  parent hash construction I'd hope for efficiency, simplicity and/or engineering arguments because I'm not (yet?) seeing the security ones (at least, under the premise we won't communicate the added security guarantees to users).

@kkohbrok I'm struggling a bit with what you wrote about un-communicated guarantees still being worth it. Abstractly stated: I get that users dont need to (understand or) be aware of the *conditions* under which a security guarantee holds. But I have a hard time justifying a *guarantee itself* that user wont be aware of; at least when its easy for the adversary to create honest looking executions where the guarantee doesn't hold.

Concretely for this PR: Suppose we implement "tree-hash parent hash". The new guarantees we get (relative to "resolution parent hash") form some space of guarantees. A point in the space might be "C attested to being in the group with D at some point. A more recently attested to being in the group with B, C and D." or something along those lines...

Here's what bothers me if we don't communicate to users where they are in this space. At one end of the space (corresponding to ratchet trees looking like a freshly created group with few, if any, commits beyond group creation) we get diminishing (if any) new guarantees over resolution parent hash. Moreover, such states are both common in honest executions and the easiest for an adversary to create. So my question is: why would an adversary bother trying to mount an attack using a ratchet tree encoding guarantees deeper into the space if A) building such a ratchet tree is never easier and likely harder B) the extra guarantees can't make the attack easier but might make it harder yet C) the user can't tell the difference from being given a ratchet tree with fewer guarantees? 

Yet, if the answer to the question is "the adversary wouldnt bother" then I wonder what the point is a mechanism to provide said guarantees.

psyoptix:
If we had some language to talk about the guarantees encoded in a ratchet-tree T that is:
 - clean, succinct & simple
 - communicates significant and meaningful amount of information about the guarantees
 - clearly conveys an intuition that accurately reflects (parts of) the formal guarantee in T
then I'd probably change my mind about the security benefits of the new parent hash (assuming we also give MLS an interface for applications to extract statements in this language about the current epoch; especially for welcome msgs).

kkohbrok:
@psyoptix good points. If there is indeed an easier-to-create ratchet tree that would be equally plausible (as is the case here), then I agree that the guarantees are not worth much. In principle, however, I still think that guarantees might be worth having even if the user isn't aware of it. One example could be that the application could consider the effective guarantees that we get from a ratchet tree and then, for example, hide the group from the user until enough members have updated, such that group membership was sufficiently attested to (for some degree of "sufficiently"). Although for usability purposes that might not be an option, so in this particular point, it probably means that it's not worth having if we can't communicate it to the user.

Regarding the interface: I don't believe MLS as specified right now has any sort of interface. It's really only specified how clients communicate with one-another. They way I understand it, an MLS implementation can then decide what information about their state they make available to the application or what actions the application can take.

Regarding language: I'm not sure I understand what you have in mind, but let's give it a try.

# Security Model

Let T be a ratchet tree freshly received by user U. Then let M_T be an array in the state of user U indexed by the members of the tree T, where for each member m, M_T[m] contains the set of identities attested to by m. We can now create an oracle that allows the adversary to query this store. 

In the real world, the store is filled by checking which leaves are covered by the individual parent hashes present in the member's key packages. In the ideal world, each sender maintains a log (per tree) containing the members to which they attest and the receiver just copies the information from there upon receiving a ratchet tree. (Similar to log-decryption in real-vs.-ideal AE models.)

# Resolution Based Parent Hash

I would now claim that with resolution based parent hashes, there is an attack that goes as follows:

The adversary takes a valid tree, where a user A has already made a commit. The adversary generates a credential and replaces the credential of one of the users attested to by A (let's call the user B), before re-signing the corresponding key package. The adversary then sends it to their victim V as part of a Welcome message. V validates the parent hashes, which should be fine, since we didn't touch any hpke keys. The adversary then queries the oracle defined above using the gid of the new group, as well as the cids of V and A. If the user B shows up in M_T[A] then the adversary knows it's the ideal world, otherwise it's the real world.

# Tree-Hash Based Parent Hash 

If we have tree-hash based parent hash, the attack described above will not work. This is because upon receiving the Welcome message, V will fail to verify the parent hashes since the tree hash covers the credential as well.

# Caveat: AS

The above attack will _not_ work in the first scenario if the AS (which V will consult upon receiving the Welcome) rejects the tampered-with key package (or the credential within) created by the adversary. I still think there is some relevance to the attack, since for many messengers in use today, the AS essentially requires users to verify identities OOB ("safety numbers", etc.), which I would claim is not done in most cases.

TWal:
When working through the proof, I came up with another interesting pathological case, which has nothing to do with unmerged leaves:
```
_____________
             \
              X
        ______|______
       /             \
      V               _
    __|__           __|__
   /     \         /     \
  U       _       _       _
 / \     / \     / \     / \
A   B   C   _   _   _   _   H
```
If A did the last commit in this subtree, and someone in the left part of the tree (hidden) removes H, the tree will be truncated up to C so the sibling tree hash will change for V.

Hopefully the fixes I proposed earlier work against this pathological case.

TWal:
I am trying to abstract a bit how the parent hash is constructed, and when it is correct (i.e. participants processing a Welcome will accept a non-tampered tree).

# A proof framework for parent hash

We will consider separately the following operations on the tree:
- Blanking internal nodes along a path (and optionally the leaf): Update and first phase of Remove
- UpdatePath: full commit
- Adding a leaf in a blank leaf (add-only commit, add the leaf in the unmerged_leaves list along a path): second phase of Add
- Truncation: second phase of Remove
- Extension: first part of Add

The last three operations (adding a leaf, truncation, extension) will be called non-updating operations (because they don't update the `parent_hash` values in the nodes of the tree).

All parent hash design we came up so far revolve around two functions:
- a function `compute_sibling_data(sibling)` that is used when processing a Commit
- a function `compute_original_sibling_data(parent.content, sibling)` that is used when processing a Welcome
For the result to be meaningful, to use these functions we require that `sibling` is a descendant of the node `parent`.
(I'm using an abuse of notation, the function should be `compute_original_sibling_data(parent, sibling)` but I'm using `parent.content` instead of `parent` to say that only the content of the parent is inspected.)

On these two functions, if we have the following properties:
1. after an UpdatePath goes through P with sibling S, `compute_sibling_data(S) == compute_original_sibling_data(P.content, S)` (with P and S corresponding the values after the UpdatePath)
2. when a non-updating operation is applied to the tree, if P is modified then `compute_original_sibling_data(P_old.content, S_old) == compute_orginal_sibling_data(P_new.content, S_new)`

With these properties, we can prove the following theorem: if the parent hash verification procedure passes before applying an operation, then it passes after applying the operation.

The parent hash verification procedure works by computing an "expected parent hash" and "computed parent hash" for each node. The computed parent hash uses the `compute_original_sibling_data` function, and the expected parent hash is retrieved in a descendent of the node.

## Blanking internal nodes along a path

### Modification on the tree

Given a non-blank leaf L, every parent node on the path from L to the root becomes blank. L can either become blank (first phase of Remove) or get a new non-blank value (Update).

### Correctness argument

The key property for this operation is: for all internal node N, if N's subtree changes then N.content becomes blank.
From this we deduce that if the parent hash verification procedure validated the tree before the operation, it will validate the tree after the operation.
This is because for all non-blank nodes in the new tree, their subtree is not changed by this procedure.

## UpdatePath

### Modification on the tree

Given a non-blank leaf L, L get a new non-blank value, every node N on the path from L to the root are modified the following way:
- `N.content.unmerged_leaves` becomes empty
- `N.content.public_key` gets a new value
- `N.content.parent_hash` is updated with its parent parent-hash using `compute_sibling_data(N.sibling)`

### Correctness argument

The key property for this operation is: for all internal node N, if N's subtree changes then N is on the path (with child C and sibling S), N.content becomes non-blank, and C.parent_hash is computed using `compute_sibling_data(S)`.
Using this key property and property 1, we can deduce that for all node N, either N's subtree is unchanged, or N's expected parent hash and computed parent hash are identical.
Therefore the parent hash verification procedure still verifies the new tree.

## Adding an unmerged leaf in a blank leaf

### Modification on the tree

Given a blank leaf L, L gets a new non-blank value, and L is added to `N.content.unmerged_leaves` for every parent node N in the path from L to the root.

### Correctness argument

For all node N, the computed parent hash don't change using property 2. The expected parent hash is not changed by this operation.

## Truncating the tree

### Modification on the tree

The rightmost leaf and its parent are removed. As described in the section Remove, it happens only if the leaf is blank and its parent is either blank or is the root.

### Correctness argument

For all node N, the computed parent hash don't change using property 2.
With the "careful truncation", we can prove that the expected parent hash value do not change.
Two nodes are removed: the rightmost leaf R and its parent P.
The removal of the rightmost leaf don't change any expected parent hash since it is blank.
If P is the root then removing P don't change any expected parent hash.
Otherwise P is blank, therefore the only way for the removal of P to interfere with the computation of the expected parent hash for some node is if P is on the "left-blank path" in the "right node" case.
Removing P will simply shorten the "left-blank path" by one node and the expected parent hash will therefore stay the same.

## Extending the tree

### Modification on the tree

Add a blank leaf and a blank parent as described in the Adding and Removing Leaves section.

### Correctness argument

This operation adds one blank leaf L and one blank internal node P.
The key property for this operation is: P is either the root or P is a right node.
This is deduced by the fact that we manipulate left-balanced trees.

From this key property, we deduce that for each node, the expected parent hash value do not change:
- if P is the root then this is obvious
- if P in a right node, only the expected parent hash from P.parent can change, but actually it will stay the same thanks to the "left-blank path" procedure.
Using property 2, we deduce that for each node, the computed parent hash value do not change.

# Complexity

If `compute_original_sibling_data(parent.content, sibling)` has complexity O(|sibling|) then processing a Welcome for a tree with n leaves can be done in O(n log n).
Let P be a node, L and R be its children. If `compute_sibling_data(S)` can be computed in O(1) from `compute_sibling_data(L)`, `compute_sibling_data(R)` and `P.content` then processing a Commit can be done in O(log n).
Let P be a node, L and R be its children. If `compute_original_sibling_data` only uses `parent.content.unmerged_leaves` and `compute_original_sibling_data(excluded_leaves, P)` can be computed in O(1) using `compute_original_sibling_data(intersection(excluded_leaves, L.leaves), L)` and `compute_original_sibling_data(intersection(excluded_leaves, R.leaves), R)`, then processing a Welcome for a tree with n leaves can be done in O(n log log n).

# Actual procedures for `compute_original_sibling_data`

## The resolution approach

### Definition
The current implementation of parent hash uses:
- `compute_sibling_data(sibling) = resolution(sibling)`
- `compute_original_sibling_data(parent.content, sibling) = resolution(sibling) - hpke_keys(parent.content.unmerged_leaves)`

### Proof

We can easily prove property 1 and 2, therefore making the current parent hash correct.

### Complexity

The complexity to process a Welcome is O(n log n) and the complexity to process a Commit is in O(n).

## The full-extension + tree hash approach

### Definition

```
function full_extension(tree):
    extend `tree` until its number of leaves is a power of 2

function remove_leaves(tree, leaves):
    change the leaves `leaves` to a blank leaves, and remove `leaves` in every internal node's unmerged_leaves list
```

We define:
- `compute_sibling_data(sibling) = tree_hash(full_extension(sibling))`
- `compute_original_sibling_data(parent.content, sibling) = tree_hash(remove_leaves(full_extension(sibling), parent.content.unmerged_leaves))`

### Proofs

Property 1 is given by the fact that `remove_leaves(tree, []) == tree`.
For property 2:
- if a blank leaf is populated under `parent`, it is populated either in:
  - its `child` node, therefore the only modification in the pair `(parent.content, sibling)` is a leaf from `child` being added to `parent.content.unmerged_leaves`, and this doesn't change the value computed by `remove_leaves`
  - its `sibling` node, and in this case the addition of the leaf in `parent.content.unmerged_leaves` makes `remove_leaves` undo the changes done by the addition
- if a truncation or extension changes `parent`, this doesn't change `full_extension(sibling)` so `compute_original_sibling_data(parent.content, sibling)` stays the same.

### Complexity

We can use the fact that `remove_leaves(full_extension(tree), leaves) == full_extension(remove_leaves(tree, leaves))`.
The complexity to process a Welcome is in O(n log log n) and the complexity to process a Commit is in O(log n).

bifurcation:
Discussion on virtual interim:

* @TWal and @bifurcation to get to a merge-able state, then make decision on mailing list

draft-ietf-mls-protocol.md
```diff
@@ -2029,13 +2029,59 @@ is the root, then `parent_hash` is set to a zero-length octet string.
...
-D has an array with one element in it: the HPKE public key of D.
+Finally, `original_sibling_tree_hash` is the original tree hash of S. The
+original tree hash corresponds to the tree hash of S the last time P was
+updated. It can be computed as the tree hash of S in the ratchet tree modified
+the following way:
+
+* reset the leaves in P.unmerged_leaves to blanks
+* remove P.unmerged_leaves from all unmerged_leaves lists
+* truncate the ratchet tree as described in {{remove}}
+
```
bifurcation, line 2040:
```suggestion
Note that no recomputation is needed if the tree hash of S is unchanged since
the last time P was updated.  This is the case for computing or processing a
Commit whose UpdatePath traverses P, since the Commit itself resets P.  (In
other words, it is only necessary to recompute the original sibling tree hash
when validating group's tree on joining.) More generally, if none of the entries
in `P.unmerged_leaves` is in the subtree under S (and thus no nodes were truncated),
then the original tree hash at S is the tree hash of S in the current tree.

If it is necessary to recompute the original tree hash of a node, the efficiency
of recomputation can be improved by caching intermediate tree hashes, to avoid
recomputing over the subtree when the subtree is included in multiple parent
hashes.  A subtree hash can be reused as long as the intersection of the
parent's unmerged leaves with the subtree is the same as in the earlier
computation.
```

bifurcation (approved):
I suggested some text to point people toward more efficient implementations.  Other than that, I think this is good to go.

bifurcation (approved):

PR #529, opened 2021-12-17
Refactor Group Joining and fix External Inits

kkohbrok:
This PR refactors the two sections on Group Joining (via Welcome and via External Init) by changing the `GroupInfo` struct that can be used for both cases. The `external_pub` used in the course of an External Init is added as an optional field. This PR also:

- removes the redundant ProtocolVersion in the GroupInfo
- adds a GroupInfoTBS struct to follow the convention of specifying a TBS struct for signed messages
- implicitly, the `confirmed_transcript_hash` and `confirmation_tag` is now used for external inits instead of the `interim_transcript_hash`, which was not sufficient to create a valid commit

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3486,7 +3486,7 @@ places:
 
 * In KeyPackages, to describe client capabilities and aspects of their
   participation in the group (once in the ratchet tree)
-* In the Welcome message, to tell new members of a group what parameters are
+* In the GroupInfo, to tell new members of a group what parameters are
   being used by the group, and to provide any additional details required to
   join the group
 * In the GroupContext object, to ensure that all members of the group have the
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3134,7 +3134,6 @@ struct {
     Extension group_context_extensions<0..2^32-1>;
     Extension other_extensions<0..2^32-1>;
     MAC confirmation_tag;
-    optional<HPKEPublicKey> external_pub;
     KeyPackageRef signer;
 } GroupInfoTBS;
 ~~~
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3097,7 +3097,53 @@ MLSPlaintext.confirmation_tag =
     MAC(confirmation_key, GroupContext.confirmed_transcript_hash)
 ~~~~~
 
-### External Commits
+### Adding Members to the Group
+
+New members can join the group in two ways. Either by being added by a group
+member, or by adding themselves through an external Commit. In both cases, the
+new members need information to bootstrap their local group state.
+
+~~~
+struct {
+    ProtocolVersion version = mls10;
+    CipherSuite cipher_suite;
+    opaque group_id<0..255>;
+    uint64 epoch;
+    opaque tree_hash<0..255>;
+    opaque confirmed_transcript_hash<0..255>;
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -3127,48 +3110,78 @@ struct {
     opaque group_id<0..255>;
     uint64 epoch;
     opaque tree_hash<0..255>;
-    opaque interim_transcript_hash<0..255>;
+    opaque confirmed_transcript_hash<0..255>;
     Extension group_context_extensions<0..2^32-1>;
     Extension other_extensions<0..2^32-1>;
-    HPKEPublicKey external_pub;
+    MAC confirmation_tag;
+    optional<HPKEPublicKey> external_pub;
```
bifurcation, line 3117: I would probably make this an extension (in `other_extensions`, like `ratchet_tree`) instead of an optional field.

draft-ietf-mls-protocol.md
```diff
@@ -3127,48 +3110,78 @@ struct {
...
+* epoch ID
+* ciphersuite
+* public tree hash
+* confirmed transcript hash
+* confirmation tag of the most recent Commit
+* group extensions
+* external public key
+
+Thus to join a group via an External Commit, a new member needs a GroupInfo with
+an `external_pub` present.
```
bifurcation, line 3169: Suggest saying a little more about the dynamics here:

```suggestion
In other words, to join a group via an External Commit, a new member needs a GroupInfo with
an `external_pub` present.  Thus, a member of the group can enable new clients to join by
making a GroupInfo object available to them.  Note that because a GroupInfo object is specific 
to an epoch, it will need to be updated as the group advances.  In particular, each GroupInfo
object can be used for one external join, since that external join will cause the epoch to change.
```

bifurcation (approved):
So basically, we're converging the internal and external join cases as a result of fixing a bug in external join.  Seems like a fine idea.

I will note that this promotes GroupInfo into a "public" / "on-the-wire" object, as opposed to always being wrapped in a Welcome.  In the internal join case, the GroupInfo is always sent encrypted to the new joiner; in the external join case, it isn't necessarily encrypted (up to the application).  Doesn't need to be a blocker, but might be worth elaborating on in an implementation document.

Marking as approved, but I would pretty strongly prefer that `external_pub` be in an extension vs. an optional field.

kkohbrok:
Thanks for the suggestions! I agree that it's a bit nicer to have an extension rather then an optional field.

draft-ietf-mls-protocol.md
```diff
@@ -3097,7 +3097,49 @@ MLSPlaintext.confirmation_tag =
...
+struct {
+    CipherSuite cipher_suite;
+    opaque group_id<0..255>;
+    uint64 epoch;
+    opaque tree_hash<0..255>;
+    opaque confirmed_transcript_hash<0..255>;
+    Extension group_context_extensions<0..2^32-1>;
+    Extension other_extensions<0..2^32-1>;
+    MAC confirmation_tag;
+    optional<HPKEPublicKey> external_pub;
```
raphaelrobert, line 3137: Since it's now an extension, do we still need it here?

raphaelrobert:
I agree it's good to align `PublicGroupState` and `GroupInfo`!

draft-ietf-mls-protocol.md
kkohbrok: Ah, sorry. Forgot to remove that one. Thanks!

bifurcation (approved):

PR #532, opened 2021-12-27
Get rid of explicit protocol version fields

kkohbrok:
In all instances that I could find in the spec, the `ProtocolVersion` is redundant, as it appears alongside the `ciphersuite, which implies an MLS version. This PR thus removes explicit handling of MLS versions outside of ciphersuite concerns.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1290,12 +1290,6 @@ The input to the signature computation comprises all of the fields
 except for the signature field.
 
 ~~~~~
-enum {
-    reserved(0),
-    mls10(1),
-    (255)
-} ProtocolVersion;
-
 // See IANA registry for registered values
 uint16 ExtensionType;
 
@@ -1305,7 +1299,6 @@ struct {
 } Extension;
 
 struct {
-    ProtocolVersion version;
     CipherSuite cipher_suite;
     HPKEPublicKey hpke_init_key;
     opaque endpoint_id<0..255>;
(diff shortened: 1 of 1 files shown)


ekr:
This seems like a bad idea to me. We may well want to change ciphers without changing protocols or vice versa.

bifurcation (changes requested):
Yeah, I think the actual error here is that we tied ciphersuites to version in the first place.  I seem to recall objecting at the time, but relenting because it was a soft association -- having a separate version number meant that even if a ciphersuite was allegedly tied to a given version, you could have a new version adopt it.  So I would object to making this a hard association.

By way of comparison: TLS went through three iterations over 20 years without having to redefine all of its ciphersuites, which is a good thing because there are 350 entries!  TLS 1.3 made a bit of a cut-over, but that still doesn't argue for a hard single-version association.  The pre-1.3 suites work with earlier versions, and the 1.3 suites are expected to be good for the next few versions.

So if we do anything here, we should strip the version name from the ciphersuite name, and maybe add a column to the registry to indicate what version the suite can be used with.

kkohbrok:
I don't have a strong opinion either way, but the way it currently is doesn't makes the version fields redundant. @bifurcation do you want to make a PR with a scheme that detaches the ciphersuites from the versions? Although that's probably a discussion we should have on the mailing list as well.

raphaelrobert:
I agree that we should really choose between:

a) Having protocol version-specific ciphersuites and not repeat the protocol version (note that ciphersuites could still be changed for a given protocol version)
b) Not have protocol version-specific ciphersuites and mention the protocol version wherever needed

IIRC we discussed this at the interim in 2020 in NYC and the reason for deviating from what TLS does were the following:

 - MLS is unlikely to ever have as many ciphersuites as TLS
 - The meaning/set of the crypto primitives mentioned in the ciphersuite might considerably change between MLS versions. If the primitives are used differently in a newer version (their meaning changes), it might be a security issue (e.g. a certain security level is now no longer considered sufficient). If the set changes, ciphersuites can't be carried over between versions anyway.

I'm not very hung up on this, but we did have reasons to do what we did at the time.

bifurcation:
As discussed on working call 2022-01-20, closing in favor of #544

PR #535, opened 2022-01-10
Require extensions to be unique in any given extensions vector

### LINKED ISSUE #530: Require uniqueness of RatchetTreeExtensions in GroupInfo
If a GroupInfo (or a PublicGroupState) should only ever contain one RatchetTreeExtension. Either that or it should be specified what happens if there is more than one.

kkohbrok:
Fixes #530.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3422,11 +3422,11 @@ def subtree_root(nodes):
 ~~~~~
 
 (Note that this is the same ordering of nodes as in the array-based tree representation
-described in {{array-based-trees}}.  The algorithms in that section may be used to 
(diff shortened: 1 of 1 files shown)


raphaelrobert (approved):

draft-ietf-mls-protocol.md
```diff
@@ -3518,6 +3518,8 @@ handle extensible fields:
 * The extensions populated into a GroupContext object are drawn from those in
   the GroupInfo object, according to the definitions of those extensions.
 
+* Any extension field MUST NOT have more than one extension of any given type.
```
bifurcation, line 3521:
```suggestion
* Any field containing a list of extensions MUST NOT have more than one extension of any given type.
```

PR #541, opened 2022-01-14
Make some sections easier to read.

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -1542,11 +1537,11 @@ Different changes to the group will have different effects on the group state.
 These effects are described in their respective subsections of {{proposals}}.
 The following general rules apply:
 
-* The `group_id` field is constant
+* The `group_id` and `extensions` fields are constant.
```
bifurcation, line 1540:
```suggestion
* The `group_id` field is constant.
* The `extensions` field is updated whenever a GroupContextExtensions proposal is committed.
```

(And the `extensions` part should go lower down.)

bifurcation (approved):
Overall lgtm, one small comment.

draft-ietf-mls-protocol.md
bifurcation: Will handle this in a small follow-on.

PR #543, opened 2022-01-19
Be consistent about version labeling

### LINKED ISSUE #531: MLS version is used inconsistently in strings
Both "mls10" and "MLS 1.0" are used. I'd go for "mls10" everywhere.

bifurcation:
Fixes #531


bifurcation:
@TWal suggests moving everything to `MLS 1.0` instead.  Will do.

PR #544, opened 2022-01-19
Remove version from ciphersuites

bifurcation:
#532 proposes to remove redundancy between version fields and ciphersuites by removing the version fields.  This PR removes the version from the ciphersuite instead.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3520,6 +3520,46 @@ using the tree for MLS operations.
 
 # Extensibility
 
+The base MLS protocol can be extended in a few ways.  New ciphersuites can be
+added to enable the use of new cryptographic algorithms.  New types of proposals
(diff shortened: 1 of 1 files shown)


bifurcation:
@raphaelrobert suggests adding some advisory text in the extensibility section to the effect that future versions of the protocol should consider whether they want to keep all defined ciphersuites.

seanturner:
@raphaelrobert here is some text we put in RFC 8447 (IANA registries for TLS) that was also included in the IANA registry to warn about cipher suites getting weaker over time. This could be modified to mention culling at different version.

    WARNING: : Cryptographic algorithms and parameters will be
    broken or weakened over time. Blindly implementing cipher
    suites listed here is not advised. Implementers and users
    need to check that the cryptographic algorithms listed continue
    to provide the expected level of security.

raphaelrobert (approved):

PR #546, opened 2022-01-20
Downgrade MUST to SHOULD for commit senders including all valid commits

### LINKED ISSUE #542: Conflicting proposal / commit behavior
There are several rules in terms of selecting proposals and issuing commits against them. I have distilled the existing wording in the document into the following rules

1. If there are multiple updates for the same leaf, cherry pick one.
2. If there is a remove and an update/add for the same key package, always choose remove
3. If there are multiple group context extension updates, cherry pick one
4. The sender of a commit MUST use all valid proposals it has received 
5. In the event that a valid proposal is omitted from the next Commit, the sender of the proposal SHOULD retransmit it in the new epoch

It seems that rule 4 and rule 5 conflict with each other in that one is suggesting that all proposals must be consumed or there should be a hard error, and the other assumes it is possible for a commit to purposefully leave out a proposal and it can optionally be retransmitted.

I have been trying to write code that reasonably handles the cherry picking logic around multiple proposals referencing the same key package or multiple changes to group context while also enforcing the "must use all valid proposals received" rule, and it does introduce some complexity that maybe would be avoided by just eliminating that rule?

tomleavy:
Addresses the comments in ( Resolves #542 )

Since a member performing a commit including all valid proposals within an epoch is best effort, it makes sense to downgrade the requirement from MUST to SHOULD. 

IMO, MUST should be reserved for instances where the receiver would throw an error in the condition the sender did not 100% comply with the rule. Since in this case it is not a hard error under certain conditions, such as the member performing the commit having not received a proposal in transit while generating the commit, SHOULD is a better choice.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2858,8 +2858,9 @@ key with a Credential in any leaf of the group, or indeed if the KeyPackage
 shares the same `hpke_init_key` with another KeyPackage in the group.
 
 The Commit MUST NOT combine proposals sent within different epochs. In the event
-that a valid proposal is omitted from the next Commit, the sender of the
-proposal SHOULD retransmit it in the new epoch.
+that a valid proposal is omitted from the next Commit, and that proposal is
+still valid in the current epoch, the sender of the proposal MAY retransmit
+it.
 
 A member of the group MAY send a Commit that references no proposals at all,
 which would thus have an empty `proposals` vector.  Such
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2833,7 +2833,7 @@ a Commit message before sending application data. This ensures, for example,
 that any members whose removal was proposed during the epoch are actually
 removed before any application data is transmitted.
 
-The sender of a Commit SHOULD include all valid proposals that it has received
+The sender of a Commit MUST include all valid proposals that it has received
 during the current epoch. Invalid proposals include, for example, proposals with
 an invalid signature or proposals that are semantically invalid, such as an Add
 when the sender does not have the application-level permission to add new users.
@@ -2857,10 +2857,12 @@ proposal if the Credential in the contained KeyPackage shares the same signature
 key with a Credential in any leaf of the group, or indeed if the KeyPackage
 shares the same `hpke_init_key` with another KeyPackage in the group.
 
-The Commit MUST NOT combine proposals sent within different epochs. In the event
-that a valid proposal is omitted from the next Commit, and that proposal is
-still valid in the current epoch, the sender of the proposal MAY retransmit
-it.
+The Commit MUST NOT combine proposals sent within different epochs. Due to the
+asynchronous nature of proposals, receivers of a Commit SHOULD NOT enforce that
+all valid proposals sent within the current epoch are referenced by the next
+Commit. In the event that a valid proposal is omitted from the next Commit, and
+that proposal is still valid in the current epoch, the sender of the proposal
+MAY retransmit it.
 
 A member of the group MAY send a Commit that references no proposals at all,
 which would thus have an empty `proposals` vector.  Such
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2833,7 +2833,7 @@ a Commit message before sending application data. This ensures, for example,
 that any members whose removal was proposed during the epoch are actually
 removed before any application data is transmitted.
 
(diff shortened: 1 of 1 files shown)


bifurcation (approved):

bifurcation:
Re: non-redundant: I'm not sure we need to say "valid non-redundant".  We can just include non-redundancy in the definition of "valid".

Re: retransmit: I would frame this as "If a group member sent a proposal within an epoch that was not included in a Commit that initiates the next epoch, then the proposer MAY retransmit the proposal in the new epoch."

tomleavy:
I believe the existing language in the "if there are multiple proposals ..." paragraph covers the non-redundant aspect of the requirements.

How about this for retransmit? I can push a new commit if we think this helps

"The Commit MUST NOT combine proposals sent within different epochs. In the event that a valid proposal is omitted from the next Commit, and that proposal is still valid in the current epoch, the sender of the proposal MAY retransmit it."

raphaelrobert:
I think that works.

tomleavy:
Pushed a commit to add the copy in my comment

draft-ietf-mls-protocol.md
```diff
@@ -2833,7 +2833,7 @@ a Commit message before sending application data. This ensures, for example,
 that any members whose removal was proposed during the epoch are actually
 removed before any application data is transmitted.
 
-The sender of a Commit MUST include all valid proposals that it has received
+The sender of a Commit SHOULD include all valid proposals that it has received
```
Bren2010, line 2836: I don't really agree with downgrading this from a MUST to a SHOULD. Quoting RFC 2119:

```
Imperatives of the type defined in this memo must be used with care
and sparingly.  In particular, they MUST only be used where it is
actually required for interoperation or to limit behavior which has
potential for causing harm (e.g., limiting retransmisssions)
```

This is a case where we're limiting potentially harmful behavior (intentionally dropping proposals and forcing others to retransmit) such that implementations that don't follow the MUST are non-compliant with the spec.

tomleavy: I see your point, my primary concern was reducing how often commits are rejected due to the async nature of the protocol since those packets likely have a higher computational and bandwidth cost.

"A group member that has observed one or more proposals within an epoch MUST send a Commit message" also plays into this choice. If a group is receiving an uptick in proposals, which tends to happen when they are first created and a large number of users in size, then commits will fail and block members from sending messages until a new commit is generated and uploaded. When you are using the protocol on something like a mobile app on a bad connection this becomes a problem due to high amounts of latency between message send and receive.

Bren2010: Commits wouldn't be rejected if they don't have 100% of the proposals. There are valid reasons a Commit would not have all proposals: lossy delivery channel, race condition. So like in the case where 1.) I start preparing a Commit, 2.) You send a new proposal, 3.) I send my Commit without your proposal -- your proposal would be dropped and you'd have the option to re-send it in the new epoch. There's not a technical enforcement of this "MUST", its more about policy.

I believe a Commit would only be rejected if it was sent at the same time as another Commit, and the Delivery Service chose to use the other Commit instead. That would be less common because it would only happen when two users are trying to send application data at the same time.

bifurcation: This is why I think SHOULD is appropriate here -- I hesitate to have a MUST-level requirement for behavior that is not observable to remote parties.  After all, if it's not observable, then it can hardly be a requirement for interop!

rohan-wire: In my experience it is totally reasonable to split behavior into sending/receiving roles for example. It's fine to say that when you send a commit you MUST include all valid proposals you have received, as long as the receiving behavior makes it clear what to do (and under what circumstances the receiver might see a proposal not in the commit: ex: invalid, or hadn't arrived yet).

If we make this a SHOULD, we need an extra sentence explaining under what circumstances the sender might choose not to include a specific proposal. SHOULD's are always more complicated. My general IETF advice is to use them sparingly.

bifurcation: I can live with that.  Then maybe the way to resolve the confusion @tomleavy notes is to add a note here that receivers SHOULD NOT enforce that everything they sent is included?

bifurcation:
Discussion on working call:

* Agreement to soften receiver requirement, keep MUST on sender
* @tomleavy to update, then we can merge

tomleavy:
Just posted new wording, let me know what you think

Bren2010 (approved):

bifurcation (approved):

bifurcation:
Looks good to me, thanks @tomleavy

PR #547, opened 2022-01-21
Domain separation for KeyPackage and Proposal references

kkohbrok:
It doesn't seem super critical, but I think right now there could be a collision between references to Update/Add proposals and KeyPackages. Also, it's essentially free and it just feels wrong not to have proper domain separation for these values 😬.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1175,7 +1175,8 @@ to Proposals they cover.  These identifiers are computed as follows:
 ~~~~~
 opaque HashReference[16];
 
-MakeHashRef(value) = KDF.expand(KDF.extract("", value), "MLS 1.0 ref", 16)
+MakeKeyPackageRef(value) = KDF.expand(KDF.extract("", value), "MLS 1.0 KeyPackage Reference", 16)
+MakeProposalRef(value) = KDF.expand(KDF.extract("", value), "MLS 1.0 Proposal Reference", 16)
 
 HashReference KeyPackageRef;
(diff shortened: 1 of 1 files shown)


Bren2010 (approved):

TWal:
I also thought about this but didn't see what are the security implications.

In any case, that's always good hygiene and doesn't cost much!

kkohbrok:
Yeah, no particular security implications. It just seems right to do it this way and it doesn't seem to hurt.

draft-ietf-mls-protocol.md
```diff
@@ -1175,7 +1175,8 @@ to Proposals they cover.  These identifiers are computed as follows:
 ~~~~~
 opaque HashReference[16];
 
-MakeHashRef(value) = KDF.expand(KDF.extract("", value), "MLS 1.0 ref", 16)
+MakeKeyPackageRef(value) = KDF.expand(KDF.extract("", value), "MLS 1.0 KeyPackage Reference", 16)
+MakeProposalRef(value) = KDF.expand(KDF.extract("", value), "MLS 1.0 Proposal Reference", 16)
```
bifurcation, line 1179:
```suggestion
MakeKeyPackageRef(value) = KDF.expand(
  KDF.extract("", value), "MLS 1.0 KeyPackage Reference", 16)

MakeProposalRef(value) = KDF.expand(
  KDF.extract("", value), "MLS 1.0 Proposal Reference", 16)
```

Wrapping to fit within RFC Editor's preferred line lengths.

bifurcation (approved):

PR #549, opened 2022-01-26
Add branch/init overview

bifurcation:
Adds the overview text proposed in #525.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -691,6 +691,88 @@ A              B     ...      Z          Directory       Channel
 ~~~~~
 {: title="Client Z removes client B from the group"}
 
+## Relationships Between Epochs
+
+A group comprises a single linear sequence of epochs. However, it can sometimes
+be useful to express relationships between epochs, either within a group or
+across groups.  MLS derives a resumption pre-shared key (PSK) from each epoch to
+allow entropy extracted from one epoch to be injected into a future epoch,
+effectively creating a causal link between the two epochs.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -691,6 +691,88 @@ A              B     ...      Z          Directory       Channel
 ~~~~~
 {: title="Client Z removes client B from the group"}
 
+## Relationships Between Epochs
+
+A group comprises a single linear sequence of epochs. However, it can sometimes
+be useful to express relationships between epochs, either within a group or
+across groups.  MLS derives a resumption pre-shared key (PSK) from each epoch to
+allow entropy extracted from one epoch to be injected into a future epoch,
+effectively creating a causal link between the two epochs.
```
kkohbrok, line 700: I feel like it doesn't become clear here, what the actual benefits of this are. Although the full details should probably go into the architecture docs. Something as follows maybe?

```suggestion
A group comprises a single linear sequence of epochs and groups are generally
independent of one-another. However, it can sometimes be useful to link epochs
cryptographically, either within a group or across groups. MLS derives a
resumption pre-shared key (PSK) from each epoch to allow entropy extracted from
one epoch to be injected into a future epoch. This link guarantees that members
entering the new epoch agree on a key if and only if were members of the group
during the epoch from which the resumption key was extracted.
```

kkohbrok (changes requested):
Thanks for putting together the overview. The figures are great :-)

Looks good to me with one proposition.

bifurcation:
Thanks @kkohbrok !

PR #554, opened 2022-01-26
Use common-sense definition of _descendant_

rohan-wire:
Make the definition of _descendant_ match the common-sense definition (you are not your own descendant), adjust the text which uses the definition. Note: this allows for a much clearer definition of the tree invariant as well.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -717,13 +717,13 @@ _leaf_ if it has no children, and a _parent_ otherwise; note that all
 parents in our trees have precisely
 two children, a _left_ child and a _right_ child. A node is the _root_
 of a tree if it has no parents, and _intermediate_ if it has both
-children and parents. The _descendants_ of a node are that node, its
+children and parents. The _descendants_ of a node are that node's
 children, and the descendants of its children, and we say a tree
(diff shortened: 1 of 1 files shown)


Bren2010 (approved):

draft-ietf-mls-protocol.md
```diff
@@ -717,13 +717,13 @@ _leaf_ if it has no children, and a _parent_ otherwise; note that all
 parents in our trees have precisely
 two children, a _left_ child and a _right_ child. A node is the _root_
 of a tree if it has no parents, and _intermediate_ if it has both
-children and parents. The _descendants_ of a node are that node, its
+children and parents. The _descendants_ of a node are that node's
 children, and the descendants of its children, and we say a tree
-_contains_ a node if that node is a descendant of the root of the
-tree. Nodes are _siblings_ if they share the same parent.
+_contains_ a node if that node is a descendant of the root of, or the root 
+of, the tree. Nodes are _siblings_ if they share the same parent.
```
bifurcation, line 723:
```suggestion
_contains_ a node if that node is a descendant of the root of the tree,
or if the node itself is the root of the tree. Nodes are _siblings_ if they share the same parent.
```

bifurcation (approved):

PR #555, opened 2022-01-26
Add clarifying note about _resolution_ in the most common case.

rohan-wire:
In the most basic case, the resolution of a node is the node, so let's say that.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -810,7 +810,7 @@ present are laid out in {{views}}.
 A node in the tree may also be _blank_, indicating that no value is present at
 that node.  The _resolution_ of a node is an ordered list of non-blank nodes
 that collectively cover all non-blank descendants of the node.  The resolution
-of a non-blank node with no unmerged leaves is just the node itself. The resolution
+of a non-blank node with no unmerged leaves is just the node itself. More generally, the resolution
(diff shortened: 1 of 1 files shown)


Bren2010:
Fwiw this is stated immediately below your note:
```
* The resolution of a non-blank node comprises the node itself,
followed by its list of unmerged leaves, if any
```

rohan-wire:
Bren2010 said:
>Fwiw this is stated immediately below your note:

Yes. And it took me three readings to understand that the most complicated scenario was being described first and the most basic scenario is not explicitly stated. Most readers will not understand in what cases there will be unmerged leaves by this point.

draft-ietf-mls-protocol.md
```diff
@@ -810,6 +810,7 @@ present are laid out in {{views}}.
 A node in the tree may also be _blank_, indicating that no value is present at
 that node.  The _resolution_ of a node is an ordered list of non-blank nodes
 that collectively cover all non-blank descendants of the node.  The resolution
+of a non-blank node with no unmerged leaves is just the node itself. The resolution
```
bifurcation, line 813:
```suggestion
of a non-blank node with no unmerged leaves is just the node itself. More generally, the resolution
```

bifurcation (approved):
Looks good with one minor edit.

PR #560, opened 2022-01-27
Exempt default features from RequiredCapabilities

### LINKED ISSUE #537: Clarity around extensions and `RequiredCapabilities`
The description of `RequiredCapabilities` is clear that proposals defined in the RFC are not to be included because they are considered standard, however there is no such clarification around extensions. 

In general, extension requirements are not all that clear because in several places (such as Lifetime) it is mentioned that an extension must be included, however they are all listed as recommended rather than required in the table. It also raises a question in that if an extension is required to be implemented as part of the specification, and required in every KeyPackage, should it just be a field rather than an extension?

bifurcation:
Fixes #537

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2557,12 +2557,16 @@ struct {
 ~~~~~
 
 This extension lists the extensions and proposal types that must be supported by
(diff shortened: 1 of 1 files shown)


bifurcation:
cc @tomleavy

tomleavy:
Looks good to me, thanks!

PR #562, opened 2022-01-27
Forbid proposals by reference in external commits

### LINKED ISSUE #540: Proposals by reference in external commit
The spec currently doesn't distinguish much between an external and a member commit w.r.t. how proposals by reference are handled. In theory, the server could keep track of proposals and send the external committer all proposals that were sent up until the time they want to commit externally. However, there are a few difficulties in committing proposals by reference in an external commit.

* Membership tags: The external committer doesn't have the ability to verify membership tags on proposals. This is probably not a show-stopper, as proposals still need to be signed by a group member, but if nothing else, this should be noted in the spec.
* Leaf positioning: The spec was changed such that the committer is added into the tree without an Add proposal being present. If there are other Add (and Remove) proposals present, it is unclear in what order they should be processed. Should the committer be added when the path is processed? Or should it still be processed as if it were an Add?

There are probably other things that can go wrong when committing referenced proposals externally, but those are the first that come to mind.

The simplest solution would be to simply ignore proposals by reference, although this would break the current principle that a commit must always include all valid proposals that were sent during the epoch. The consequence would be that Parties that sent proposals will have to re-send them if the epoch is ended by an external commit. It should be noted, that a Preconfigured sender will have no way to detect if their proposal was committed or not.

bifurcation:
Fixes #540

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3342,6 +3342,8 @@ Commits.
 
 External Commits work like regular Commits, with a few differences:
 
+* The Commit MUST NOT include any proposals by reference, since an external
+  joiner cannot determine the validity of proposals sent within the group
 * The proposals included by value in an External Commit MUST meet the following
   conditions:
   * There MUST be a single ExternalInit proposal


bifurcation:
cc @kkohbrok

PR #566, opened 2022-01-28
Split ratchet tree into concepts and operations

bifurcation:
Concepts go into the introduction, operations become their own top-level section.


rohan-wire (changes requested):
## Ratchet Tree Concepts should be 
# Ratchet Tree Concepts

to keep normative and non normative separate

PR #567, opened 2022-01-28
Reorg move crypto forward

bifurcation:
Depends on #566 

Moves the `Cryptographic Objects` and `Message Framing` sections ahead of `Ratchet Tree Operations`, to reduce forward references and have ratchet tree / key schedule / secret tree sections together.

image


bifurcation:
If you look at this with `git diff main --diff-algorithm=minimal`, it shows the session move more clearly.

rohan-wire (approved):

PR #568, opened 2022-01-28
Consolidate ratchet-tree-related sections

bifurcation:
Depends on #567 

Some subsections of `Key Packages` are more natural within the ratchet tree operations section.  This PR pulls them up into the `Ratchet Tree Operations` section.

image

PR #569, opened 2022-01-31
Move Key Packages section down

bifurcation:
Depends on #568 

This swaps the "Key Packages" and "Key Schedule" sections, so that the former is just before "Group Creation".

image


bifurcation:
As usual, best reviewed with `git diff --diff-algorithm=minimal`

Bren2010 (approved):

PR #571, opened 2022-01-31
Move up group state

bifurcation:
Depends on #570 

This PR moves the "Group State" subsection up into the "Key Schedule" section, where it fits more naturally.

image


bifurcation:
`git diff reorg-split-out-secret-tree --diff-algorithm=minimal`

Bren2010 (approved):

PR #573, opened 2022-02-01
Remove endpoint_id

### LINKED ISSUE #558: Do we still need `endpoint_id`?

bifurcation:
Fixes #558

This involved a couple of changes other than just removing the field and references to it:

* Detecting duplicative Adds is now a little more ambiguous, since you can't just match on `endpoint_id`.  However (a) this doesn't degrade the security of the protocol, just makes it a little less efficient (b) as discussed in #558, if applications want to sharpen it, they can add an `endpoint_id`-like identifier in an extension.
* The External Commit language around resync got generalized to say that the KeyPackage needs to meet the requirements of an Update for the removed node, which captures the successor notions we've discussed.

I also noticed that when we forbade proposals by reference in #562, we didn't totally remove them.  So I went ahead and fixed that here.


kkohbrok (approved):
Thanks for taking care of the cleanup here after the discussion on #509.

If the application has to provide a function that determines if a given KeyPackage represents a valid update, we could use the same function to determine if an Add is a duplicate. Although admittedly, in that case one would have to compare with all other KeyPackages in the group and the function might not be very efficient.

Generally, I'm ok with this solution, though 👍🏻.

Bren2010 (approved):

PR #582, opened 2022-02-07
Figure 1 is more accurate showing init_secret as start and end secrets.

rohan-wire:
change Figure 1 to be more accurate:

before: epoch_secret -> epoch_secret -> epoch_secret

after: init_secret -> epoch_secret -> init_secret


bifurcation:
I actually disagree with this.  This figure needs to show one epoch secret chaining to the next.  If you're going to show the init secret, you still need to show the epoch secret.  But that gets too cluttered.  So even though it's true that the init secret is in the middle, the underlying message here is that one epoch secret feeds into the next, and the intervening derivations can be elided.

bifurcation:
Discussed on virtual interim: epoch->epoch better for newcomers.

PR #584, opened 2022-02-08
Move validation of identity change to the AS

kkohbrok:
This PR moves the remark that there has to be validation of the attested identities upon a credential update to the KeyPackage validation subsection. It also changes the responsibility for the check to be that of the AS rather than that of the application. Note, that the boundaries are somewhat arbitrary at this point, since the application running on the local device may well implement parts of the AS. The reason for this change is to ensure any validation checks that pertain to authentication are part of the AS.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2524,9 +2524,6 @@ The client verifies the validity of a KeyPackage using the following steps:
   the set of identities attested by the credential in the new KeyPackage is
   acceptable relative to the identities attested by the old credential.
 
-* Verify that the set of identities attested by the credential is acceptable
-  to the application for the participant being updated.
-
 * Verify that the signature on the KeyPackage is valid using the public key
   in the KeyPackage's credential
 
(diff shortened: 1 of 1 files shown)


bifurcation (approved):

bifurcation:
@kkohbrok Please make any required changes to the architecture document as well.  Since I think that's where the AS is defined.

draft-ietf-mls-protocol.md
```diff
@@ -2517,9 +2517,15 @@ The validity of a KeyPackage needs to be verified at a few stages:
...
+  authentication service and the client's local policy. These actions MUST be
   the same regardless of at what point in the protocol the KeyPackage is being
-  verified.
+  verified with the following exception: If the KeyPackage is an update to
+  another KeyPackage, the authentication service MUST additionally validate that
+  the set of identities attested by the credential in the new KeyPackage is
+  acceptable relative to the identities attested by the old credential.
+
+* Verify that the set of identities attested by the credential is acceptable
+  to the application for the participant being updated.
```
bifurcation, line 2528: DIscussion on the virtual interim:

* This bullet seems redundant given the changes to the first bullet.  It is also out of place here in KeyPackage validation, since there might not be an update.
* TODO(@kkohbrok) Remove this bullet, then clear to merge

kkohbrok: Whoops, I think this wasn't supposed to be there in the first place. I'll remove it.

bifurcation (approved):

Issue #559, opened 2022-01-27
Remove redundant nodes from the tree


bifurcation:
Right now, the UpdatePath generation and processing algorithms populate "redundant" parent nodes in the tree, in the sense that they have only one child node.  These redundant nodes are harmless except that they make tree truncation less effective.  We should update the UpdatePath algorithms so that they do not add redundant nodes to the tree.

raphaelrobert:
I think the definition of redundant is slightly different than what you propose. Redundant nodes are nodes with a resolution of only one leaf node, where all parent nodes are filtered out of the resolution. I think in the very early days we called that a "shadow".

With the definition you propose, I think it would become impossible to do the first-ever full commit in an otherwise empty tree.

bifurcation:
Fixed by #587

PR #587, opened 2022-02-08
Restored aggressive truncation and introduced filtered direct paths t…

MartaMularczyk:
This PR is motivated by an observation made independently by me and @bifurcation that MLS currently generates many "redundant" PKE keypairs which never have to be used. The proposal is to stop generating such key pairs, which greatly improves efficiency. In particular, in some scenarios the solution from this PR results in ~3 times smaller ratchet trees compared to the current MLS (we found such a scenario and @bifurcation illustrated it here [careful-truncate.pdf](https://github.com/mlswg/mls-protocol/files/8026064/careful-truncate.pdf)). This PR addresses the issue #559.

### Problem

To give a better intuition, we explain the problem on an example. Take the following ratchet tree.

>                         T
>                 ________|_______
>                /                \
>               U                 |
>         ______|______           |
>        /             \          |
>       V               W         |
>     __|__           __|__       |
>    /     \         /     \      |
>   X       Y       Z       -     |
>  / \     / \     / \     / \    |
> A   B   C   D   E   -   -   -   F

Currently, when E rekeys her path in a commit, she generates new PKE key pairs for all nodes on her direct path, i.e., Z, W, U and T. We notice that the key pairs for nodes Z and W are redundant -- since the subtrees of their children on the copath are all blank, E's leaf key pair can be used equivalently. Moreover, we notice that generating redundant key pairs is inefficient for 2 reasons:
1. The redundant public keys have to be sent as part of E's commit and potentially welcome messages.
2. The redundant public keys together with the parent hash mechanism prevent "aggressive truncation". Aggressive truncation was introduced in #461 in order to "heal" the ratchet tree (which decreases complexity). However, it broke the protocol due to incompatibility with the parent hash mechanism. For this reason, it was replaced in #524 by a far less efficient "careful truncation" mechanism. 

### Solution

This PR proposes to stop generating redundant key pairs, getting rid of both of the above inefficiencies. More precisely, we say that a node on E's path is _redundant_ if its copath child (or in general any child) has an empty resolution (which implies that its subtree is empty). Further, we say that E's _filtered direct path_ is her direct path with all redundant nodes removed. When E commits, she rekeys only her filtered direct path. The redundant nodes are blanked. In the example above, E's filtered direct path consists of Y and Z. After her, commit we get the following tree.

>                         T
>                 ________|_______
>                /                \
>               U                 |
>         ______|______           |
>        /             \          |
>       V               -         |
>     __|__           __|__       |
>    /     \         /     \      |
>   X       Y       -       -     |
>  / \     / \     / \     / \    |
> A   B   C   D   E   -   -   -   F

This way, E doesn't send redundant public keys for Z and W. Moreover, we can restore aggressive truncation -- the blank nodes above E can be truncated without destroying the parent hash. The following picture by @bifurcation shows how the above solution can result in ~3 times smaller trees than currently generated by MLS [careful-truncate.pdf](https://github.com/mlswg/mls-protocol/files/8026064/careful-truncate.pdf).

### Parent Hash

The above solution requires modifying how the parent hash is verified by new members joining the group. Roughly, before the proposed change, the new member for each node U with children V and W performed the following check: either A) `V.parent_hash` is equal to the parent hash of U with copath child W or B) `W.parent_hash` is equal to the parent hash of U with copath child V. In the running example, B) would be true.

After the proposed change, W is blank and the parent hash of U with copath child V is stored in the node E instead. Accordingly, the new member performs the following modified check: either A) there exists a node X in the resolution of V s.t. `X.parent_hash` is equal to the parent hash of U with copath child W or B) [modified analogous to A)].

From the security perspective, the new check fulfils the same purpose as we meant for the original check: for every node U, the member who last rekeyed U attests to the publickey U.pk they generated and the public keys they told U.sk to.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1518,15 +1518,24 @@ to other members in the group via an UpdatePath message (see {{update-paths}}) .
 All other group members then apply the key material in the UpdatePath to their
 own local tree state to derive the group's now-updated shared secret.
 
-To begin, the generator of the UpdatePath updates its leaf
-KeyPackage and its direct path to the root with new secret values.  The
-HPKE leaf public key within the KeyPackage MUST be derived from a freshly
-generated HPKE secret key to provide post-compromise security.
+To begin with, the generator of the UpdatePath updates its leaf's KeyPackage and
+its leaf's _filtered direct path_ with new key pairs. The filtered direct path 
+of a leaf is obtained from the leaf's direct path by removing all nodes whose
+child on the leaf's copath has an empty resolution. Such a removed node does not
+need a key pair, since after blanking it, its resolution consists of a single
+node on the filtered direct path. Using the key pair of the node in the
+resolution is equivalent to using the key pair of the removed node.
 
-The generator of the UpdatePath starts by sampling a fresh random value called
-"leaf_secret", and uses the leaf_secret to generate their leaf HPKE key pair
-(see {{key-packages}}) and to seed a sequence of "path secrets", one for each
-ancestor of its leaf. In this setting,
+The generator of the UpdatePath starts by updating the KeyPackage of its leaf.
+The HPKE leaf public key within the KeyPackage MUST be
+derived from a freshly generated HPKE secret key to provide post-compromise
+security.
+
+Further, the generator of the UpdatePath blanks all nodes on its direct path
+that are not on its filtered direct path. Then, it samples a fresh random value
+called "leaf_secret", and uses the leaf_secret to generate their leaf HPKE key
+pair (see {{key-packages}}) and to seed a sequence of "path secrets", one for
+each node on its filtered direct path. In this setting,
 path_secret\[0\] refers to the leaf's parent,
 path_secret\[1\] to the parent's parent, and so on. At each step, the path
 secret is used to derive a new secret value for the corresponding
@@ -1661,13 +1670,13 @@ apply it to keep their local views of the tree in
 sync with the sender's.  More specifically, when a member commits a change to
 the tree (e.g., to add or remove a member), it transmits an UpdatePath
 containing a set of public keys and encrypted path secrets
-for intermediate nodes in the direct path of its leaf. The
+for intermediate nodes in the filtered direct path of its leaf. The
 other members of the group use these values to update
 their view of the tree, aligning their copy of the tree to the
 sender's.
 
 An UpdatePath contains
-the following information for each node in the direct path of the
+the following information for each node in the filtered direct path of the
 sender's leaf, including the root:
 
 * The public key for the node
@@ -1683,7 +1692,7 @@ of the non-updated child.
 The recipient of an UpdatePath processes it with the following steps:
 
 1. Compute the updated path secrets.
-   * Identify a node in the direct path for which the local member
+   * Identify a node in the filtered direct path for which the local member
      is in the subtree of the non-updated child.
    * Identify a node in the resolution of the copath node for
      which this node has a private key.
@@ -1694,13 +1703,14 @@ The recipient of an UpdatePath processes it with the following steps:
    * The recipient SHOULD verify that the received public keys agree
      with the public keys derived from the new path_secret values.
 2. Merge the updated path secrets into the tree.
-   * For all updated nodes,
-     * Replace the public key for each node with the received public key.
+   * Blank all nodes on the direct path of the sender's leaf.
+   * For all nodes on the filtered direct path of the sender's leaf,
+     * Set the public key to the received public key.
      * Set the list of unmerged leaves to the empty list.
-     * Store the updated hash of the node's parent (represented as a ParentNode
-       struct), going from root to leaf, so that each hash incorporates all the
-       nodes above it. The root node always has a zero-length hash for this
-       value.
+     * Store the updated hash of the next node on the filtered direct path
+       (represented as a ParentNode struct), going from root to leaf, so that
+       each hash incorporates all the non-blank nodes above it. The root node
+       always has a zero-length hash for this value.
    * For nodes where a path secret was recovered in step 1,
      compute and store the node's updated private key.
 
@@ -1779,22 +1789,18 @@ left and right children, respectively.
 
 ## Parent Hash {#parent-hash}
 
-The `parent_hash` extension carries information to authenticate the structure of
-the tree, as described below.
+The `parent_hash` extension carries information to authenticate the HPKE keys
+in the ratchet tree, as described below.
 
 ~~~~~
 opaque parent_hash<0..255>;
 ~~~~~
 
-Consider a ratchet tree with a parent node P and children V and S. The parent hash
-of P changes whenever an `UpdatePath` object is applied to the ratchet tree along
-a path traversing node V (and hence also P). The new "Parent Hash of P (with Co-Path
-Child S)" is obtained by hashing P's `ParentHashInput` struct using the resolution
-of S to populate the `original_child_resolution` field. This way, P's Parent Hash
-fixes the new HPKE public keys of all nodes on the path from P to the root.
-Furthermore, for each such key PK the hash also binds the set of HPKE public keys
-to which PK's secret key was encrypted in the Commit that contained the
-`UpdatePath` object.
+Consider a ratchet tree with a non-blank parent node P and children V and S.
+The parent hash of P changes whenever an `UpdatePath` object is applied to
+the ratchet tree along a path from a leaf U traversing node V (and hence also
+P). The new "Parent Hash of P (with Co-Path Child S)" is obtained by hashing P's
+`ParentHashInput` struct.
 
 ~~~~~
 struct {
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1518,15 +1518,24 @@ to other members in the group via an UpdatePath message (see {{update-paths}}) .
...
-ancestor of its leaf. In this setting,
+To begin with, the generator of the UpdatePath updates its leaf's KeyPackage and
+its leaf's _filtered direct path_ with new key pairs. The filtered direct path 
+of a leaf is obtained from the leaf's direct path by removing all nodes whose
+child on the leaf's copath has an empty resolution. Such a removed node does not
+need a key pair, since after blanking it, its resolution consists of a single
+node on the filtered direct path. Using the key pair of the node in the
+resolution is equivalent to using the key pair of the removed node.
+
+The generator of the UpdatePath starts by updating the KeyPackage of its leaf.
```
kkohbrok, line 1529: This is already mentioned in the paragraph above.

draft-ietf-mls-protocol.md
```diff
@@ -1518,15 +1518,24 @@ to other members in the group via an UpdatePath message (see {{update-paths}}) .
...
+The HPKE leaf public key within the KeyPackage MUST be
+derived from a freshly generated HPKE secret key to provide post-compromise
```
kkohbrok, line 1531: I'm not sure if that is what you meant here, but I believe the HPKE key pair is generated from a given (symmetric) secret rather than the public key being generated using the private key.

draft-ietf-mls-protocol.md
```diff
@@ -1804,29 +1810,33 @@ struct {
...
 
-Finally, `original_child_resolution` is the array of `HPKEPublicKey` values of the
+Finally, `original_child_resolution` is the array of HPKE public keys of the
 nodes in the resolution of S but with the `unmerged_leaves` of P omitted. For
 example, in the ratchet tree depicted in {{resolution-example}} the
 `ParentHashInput` of node Z with co-path child C would contain an empty
 `original_child_resolution` since C's resolution includes only itself but C is also
 an unmerged leaf of Z. Meanwhile, the `ParentHashInput` of node Z with co-path child
 D has an array with one element in it: the HPKE public key of D.
+This way, P's Parent Hash fixes the set of HPKE public keys to which the new HPKE
```
kkohbrok, line 1826: A pretty similar sentence is already in the paragraph above.

draft-ietf-mls-protocol.md
```diff
@@ -1518,15 +1518,24 @@ to other members in the group via an UpdatePath message (see {{update-paths}}) .
...
-To begin, the generator of the UpdatePath updates its leaf
-KeyPackage and its direct path to the root with new secret values.  The
-HPKE leaf public key within the KeyPackage MUST be derived from a freshly
-generated HPKE secret key to provide post-compromise security.
-
-The generator of the UpdatePath starts by sampling a fresh random value called
-"leaf_secret", and uses the leaf_secret to generate their leaf HPKE key pair
-(see {{key-packages}}) and to seed a sequence of "path secrets", one for each
-ancestor of its leaf. In this setting,
+To begin with, the generator of the UpdatePath updates its leaf's KeyPackage and
```
kkohbrok, line 1521: I'm aware that this is not the style this section was written in previously, but when trying to implement an algorithm like this, I found it very helpful in the past if the description is a bit more structured. An example of this is the parent hash verification algorithm that (in the shape of bullet points) has pretty clear instructions on what should happen when.

kkohbrok (commented):
Neat solution! Just a few editorial comments.

Looking forward to implementing and testing it.

TWal:
I feel a bit uncomfortable with the part "there exists a node X in the resolution of V such that …". I think we can replace "resolution" with "original resolution" and furthermore enforce the original resolution to only contain one node?

draft-ietf-mls-protocol.md
MartaMularczyk: It's similar, but there's a key difference: the paragraph above considers attesting to the public key V.pk for each node V on the path. This paragraph considers attesting to the public keys U.pk to which V.sk was encrypted.

Maybe giving names to nodes would make it clearer? Like "This way, P's parent hash fixes V.pk for each V on the filtered path from P to the root" and then "This way, P's parent hash fixes all public key to which V.pk was encrypted for each V[...]"?

kkohbrok (approved):
Thanks for addressing my comments! Looks good to me now.

draft-ietf-mls-protocol.md
MartaMularczyk: I agree. But I also think that this text could use a general improvement. I created a separate PR for this #593

MartaMularczyk:
> I feel a bit uncomfortable with the part "there exists a node X in the resolution of V such that …". I think we can replace "resolution" with "original resolution" and furthermore enforce the original resolution to only contain one node?

I think that the "uneasy" feeling may come from the fact that we don't agree about the intuition for what the parent hash attempts to achieve. Maybe it would help to explain it better in the draft, perhaps in "security considerations". Details follow.

I noticed that many people think that the parent hash attempts to authenticate the structure of the tree. I have a different opinion. The parent hash attempts to enforce is the tree invariant. More precisely, say that a signature or a PKE key pair is "good" if the secret key is not leaked to or chosen by the adversary. The tree invariant requires that for each node W, if the signature keys of all members in W's subtree are good then W's PKE key pair is good. Looking ahead, this allows to give the following guarantee to joiners: if the signature keys of all group members are good, then all PKE key pairs in the tree are good. Thus, the next commit heals the group key.

Thinking about authenticating the structure, one naturally thinks that via the parent hash, the committer authenticates their path together with its properties, e..g, the length. This PR suggests to not authenticate some of such properties which naturally makes people uneasy.

But I think that authenticating such properties is not the job of the parent hash. Having in mind enforcing the tree invariant, I thinks that via the parent hash, the committer authenticates, the PKE public key of each (individual) node W on its path.

The most direct way to achieve the above would be to include in each node W on the path the committer's signature over W's PKE public key and W's original child resolution. However, this is inefficient. So instead, we use the current "hash to the leaf & sign" method. It's important to understand that it's only an efficiency improvement.

We can arrive at the "hash to the leaf & sign" solution by the following reasoning. If there was only one committer A, she could simply sign her whole path with a single signature. This doesn't work with many committers, because the last part of the path may be rekeyed by a different committer B. In this case, A's signature can no longer be verified (unless we store old public keys, which would be secure but inefficient). Therefore, we modify the solution so that one can verify any prefix of the sequence signed by A. The "hash to the leaf & sign" method allows to do just this.

TWal:
I agree that security-wise, as long as the public key and sibling resolution of every node is ultimately signed by a leaf, we are fine. And this property still holds after this PR.

Here is how I understand parent hash: when generating a Commit, the leaf must sign the new public key for each node that was modified, and also sign who knows the private keys associated with the new public keys (== who can decrypt the data when we encrypt to the new public key).
(Big parenthesis)
The resolution approach works by "recursion" in some sense, the signature says something like "the people knowing the private part of this key are the people knowing the private part of these other public keys".
The "tree authentication" approach is more straightforward, the signature encapsulate every leaf who could have decrypted the private part of the public key.
(End of parenthesis)
Finally as you explained, the parent hash construct is a way to do all these signature using only one signature, and furthermore make it resistant against future commits.

If we look at every node whose content was generated by a given leaf, we can notice it has a nice structure: it corresponds to a path starting to the leaf and going up to the root, possibly reaching the root if the leaf was the last committer, or stopping earlier in most cases.

With that said, here is the first thing that makes me uncomfortable: the parent hash in the current document exactly covers path with the shape described above: every cover of the tree with such paths can be correctly encoded using parent hash, and valid parent hashes on a tree encode such a cover of paths on the tree. With your proposal, we lose the second property.

The second thing that makes me uncomfortable is that in the parent hash correctness proof I've worked on, I use the notion of "expected parent hash for a node" (well I should talk about "left expected parent hash" and "right expected parent hash"), but the thing is we can give this a name because it is unique. Saying that we must search inside the resolution makes it not unique anymore.


During the last virtual interim I think you said that my proposal is only an optimization (hence optional), and I agree.
However thinking more about it, I have some questions:
- the goal of this PR is to optimize MLS, so why not do all the optimizations available?
- saying "a node U in the subtree of V" is conceptually simpler than "a node U in the resolution of V" while achieving the same security guarantee, so using the resolution is already an optimization?

bifurcation:
@TWal - Note that this proposal is actually synergistic with using tree hashes to represent siblings in parent hashes.  TH-for-PH is faster with sparser trees, and this proposal makes trees as sparse as possible. 

As far as "unease", I think there's a solution that removes the ambiguity of "searching" for a child in the resolution and is compatible with the changes in PR.  Since we're trying to verify that every node chains to some leaf via the parent hash, simply update the parent hash validation to work up from the leaves instead of processing parent independently:

* Let validated = set()
* For each leaf node:
  * Let N = leaf node
  * while N != root:
    * P = N.parent
    * while !validated.contains(P) and P is blank and P != root: validated.add(P); P = P.parent;
    * if validated.contains(P); break;
    * if P is blank; validated.add(P); break;
    * if !parent_hash_match(P, N): break;
    * validated.add(P)
    * N = P
* return validated.len() == nodes.len()

The only ambiguity here is in the number of blank nodes you traverse before hitting the next non-blank parent.  (Without this PR, you would break if P is blank.)  But here again, the TH-for-PH helps by authenticating the subtrees: The most recent UpdatePath signs the whole structure of the tree, including all the blanks.  And the blanks in that UpdatePath are guaranteed by the filtered direct path computation and the height of the copath subtrees.

So it seems like if we do this PR, the TH-for-PH PR, and revise the parent hash validation as above, we end up with better performance and a cleaner semantic for the parent hash.

TWal:
I'm noticing that I forgot to say that I was globally in favor of this PR!

The only itch for me really is the resolution part. The easiest fix for this would be to check that the original child resolution only contain one node and expected parent hash = this node's content.parent_hash.

Why is it true that the original child resolution only contain one node?
- It's true when processing and UpdatePath (the child is non-blank since it's part of an UpdatePath so its resolution only contain itself)
- It remains true when extending the tree or truncating the tree (since the resolution is not affected by adding or removing blank nodes)
- It remains true when adding a leaf in an available blank leaf (thanks to the "original" in "original child resolution").

bifurcation:
I resolved merge conflicts in the most obvious way possible.  Along the way I noticed that we reference the `parent_hash` extension, which is no longer relevant.  I will file a follow-up to clean this up.

draft-ietf-mls-protocol.md
```diff
@@ -2095,30 +2102,42 @@ struct {
...
 
-The Parent Hash of P with Co-Path Child S is the hash of a `ParentHashInput` object
-populated as follows. The field `public_key` contains the HPKE public key of P. If P
-is the root, then `parent_hash` is set to a zero-length octet string.
-Otherwise `parent_hash` is the Parent Hash of P's parent with P's sibling as the
-co-path child.
+The field `public_key` contains the HPKE public key of P. If P is the root,
+then the `parent_hash` field is set to a zero-length octet string. Otherwise,
+`parent_hash` is the Parent Hash of the next node after P on the filtered
+direct path of U (see {{resolution-example}}). This way, P's Parent Hash fixes
```
bifurcation, line 2108:
```suggestion
direct path of U. This way, P's Parent Hash fixes
```

Broken link.

PR #592, opened 2022-02-10
Add a simple ProofCredential alternative to Certificate.

Bren2010:
- Adds an `extra_data` field to BasicCredential to hold application-specific data about a user (examples: link to avatar, information about device)
- Adds a ProofCredential to cover a lot of the cases we're currently missing, particularly:
  - credentials that are just BasicCredential signed with a fixed public key
  - credentials that are authenticated by a system like Key Transparency
  - credentials authenticated by secure hardware

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1218,6 +1218,7 @@ uint16 CredentialType;
 
 struct {
     opaque identity<V>;
+    opaque extra_data<V>;
     SignatureScheme signature_scheme;
     opaque signature_key<V>;
 } BasicCredential;
@@ -1227,6 +1228,12 @@ struct {
 } Certificate;
 
 struct {
+  opaque authority_id<0..255>;
+  BasicCredential credential;
+  opaque proof<0..2^16-1>;
+} ProofCredential;
+
+struct {
     CredentialType credential_type;
     select (Credential.credential_type) {
         case basic:
@@ -1234,6 +1241,9 @@ struct {
 
         case x509:
             Certificate chain<V>;
+
+        case proof:
+            ProofCredential;
     };
 } Credential;
 ~~~~~
@@ -1245,7 +1255,9 @@ ratchet tree or the KeyPackage ciphersuite for a credential in a KeyPackage
 object.  For ciphersuites using Ed25519 or Ed448 signature schemes, the public
 key is in the format specified {{?RFC8032}}.  For ciphersuites using ECDSA with
 the NIST curves P-256 or P-521, the public key is the output of the uncompressed
-Elliptic-Curve-Point-to-Octet-String conversion according to {{SECG}}.
+Elliptic-Curve-Point-to-Octet-String conversion according to {{SECG}}. The
+`extra_data` field may contain any additional application-specific data about
+the identity.
 
 For an X.509 credential, each entry in the chain represents a single DER-encoded
 X.509 certificate. The chain is ordered such that the first entry (chain[0])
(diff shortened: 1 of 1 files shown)


rohan-wire:
I am concerned about extra_data having neither semantics nor name spacing. If extra_data is scoped to an authority or a namespace, then i don't see the harm, but as it is now, I think it is harmful to interoperability.

Bren2010:
`extra_data` is a place for general application data, similar to `authenticated_data`, and the protocol would be agnostic to what it contains. I think if you did want to specify what goes into it for interop, that might go in the "Federation" draft.

bifurcation:
I'm not sure what the extra mechanism here is buying on top of (a) the `BasicCredential.identity` field and (b) extensions.  A lot of the use cases described would work well in extensions.  For example, just like a CT log inclusion proof can be sent in a TLS ServerHello extension, it seems like you could put a KT inclusion proof in a LeafNode extension.  And if you want to stick within the basic framework, the `identity` field is entirely application defined, so you can just use it like you would `extra_data`.

kkohbrok:
I'm not opposed to this change specifically, but I'm generally wondering if we need concrete credentials in the spec beyond the basic credential, which is useful for interop. After we've pushed so much of the authentication logic into the AS and the implementation, the spec doesn't really care about what the credential looks like. I don't think we even strictly need it to provide a signature key, since message authentication just needs "the signing private key corresponding to the public key, which was authenticated by the credential". I'm still not sure why X509 needs to be in there.

Bren2010:
> A lot of the use cases described would work well in extensions. For example, just like a CT log inclusion proof can be sent in a TLS ServerHello extension, it seems like you could put a KT inclusion proof in a LeafNode extension.

I would like us to have some way to support these cases already in the protocol so you don't have to immediately go out and build something new if you want to use anything besides X.509. So I can switch the ProofCredential to some kind of Proof extension but that seems like a worse abstraction? And it's not clear why Proof would be an extension when X.509 is a type of credential.

> And if you want to stick within the basic framework, the identity field is entirely application defined, so you can just use it like you would extra_data.

It's more about the ease of having one field which contains just the unique identifier for the client and another field which may need to be parsed to be understood

bifurcation:
Discussion on the virtual interim:

* On the one hand, could implement this in a follow-on spec
* On the other hand, could be nice to avoid the need for implementations to add stuff off the bat
* TODO(@Bren2010): Flesh out use cases

rohan-wire:
Thanks @Bren2010 for writing up these use cases. My concern is that if the authority _id is under-specified we could end up with a mess where different implementations assume a different syntax/semantic/context. This definitely happened with the Subject in X.509 certificates and it took almost a decade before we had reasonable semantics defined for SubjectAltName (URI, DNS, RID, IP, dirName, or otherName) and implementations still get confused sometimes when expecting DNS but a URI is present or vice versa.

Bren2010:
There's no structure prescribed to `authority_id`, much like `group_id` it's an opaque identifier. An implementation that relies on it having a certain structure wouldn't be compliant with the specification.

rohan-wire:
> There's no structure prescribed to `authority_id`, much like `group_id` it's an opaque identifier. An implementation that relies on it having a certain structure wouldn't be compliant with the specification.

With an additional field with no structure, the ability to interoperate later will be even more difficult. I am already uncomfortable that there isn't even any optional structure in the Identity in a BasicCredential. As I said, look at what happened with Subject name in X.509.

Bren2010:
@rohan-wire I'm not sure why this concern doesn't apply to `group_id` but I could add some text to the PR along the lines of what I said in my last comment?

```
Implementations MAY construct the value of the `authority_id` field to have a particular 
format, but MUST NOT enforce that the `authority_id` field always has that format. 
```

rohan-wire:
What if every credential contains a BasicCredential and you can just build on top of it?
ex:  https://github.com/mlswg/mls-protocol/pull/628

Bren2010:
Hey @rohan-wire, that idea doesn't really address the same problem as this PR. With this PR merged, any baseline implementation of MLS would support a restricted set of credential types: those where you can condense the necessary proof information into the "proof" field, and the implementation can call an application-provided callback to verify the proof. The application only needs to provide a callback that implements the verification logic because the wire format and data plumbing is already set up.

Without this PR, and also with #628, any new credential type has to be implemented as an extension to the protocol which would reasonably require modifying the MLS implementation. That's a lot of work and creates a maintenance burden.

claucece:
I agree with @Bren2010 in that (as it was pointed out) it avoids the need for `implementations to add stuff off the bat`. I do see, though, the importance of being careful so we don't end up with the problems of the subject name in X.509. This seems reasonable for the latter point: https://github.com/mlswg/mls-protocol/pull/592#issuecomment-1086219398

seanturner:
I posted this PR to the [mls list](https://mailarchive.ietf.org/arch/msg/mls/SitToPUL1CvPgx0vLnom730Wwes/) to see if we could get more input. I got crickets in response. I am going to close this PR out because it does not appear that we have consensus to merge this at this time.

PR #593, opened 2022-02-11
Clean up the description of tree evolution.

MartaMularczyk:
Changes (apart from editorial):
* Remove the text about the key schedule. Reason: I found the text confusing, because updating paths provides FS and PCS for the HPKE keys in the ratchet tree, and not directly for the key schedule.
* Generate the leaf key pair simply as a fresh HPKE key pair, and not from the leaf_secret. I couldn't see any benefit of using the leaf_secret. I wonder if there is a good reason?

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1512,43 +1512,36 @@ describe the content of the tree and the required operations.
 
 ## Ratchet Tree Evolution
 
-A member of an MLS group advances the key schedule to provide forward secrecy
-and post-compromise security by providing the group with fresh key material to
-be added into the group's shared secret.
-To do so, one member of the group generates fresh key
-material, applies it to their local tree state, and then sends this key material
-to other members in the group via an UpdatePath message (see {{update-paths}}) .
-All other group members then apply the key material in the UpdatePath to their
-own local tree state to derive the group's now-updated shared secret.
+Updating paths heals the ratchet tree, which in turn allows to heal the group key
+In order to provide forward secrecy and post-compromise security, whenever a
+member initiates an epoch change (i.e., commits; see {{commit}}), they refresh
+all HPKE secret keys in the ratchet tree that they know, that is, the secret
+keys of all nodes on their direct path.
 
-To begin, the generator of the UpdatePath updates its leaf
-KeyPackage and its direct path to the root with new secret values.  The
-HPKE leaf public key within the KeyPackage MUST be derived from a freshly
-generated HPKE secret key to provide post-compromise security.
+The member generates the fresh key pairs using the following procedure. The
+procedure is designed in a way that allows to efficiently communicate the fresh
+secret keys to other group members, as described in {{update-paths}}. 
 
-The generator of the UpdatePath starts by sampling a fresh random value called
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1504,51 +1504,42 @@ member of the group.
...
+all HPKE key pairs in the ratchet tree for which they know the secret keys.
+More precisely, they refresh the key pairs of their leaf and of all nodes on
+their leaf's direct path.
 
-To begin, the generator of the UpdatePath updates its leaf
-KeyPackage and its direct path to the root with new secret values.  The
-HPKE leaf public key within the KeyPackage MUST be derived from a freshly
-generated HPKE secret key to provide post-compromise security.
+The member initiating the epoch change generates the fresh key pairs using the
+following procedure. The procedure is designed in a way that allows to
```
bifurcation, line 1522: "allows" needs a subject here.

```suggestion
following procedure. The procedure is designed in a way that allows group members to
```

draft-ietf-mls-protocol.md
```diff
@@ -1504,51 +1504,42 @@ member of the group.
...
-"leaf_secret", and uses the leaf_secret to generate their leaf HPKE key pair
-(see {{key-packages}}) and to seed a sequence of "path secrets", one for each
-ancestor of its leaf. In this setting,
-path_secret\[0\] refers to the leaf's parent,
-path_secret\[1\] to the parent's parent, and so on. At each step, the path
-secret is used to derive a new secret value for the corresponding
-node, from which the node's key pair is derived.
+* Generate a fresh HPKE key pair for the leaf.
+* Generate a sequence of path secrets, one for each node on the leaf's direct
+path as follows. Sample a fresh random value `path_secret[0]` and compute
```
bifurcation, line 1528: I would split the middle bullet here:

* Generate the leaf key pair
* Generate path secrets
    * `path_secret[0]` is sampled at random
    * `path_secret[n] = DeriveSecret(path_secret[n-1], "path")`
* Derive node key pairs from path secrets

bifurcation (approved):
A couple of editorial comments, but in general, LGTM

draft-ietf-mls-protocol.md
bifurcation: Note also that the current version doesn't render quite properly.  "In this context..." isn't rendered as part of the bullet.  You might be able to fix this by indenting that part by 2 spaces.  Or you might have better success putting the code block after the prose.

rohan-wire:
> * Generate the leaf key pair simply as a fresh HPKE key pair, and not from the leaf_secret. I couldn't see any benefit of using the leaf_secret. I wonder if there is a good reason?

I am concerned that we are asking for a LOT of fresh entropy if (in addition to nonces) we need a crypto-random value every epoch for every group (a Wire client can easily a member of thousands of groups at a time). Do you have a concern about deriving it?

bifurcation:
Discussion on the virtual interim:

* To @rohan-wire's concern: This allows the base path secret to be derived, e.g., using a KDF, so this doesn't *necessarily* increase the amount of entropy needed
* Clear to merge

bifurcation (approved):

PR #594, opened 2022-02-13
TLS syntax updates

bifurcation:
There are two changes in this PR:

* Pull the `optional<T>` extension to the TLS syntax up into the Terminology section
* Add an extension that lets vectors have variable-length headers, allowing them to be efficient at small sizes while also accommodating large sizes

All vectors are converted to the variable-length header representation.  It seems like this should simplify implementation, since the encoder will no longer have to be told how many bytes of header to use for a given vector.  This has been the source of several annoying interop bugs.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -419,8 +419,69 @@ Application Message:
 Terminology specific to tree computations is described in
 {{ratchet-tree-terminology}}.
 
-We use the TLS presentation language {{!RFC8446}} to
-describe the structure of protocol messages.
+## Presentation Langauge
+
+We use the TLS presentation language {{!RFC8446}} to describe the structure of
+protocol messages.  In addition to the base syntax, we add two additional
+features, the ability for fields to be optional and the ability for vectors to
+have variable-size length headers.
+
+### Optional Value
+
+An optional value is encoded with a presence-signaling octet, followed by the
+value itself if present.  When decoding, a presence octet with a value other
+than 0 or 1 MUST be rejected as malformed.
+
+~~~~~
+struct {
+    uint8 present;
+    select (present) {
+        case 0: struct{};
+        case 1: T value;
+    }
+} optional<T>;
+~~~~~
+
+### Variable-size Vector Headers
+
+In the TLS presentation language, vectors are encoded as a sequence of encoded
+elements prefixed with a length.  The length field has a fixed size set by
+specifying the minimum and maximum lengths of the encoded sequence of elements.
+
+In MLS, there are several vectors whose sizes vary over significant ranges.  So
+instead of using a fixed-size length field, we use a variable-size length using
+the variable-length integer encoding described in Section 16 of  {{?RFC9000}}.
+Instead of presenting min and max values, the vector description simply includes
+a `V`. For example:
+
+~~~~~
+struct {
+    uint32 fixed<0..255>;
+    opaque variable<V>;
+} StructWithVectors;
+~~~~~
+
+Such a vector can represent values with length from 0 bytes to 2^62 bytes.
+The variable-length integer encoding reserves the two most significant bits
+of the first byte to encode the base 2 logarithm of the integer encoding length
+in bytes.  The integer value is encoded on the remaining bits, in network byte
+order.
+
+This means that integers are encoded on 1, 2, 4, or 8 bytes and can encode 6-,
+14-, 30-, or 62-bit values respectively.
+
+| 2Bit | Length | Usable Bits | Range                 |
+|:-----|:-------|:------------|:----------------------|
+| 00   | 1      | 6           | 0-63                  |
+| 01   | 2      | 14          | 0-16383               |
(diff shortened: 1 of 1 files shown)


TWal:
Looks great!

I only have a small objection: there is now several valid ways to serialize a length, as the example shows, both `25` and `40 25` are valid ways to encode 37. This means that if you parse a structure, and later want to re-serialize it to check a signature for example, you need to remember how many bits were used to store the length.
In short, we loose the property that `serialize(parse(buffer)) == buffer`.

The fix is easy: can we add that "the length MUST be stored using the smallest amount of bits possible"?

bifurcation:
Good point, updated!

rohan-wire:
I wasn't clear under which circumstances you would use a longer length with a number that would fit in a smaller range.  
`| 00   | 1      | 6           | 0-63                  |`
`| 01   | 2      | 14          | 0-16383               |`
`| 10   | 4      | 30          | 0-1073741823          |`
`| 11   | 8      | 62          | 0-4611686018427387903 |`

If never, then the table should be:
`| prefix | length bits | usable bits | min | max |`  
`| 00   | 1  | 6  | 0 | 63 |`
`| 01   | 2  | 14 | 64 | 16383 |`
`| 10   | 4  | 30 | 16384 | 1073741823 |`
`| 11   | 8  | 62 | 1073741824 | 4611686018427387903 |`

draft-ietf-mls-protocol.md
```diff
@@ -419,8 +419,71 @@ Application Message:
...
+the variable-length integer encoding described in Section 16 of  {{?RFC9000}}.
+Instead of presenting min and max values, the vector description simply includes
+a `V`. For example:
+
+~~~~~
+struct {
+    uint32 fixed<0..255>;
+    opaque variable<V>;
+} StructWithVectors;
+~~~~~
```
Bren2010, line 462: Was the intention to convert all vectors to variable length?
- If so, there are still some fixed-length left
- If not, how did you decide which to change?

draft-ietf-mls-protocol.md
```diff
@@ -1240,10 +1303,10 @@ enum {
 } SenderType;
 
 struct {
-    SenderType sender_type;
+    SenderType sendertype;
```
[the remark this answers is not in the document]
bifurcation: I think this is a typo anyway.  Will reverse.

draft-ietf-mls-protocol.md
bifurcation: The example here is intended to have an old-style vector next to the new style, of course.

I did intend to change everything over, but you're correct that I missed a bunch with whatever regexp I used.  I'll take another pass.

bifurcation: Looks like this was due to merge conflicts.  Fixed now in any case.

draft-ietf-mls-protocol.md
```diff
@@ -419,8 +419,71 @@ Application Message:
...
+
+### Variable-size Vector Headers
+
+In the TLS presentation language, vectors are encoded as a sequence of encoded
+elements prefixed with a length.  The length field has a fixed size set by
+specifying the minimum and maximum lengths of the encoded sequence of elements.
+
+In MLS, there are several vectors whose sizes vary over significant ranges.  So
+instead of using a fixed-size length field, we use a variable-size length using
+the variable-length integer encoding described in Section 16 of  {{?RFC9000}}.
```
franziskuskiefer, line 453: RFC9000 also has pseudocode for doing the encoding. Maybe it's worth mentioning it here.

draft-ietf-mls-protocol.md
```diff
@@ -419,8 +419,71 @@ Application Message:
...
+    uint32 fixed<0..255>;
+    opaque variable<V>;
+} StructWithVectors;
+~~~~~
+
+Such a vector can represent values with length from 0 bytes to 2^62 bytes.
+The variable-length integer encoding reserves the two most significant bits
+of the first byte to encode the base 2 logarithm of the integer encoding length
+in bytes.  The integer value is encoded on the remaining bits, in network byte
+order.  The encoded value MUST use the smallest number of bits required to
```
franziskuskiefer, line 468: The MUST here differs from the definition in RFC9000 that explicitly says "Values do not need to be encoded on the minimum number of bytes necessary [...]". It should be made clear that there's difference between the algorithm here and the one in RFC9000.

bifurcation:
DIscussion on virtual interim

* Some concern about being able to send gigantic stuff
    * In adversarial cases, the 4-byte headers used for a lot of stuff might already be a problem
    * Could also have length constraints decoupled from the syntax
    * SIP experience shows that imposing arbitrary length limits isn't a good idea
    * ... but applications should have a way to signal that things are too big
    * TODO(@bifurcation): Advisory text "Beware, things can get big; consider imposing limits and clear failure cases for this"
    * Consider streaming implementations, where you might not know that the whole message fits in memory
* Modulo the above TODO, clear to merge

PR #595, opened 2022-02-15
Allowlist approach for proposals in external commit

kkohbrok:
The current description of external commits still leaves room for the committer to include various other proposals such as arbitrary proposals that are not Update or Remove proposals (e.g., Adds, RequiredCapability, PSK, etc.). I think we should narrow that down to a specific set of proposals (i.e. an allow list style) to keep the implementation and general protocol complexity low for external commits, both for creation and processing.

The way I see it, the new joiner can always issue a follow-up commit for other proposals, but I might be missing important use-cases, so I'd be grateful for feedback on this.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3628,7 +3628,8 @@ The information in a GroupInfo is not deemed public in general, but applications
 can choose to make it available to new members in order to allow External
 Commits.
 
-In principle, External Commits work like regular Commits. However, their content has to meet a specific set of requirements:
+In principle, External Commits work like regular Commits. However, their content
+has to meet a specific set of requirements:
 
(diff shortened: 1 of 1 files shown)


bifurcation:
I agree with the intent here.  Looking through the list of proposal types, the only other one that seems like it might be useful is PSK -- I could see it being useful to prove that you have a PSK when joining from the outside.  If we can add that as an option (probably in parallel to Remove), this seems good to go to me.

rohan-wire:
Would an External Commit adding yourself contain an Add or an Update Proposal?

kkohbrok:
It used to contain an `Add` proposal, but since that's pretty much redundant with the `path` the commiter has to include, it got optimized away. So an external commit should be handled as if it contained an Add proposal, where the KeyPackage in the Add is instead in the `path`.

PR #597, opened 2022-02-15
Require Commit.path by default

### LINKED ISSUE #596: Invert definition of "partial" commit
The authoritative definition (in Section [13.2](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#section-13.2)) of when to include a `path` is not the same as the summary below. For example, if an implementation adds a custom proposal type, according to the summary, a path is required, whereas according to the definition above it, no path is required.

Ultimately, it depends on the nature of the proposal and the desired security guarantees if a path is required. However, I think the safer option is to require a path by default and maybe leave it to the definition of the respective custom proposal to explicitly specify if a path is _not_ required.

bifurcation:
Fixes #596 

This PR introduces a notion of a proposal type being "path safe", in the sense that it is safe for the path to be omitted.  (Better phrasing welcome!)  The `path` field is required default, and allowed to be omitted only if all the proposals in the Commit are path-safe.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3311,42 +3311,47 @@ of the Commit.  An Update proposal can be regarded as a "lazy" version of this
 operation, where only the leaf changes and intermediate nodes are blanked out.
 
 By default, the `path` field of a Commit MUST be populated.  The `path` field
-MAY be omitted if (a) it covers at least one proposal and (b) all proposals in
-the Commit are of "path safe" types.  A proposal type is "path safe" if it
-cannot change the group membership in a way that requires the forward secrecy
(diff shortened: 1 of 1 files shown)


kkohbrok:
Thanks for taking care of this fix! One suggestion: Instead of having a global list of which types are path safe and which ones aren't, we could make it a property of the proposal. Maybe something like this:

```
allPathSafe = true

for i, id in commit.proposals:
    proposal = proposalCache[id]
    assert(proposal != null)

    allPathSafe = allPathSafe && (proposal.is_path_safe)

if len(commit.proposals) == 0 || !allPathSafe:
    assert(commit.path != null)
```

PR #600, opened 2022-02-18
Add version to MLSMessage

### LINKED ISSUE #590: Add version to MLSMessage
To enable version agility, we need MLSMessages to indicate what version they are intended for.  Since version is included in signature labels, we need to deconflict these two indicators.

bifurcation:
Fixes #590

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1318,6 +1318,12 @@ Delivery Service to examine such messages.
 ~~~~~
 enum {
     reserved(0),
+    mls10(1),
+    (255)
+} ProtocolVersion;
+
+enum {
+    reserved(0),
     application(1),
     proposal(2),
     commit(3),
@@ -1371,6 +1377,7 @@ struct {
 } MLSMessageContent;
 
 struct {
+    ProtocolVersion version;
     WireFormat wire_format;
     select (MLSMessage.wire_format) {
         case mls_plaintext:
@@ -1850,9 +1857,9 @@ equivalent to using the key pair of the removed node.
 * Blank all the nodes on the direct path from the leaf to the root.
 * Generate a fresh HPKE key pair for the leaf.
 * Generate a sequence of path secrets, one for each node on the leaf's filtered direct
-  path, as follows. In this setting, `path_secret[0]` refers to the first parent node 
+  path, as follows. In this setting, `path_secret[0]` refers to the first parent node
   in the filtered direct path, `path_secret[1]` to the second parent node, and so on.
-  
+
   ~~~~~
   path_secret[0] is sampled at random
   path_secret[n] = DeriveSecret(path_secret[n-1], "path")
@@ -2125,7 +2132,7 @@ leaves was emptied. (Observe also that `original_child_resolution` contains all
 unmerged leaves of S.) Therefore, P's Parent Hash fixes, for each node V on the
 path from P to the root, not only the HPKE public key of V, but also the set of
 HPKE public keys to which the corresponding HPKE secret key of V was encrypted by
-the generator of the `UpdatePath`. 
+the generator of the `UpdatePath`.
 
 ### Using Parent Hashes
 
@@ -2757,12 +2764,6 @@ The signature is computed by the function `SignWithLabel` with a label
 signature field.
 
 ~~~~~
-enum {
-    reserved(0),
-    mls10(1),
-    (255)
-} ProtocolVersion;
-
 // See IANA registry for registered values
 uint16 ExtensionType;
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1401,11 +1401,11 @@ The following structure is used to fully describe the data transmitted in
 plaintexts or ciphertexts.
 
 ~~~~~
-struct MLSMessageContentAuth {
+struct {
     WireFormat wire_format;
     MLSMessageContent content;
     MLSMessageAuth auth;
-}
+} MLSMessageContentAuth;
 ~~~~~
 
 ## Content Authentication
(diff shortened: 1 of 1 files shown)


rohan-wire:
> Note that this is a little bit duplicative: There's already a `version` field in KeyPackage. But removing that would require something like sending MLSMessage in Add, which seems unpleasant. We should probably either require that `KeyPackage.version == MLSMessage.version`, or just not use MLSMessage for KeyPackage.
> 
> Likewise, the version field in GroupInfo can probably be removed, since it will be on the outside MLSMessage regardless of whether the GroupInfo is sent bare or in Welcome.

First, we need to add ProtocolVersion to the beginning of MLSMessageContentTBS. 

Regarding KeyPackage and GroupInfo, the semantics here are subtle.
Example: My MLS 1.1 client could send an MLSMessage with MLS version 1.1 to the DS/Directory, which contains a KeyPackage which only advertises MLS version 1.0. Another MLS client which only supports MLS 1.0 could read it with no problem. (I could publish both 1.0 and 1.1 KeyPackages).

bifurcation:
The latest commits here address the initial comments about KeyPackage and GroupInfo, and fixes the problem @rohan-wire notes about MLSMessageContentTBS.

For GroupInfo: I have just removed the `version` field.  I don't think there's any ambiguity created by moving it out to the MLSMessage wrapper.

For KeyPackage: I have left the `version` field and required it to be the same as the MLSMessage `version` field.  The thinking here is that these are saying the same thing to different audiences (the adder and everything else), so they should be consistent.

You could allow them to differ if you said that `MLSMessage.version` is only to indicate the format/semantics of the inner KeyPackage; then `KeyPackage.version` would indicate the actual version in use.  So to @rohan-wire's point, if  MLS 1.1 used the same KeyPackage format/semantics as MLS 1.0 (but presumably differed some other way), then you could have `MLSMessage{version=1.0, KeyPackage{version=1.1, ...}}`.  But this seems harder to reason about than just saying "If I don't support the outer version, I can't use this KeyPackage".

kkohbrok:
I'm confused about what we're trying to do here. The way I understand it is that MLS messages are always sent in the context of a group. That group has a version associated with it via the group context. I see how it would make sense to add the version field to MLS messages, but I don't see how that version has anything to do with a potential KeyPackage that might or might not be a part of that message. I think the spec already has enough requirements about versions: If the version indicated by an MLSMessage differs from that of the group indicated by the group id, it's invalid. Similarly, if the version of the KeyPackage contained in a given Add proposal doesn't match that of the group in which it's meant to be used, it's invalid. Where exactly do we run into trouble if we don't explicitly relate the version in the framing and that of it's contents? The versions should already be the same, since they have to be the same as that of the group.

Regarding the example with the DS: I don't quite understand the setting. As far as I'm aware, uploading or downloading of KeyPackages is not specified in the spec, so even if it happens within an MLS group (maybe the DS and the client are in a group) the KeyPackage would be part of an application message and thus the version of the KeyPackage would be irrelevant as far as the spec is concerned. Please correct me if I'm wrong or if I misunderstand the problem at hand.

rohan-wire:
Hi Konrad,
> The way I understand it is that MLS messages are always sent in the context of a group.
Not all MLS _objects_ are sent in the context of a group. KeyPackage objects are not. The version number applies to the _protocol_ which specifies the _objects_, some of which are messages.

When some lower level protocol sends an MLSCiphertext or MLSPlaintext message, the MLS client and/or DS could support multiple versions of MLS. The DS can't tell which group is being referenced until it can parse the message. It can't do that until it knows the protocol version, so the protocol and version are typically the first fields. 

I agree that for MLSCiphertext and MLSPlaintext objects if the version of the message doesn't correspond to the version of the group, that sounds like an invalid message to me.

(I don't see when you would put a KeyPackage inside an MLS application message.)

Even though uploading and downloading KeyPackage objects is not included in the spec, we need to allow another protocol to do so and give that protocol a reasonable migration path. If my directory supports KeyPackages in v1.0 and v1.1 format, we should be able to publish both versions until all the v1.0 potential peers have migrated. The format of the KeyPackage objects may be different. Being able to distinguish these so they can be parsed without external help is a basic data-hygiene function.

rohan-wire:
>So to @rohan-wire's point, if MLS 1.1 used the same KeyPackage format/semantics as MLS 1.0 (but presumably differed some other way), then you could have `MLSMessage{version=1.0, KeyPackage{version=1.1, ...}}`. But this seems harder to reason about than just saying "If I don't support the outer version, I can't use this KeyPackage".

MLSMessage{version=1.1, KeyPackage{version=1.0,...}} is the case I am envisioning. My client supports both MLS version 1.1 and 1.0. My DS supports v1.1 so I always communicate with v1.1, but I publish keypackages in both 1.0 and 1.1 formats.

kkohbrok:
Thanks for the explanation, that makes the intent a bit clearer to me. However, I still don't see the scenario where something like `MLSMessage{version=1.1, KeyPackage{version=1.0,...}}` is a valid message in the context of the MLS spec. I agree that you should be able to publish KeyPackages of multiple versions and that version agility is important. 

Do I understand correctly, that in the scenario you describe (`MLSMessage{version=1.1, KeyPackage{version=1.0,...}}`) you're uploading a KeyPackage to the DS? I guess I don't understand why that would be an MLSMessage and not some message in the proprietary protocol you use to upload KeyPackages.

bifurcation:
@rohan-wire - You don't need `MLSMessage{mls11, KeyPackage{mls10,...}})` to allow for uploading both 1.0 and 1.1 key packages -- you just have the client upload both `MLSMessage{mls10, KeyPackage{mls10,...}})` and `MLSMessage{mls11, KeyPackage{mls11,...}})`.

Since there doesn't seem to be a compelling case here for allowing the versions to differ, I'm going to go ahead and merge with the requirement that they be the same.

PR #601, opened 2022-02-18
Remove dangling references to PublicGroupState

bifurcation:
PublicGroupState was removed as an independent struct in #529.  This PR fixes some dangling references to it.


rohan-wire (changes requested):
Also, update WireFormat:

s/mls_public_group_state(4),/mls_group_info(4),/

PR #602, opened 2022-02-18
Limit the new variable length encoding to 30-bit

raphaelrobert:
The recently introduced variable size encoding allows for very large message sizes that are not practical in any realistic scenario. Since decoders will most likely not allow such large messages anyway, this PR limits the maximum size to 30-bit.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -483,9 +483,9 @@ This means that integers are encoded on 1, 2, or 4 bytes and can encode 6-,
 
 Vectors that start with "11" prefix are invalid and MUST be rejected.
 
-For example, the four byte sequence 9d 7f 3e 7d decodes to 494878333; 
-the two byte sequence 7b bd decodes to 15293; and the single byte 25 
-decodes to 37 (as does the two byte sequence 40 25).
+For example, the four byte sequence 0x9d7f3e7d decodes to 494878333; 
+the two byte sequence 0x7bbd decodes to 15293; and the single byte 0x25 
+decodes to 37.
 
 The following figure adapts the pseudocode provided in {{RFC9000}} to add a
 check for minimum-length encoding:
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -462,7 +462,7 @@ struct {
 } StructWithVectors;
 ~~~~~
 
-Such a vector can represent values with length from 0 bytes to 2^62 bytes.
+Such a vector can represent values with length from 0 bytes to 2^30 bytes.
 The variable-length integer encoding reserves the two most significant bits
 of the first byte to encode the base 2 logarithm of the integer encoding length
 in bytes.  The integer value is encoded on the remaining bits, in network byte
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -462,29 +462,30 @@ struct {
...
 
-For example, the eight-byte sequence c2 19 7c 5e ff 14 e8 8c (in hexadecimal)
-decodes to the decimal value 151288809941952652; the four byte sequence 9d 7f 3e
-7d decodes to 494878333; the two byte sequence 7b bd decodes to 15293; and the
-single byte 25 decodes to 37 (as does the two byte sequence 40 25).
+Vectors that start with "11" prefix are invalid and MUST be rejected.
+
+For example, the four byte sequence 9d 7f 3e 7d decodes to 494878333; 
+the two byte sequence 7b bd decodes to 15293; and the single byte 25 
+decodes to 37 (as does the two byte sequence 40 25).
```
bifurcation, line 488:
```suggestion
For example, the four byte sequence 0x9d7f3e7d decodes to 494878333; 
the two byte sequence 0x7bbd decodes to 15293; and the single byte 0x25 
decodes to 37.
```

* Added `0x` to disambiguate hex vs. decimal
* Removed the final parenthetical because it doesn't obey minimum-length encoding.

draft-ietf-mls-protocol.md
```diff
@@ -462,29 +462,30 @@ struct {
...
+| 01     | 2       | 14          | 64         | 16383               |
+| 10     | 4       | 30          | 16384      | 1073741823          |
+| 11     | invalid | -           | -          | -                   |
 {: #integer-summary title="Summary of Integer Encodings"}
 
-For example, the eight-byte sequence c2 19 7c 5e ff 14 e8 8c (in hexadecimal)
-decodes to the decimal value 151288809941952652; the four byte sequence 9d 7f 3e
-7d decodes to 494878333; the two byte sequence 7b bd decodes to 15293; and the
-single byte 25 decodes to 37 (as does the two byte sequence 40 25).
+Vectors that start with "11" prefix are invalid and MUST be rejected.
```
bifurcation, line 484: An alternative approach here would be to leave the `11` prefix, but forbid encoded values larger than some limit.  That way, if there were some case where reusing this encoding would be useful (cTLS, maybe) and larger sizes were allowed, then it could just be reused.  It also just seems a bit ugly to encode range limitations in the formatting.

Suggest something like the following:


> The `11` prefix is included for completeness.  To avoid compatibility issues for devices with limited storage, however, vectors in MLS MUST NOT have a length greater than 2^32.  A vector with an encoded length greater than this limit MUST be rejected as malformed.

bifurcation:
cc @rohan-wire who had some thoughts about limits on the last virtual interim.

rohan-wire:
> cc @rohan-wire who had some thoughts about limits on the last virtual interim.
My general points are:
- Code defensively
- Fail gracefully
- Use the lower level protocol carrying MLS Messages to send a clear and relevant error message if something is too big.
   Ex: HTTP/2.0 413 Commit.Proposals exceeds 2^16 bytes

bifurcation (approved):

PR #603, opened 2022-02-18
Update parent hash validation

bifurcation:
This PR revises the parent hash verification section to reflect more recent discussion.  Right now, the definition of validity according to the parent hash ties it to the "top-down" validation algorithm, when what really matters is that each parent node has a path to a leaf.  This PR changes the definition to that more general version, and mentions the "top-down" and "bottom-up" approaches to checking this property.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2179,13 +2179,15 @@ hash is valid with respect to P\_2, and so on.
 When joining a group, the new member MUST authenticate that each non-blank
 parent node P is parent-hash valid.  This can be done "bottom up" by building
 chains up from leaves and verifying that all non-blank parent nodes are covered
-by such chains, or "top down" by verifying that there is exactly one
+by exactly one such chain, or "top down" by verifying that there is exactly one
 descendant of each non-blank parent node for which the parent node is
 parent-hash valid.
 
-When processing a Commit message, clients MUST recompute the expected value of
-`parent_hash` for the committer's new leaf and verify that it matches the
-`parent_hash` value in the supplied `leaf_node`.
+When processing a Commit message that includes an UpdatePath, clients MUST
+recompute the expected value of `parent_hash` for the committer's new leaf and
+verify that it matches the `parent_hash` value in the supplied `leaf_node`.
+After being merged into the tree, the nodes in the UpdatePath form a parent-hash
+chain from the committer's leaf to the root.
 
 ## Update Paths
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2159,13 +2159,33 @@ at V. Indeed, such a ratchet tree would violate the tree invariant.
...
+hash is valid with respect to P\_2, and so on.
+
+When joining a group, the new member MUST authenticate that each non-blank
+parent node P is parent-hash valid.  This can be done "bottom up" by building
+chains up from leaves and verifying that all non-blank parent nodes are covered
+by such chains, or "top down" by verifying that there is exactly one
+descendant of each non-blank parent node for which the parent node is
+parent-hash valid.
+
+When processing a Commit message, clients MUST recompute the expected value of
```
TWal, line 2186: Maybe add something to say the nodes on the path of the commit form a new parent-hash-chain?

TWal (commented):
Note that this doesn't completely address the "unease" feeling mentioned in #587, but I can live it.

Other than that, I think it makes the parent hash construct a lot more clear and this PR has a lot of pedagogical value. The section on parent hash in the current document is really difficult to grasp for a newcomer.

MartaMularczyk:
Great proposal! It achieves what we want much more directly.

One comment: in the description of the "top down" approach, the part "there is _exactly_ one descendant" isn't exactly consistent with being "parent hash valid", because the latter requires that P can be chained to _at least_ one leaf. The bottom up approach is consistent with being parent hash valid.
As a result, a malicious insider may be able to craft a tree that is accepted by the bottom up approach but not by the top down approach [1] which may hurt interoperability.
I'd suggest either to change "exactly one" to "at least one" in the top down approach or to change "can be chained back to a leaf" to "can be chained back to exactly one leaf" in the definition of parent hash valid (and modify the bottom up approach accordingly). I don't have a strong opinion which is better.

---------------------
[1] An example such tree: the tree has 3 nodes: the root P and its children A and B. Assuming both A and B are malicious, A signs the parent hash of P with copath child B and B signs the parent hash of P with copath child A. (This is possible with the current resolution-hash for parent-hash.)

bifurcation:
@MartaMularczyk - I am keen on the "exactly one" framing precisely because you shouldn't have these collisions in a real tree.  At least I haven't been able to think up a case; maybe you have a non-malicious case?  Given that, I'm inclined to tighten the "bottom-up" so that it fails if there are multiple valid paths to a node.

bifurcation:
Just to close the loop here:

* @TWal - thanks for the suggestion, added a sentence
* @MartaMularczyk - Updated the "bottom up" description to say that each node should be covered by exactly one chain, which is the parallel requirement to exactly one child.

PR #609, opened 2022-03-07
Rename generic LeafNode object to ClientInfo and move ahead of # Ratchet Tree Operations

rohan-wire:
The LeafNode text is self-contradictory when it tries to refer to KeyPackages. To avoid these problems, I wrote a version with a separate ClientInfo object (basically the LeafNode object today) and put the generic stuff there. There is an alias for LeafNode to ClientInfo and the KeyPackage uses ClientInfo as one of its components. Stuff which is unique to the usage as a LeafNode or KeyPackage is in the relevant section.

Comments welcome. Probably more than one way to do resolve this, but the text now has lots of duplication.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1632,46 +1632,32 @@ When parsing a SenderData struct as part of message decryption, the recipient
 MUST verify that the LeafNodeRef indicated in the `sender` field identifies a
 member of the group.
 
-# Ratchet Tree Operations
+# ClientInfo Object
 
-The ratchet tree for an epoch describes the membership of a group in that epoch,
-providing public-key encryption (HPKE) keys that can be used to encrypt to subsets of
-the group as well as information to authenticate the members.  In order to
-reflect changes to the membership of the group from one epoch to the next,
-corresponding changes are made to the ratchet tree.  In this section, we
-describe the content of the tree and the required operations.
-
-## Ratchet Tree Node Contents
-
-As discussed in {{ratchet-tree-nodes}}, the nodes of a ratchet tree contain
-several types of data describing individual members (for leaf nodes) or
-subgroups of the group (for parent nodes).  Parent nodes are simpler:
-
-~~~~~
-struct {
-    HPKEPublicKey public_key;
-    opaque parent_hash<V>;
-    uint32 unmerged_leaves<V>;
-} ParentNode;
-~~~~~
-
-The `public_key` field contains a HPKE public key whose private key is held only
-by the members at the leaves among its descendants.  The `parent_hash` field
-contains a hash of this node's parent node, as described in {{parent-hash}}.
-The `unmerged_leaves` field lists the leaves under this parent node that are
-unmerged, according to their indices among all the leaves in the tree.
-
-A leaf node in the tree describes all the details of an individual client's
-appearance in the group, signed by that client:
+The ClientInfo object contains instance information about a client which is used
+both in the context of group membership (described as a LeafNode object) and in
+offers for potential membership (each called a KeyPackage) used by other clients
+to add the issuer of the KeyPackage to the other clients' own groups.  LeafNode
+objects are used specifically inside Add and Update proposals, and Commits, all of
+three of which introduce new key material into a group.
 
 ~~~~~
 enum {
     reserved(0),
     key_package(1),
-    update(2),
-    commit(3),
+    add(2),
+    update(3),
+    commit(4),
     (255)
-} LeafNodeSource;
+} ClientInfoSource;
+
+// See IANA registry for registered values
+uint16 ExtensionType;
+
+struct {
+    ExtensionType extension_type;
+    opaque extension_data<V>;
+} Extension;
 
 struct {
     ProtocolVersion versions<V>;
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1698,85 +1684,110 @@ struct {
     HPKEPublicKey public_key;
     Credential credential;
     Capabilities capabilities;
+    Extension extensions<V>;
```
bifurcation, line 1687: I don't love the aesthetics of this.  I would like to maintain the invariants that (a) extensions are always at the end of the message and (b) FooTBS is always clearly an extension of Foo.  You've done (b) here, but it's a little hidden in the `select`.  I think the original structs here are clearer, if a bit chattier.

draft-ietf-mls-protocol.md
```diff
@@ -1641,46 +1641,32 @@ When parsing a SenderData struct as part of message decryption, the recipient
...
-    uint32 unmerged_leaves<V>;
-} ParentNode;
-~~~~~
-
-The `public_key` field contains a HPKE public key whose private key is held only
-by the members at the leaves among its descendants.  The `parent_hash` field
-contains a hash of this node's parent node, as described in {{parent-hash}}.
-The `unmerged_leaves` field lists the leaves under this parent node that are
-unmerged, according to their indices among all the leaves in the tree.
+# ClientInfo Object
```
bifurcation, line 1644: I would probably keep this under `# Ratchet Tree Node Contents`, even if we're going to rename to `ClientInfo`.  In general, struct are introduced in context, e.g., GroupInfo and Welcome are introduced in the context of the relevant join operations.

draft-ietf-mls-protocol.md
```diff
@@ -1698,85 +1684,110 @@ struct {
...
 protocol extensions, and non-default proposal types are supported by a client.
 Proposal types defined in this document are considered "default" and thus need
-not be listed.  Extensions that appear in the `extensions` field of a LeafNode
+not be listed.  Extensions that appear in the `extensions` field of a ClientInfo
 MUST be included in the `extensions` field of the `capabilities` field.
 
-The `leaf_node_source` field indicates how this LeafNode came to be added to the
-tree.
+The `client_info_source` field indicates the context in which the ClientInfo
+object is used.
```
bifurcation, line 1739: This is incorrect.  The ClientInfo is used as a leaf in the tree for TreeKEM, regardless of how it got there.  This field says how it got there, thus `source`.

draft-ietf-mls-protocol.md
```diff
@@ -1698,85 +1684,110 @@ struct {
...
+between which clients will consider the KeyPackage or LeafNode (public_key and credential) valid.  These
+times are represented as absolute times, measured in seconds since the
+Unix epoch (1970-01-01T00:00:00Z).  Applications MUST define a maximum total
+lifetime that is acceptable for a ClientInfo object, and reject any ClientInfo
+object where the total lifetime is longer than this duration.
 
-In the case where the leaf node was inserted into the tree via a Commit message,
-the `parent_hash` field contains the parent hash for this leaf node (see
-{{parent-hash}}).
+In the case where the ClientInfo refers to a leaf node inserted into the tree
```
bifurcation, line 1748:
```suggestion
In the case where the ClientInfo represents a leaf node inserted into the tree
```

draft-ietf-mls-protocol.md
```diff
@@ -1814,19 +1823,19 @@ The client verifies the validity of a LeafNode using the following steps:
 
 * Verify the `lifetime` field:
   * When validating a downloaded KeyPackage, the current time MUST be within the
-    `lifetime` range.  A KeyPackage that is expired or not yet valid MUST NOT be
+    `lifetime` range.  A ClientInfo that is expired or not yet valid MUST NOT be
```
bifurcation, line 1826: Should this be LeafNode, to be consistent with the above?

draft-ietf-mls-protocol.md
```diff
@@ -2882,83 +2897,59 @@ struct {
...
-Proposal types defined in this document are considered "default" and thus need
-not be listed.
-
-~~~~~
-struct {
-    ProtocolVersion versions<V>;
-    CipherSuite ciphersuites<V>;
-    ExtensionType extensions<V>;
-    ProposalType proposals<V>;
-} Capabilities;
```
bifurcation, line 2936: You removed this and never re-added it.  It needs to be defined somewhere.

draft-ietf-mls-protocol.md
```diff
@@ -2882,83 +2897,59 @@ struct {
...
-in the `extensions` field of a KeyPackage MUST be included in the `extensions`
-field of the `capabilities` extension.
-
 ## Lifetime
 
-The `lifetime` extension represents the times between which clients will
-consider a KeyPackage valid.  This time is represented as an absolute time,
-measured in seconds since the Unix epoch (1970-01-01T00:00:00Z). A client MUST
+The `client_info.lifetime` field represents the times between which clients will
+consider a KeyPackage valid.  A client MUST
```
bifurcation, line 2944: This trails off.

bifurcation: In any case, this section should be deleted, because Lifetime isn't an extension any more.

bifurcation (commented):
Bunch of comments inline.  Overall, I think there are a bunch of helpful clarifications in here around KeyPackages, but I'm not sure the LeafNode -> ClientInfo rename is worthwhile, since we end up having to refer to LeafNodes anyway.

draft-ietf-mls-protocol.md
rohan-wire: The idea of how it got there being relevant even days later is a little weird, so adding an extra sentence or two explaining this would be useful. I started originally trying to do a lighter edit, and trying to say something reasonable that applies to KeyPackages, Proposals, and Commits without tripping over myself was a challenge.

draft-ietf-mls-protocol.md
rohan-wire: Yours is better. What do you think of:
...represents a leaf node once it is inserted into the tree.
?

draft-ietf-mls-protocol.md
rohan-wire: There was one place in the KeyPackage section where we were talking about lifetimes. In that case, I think the lifetime originally referred to the lifetime of the KeyPackage.

draft-ietf-mls-protocol.md
rohan-wire: oops. My intent was only that it not appear in KeyPackages.

PR #610, opened 2022-03-07
Explicitly allow for deniable signature keys

kkohbrok:
So far, deniable signature keys were permitted by the description of how to perform signature keys in the Content Authentication section, but not by the description of how to verify these signatures in the "Processing a Commit" section. This PR explicitly introduces the deniable signature keys and makes the handling of signatures consistent in both sections.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1268,6 +1268,23 @@ client presents a new credential, the application MUST verify that the set
 of identifiers in the credential is acceptable to the application for this
 client.
 
+### Deniable Signature Keys
+
+Some 1-to-1 messaging protocols such as Signal achieve varying degrees of
+plausible deniability by leveraging the symmetry of the Diffie-Hellman key
+exchange.
+
+MLS uses signatures to achieve strong authentication of messages (see Section
+{{content-authentication}}). Since these signatures do not allow for plausible
+deniability, signatures on MLS messages can either be performed using the
+private key corresponding to the signature public key in the sender's
+credential, or by using another signature key that is not directly tied to the
+identity of the sender, thus allowing for some degree of deniability.
+
+Such deniable signature keys MUST be distributed and authenticated to and by
+each member out-of-band prior to any use for signature verification to ensure
+that MLS messages are still strongly authenticated.
+
 ### Uniquely Identifying Clients
 
 MLS implementations will presumably provide applications with a way to request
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1268,6 +1268,23 @@ client presents a new credential, the application MUST verify that the set
...
+MLS uses signatures to achieve strong authentication of messages (see Section
+{{content-authentication}}). Since these signatures do not allow for plausible
+deniability, signatures on MLS messages can either be performed using the
+private key corresponding to the signature public key in the sender's
+credential, or by using another signature key that is not directly tied to the
+identity of the sender, thus allowing for some degree of deniability.
+
+Such deniable signature keys MUST be distributed and authenticated to and by
+each member out-of-band prior to any use for signature verification to ensure
+that MLS messages are still strongly authenticated.
```
raphaelrobert, line 1286: Given that this mostly has to do with how the AS works (and therefore it's out-of-scope for the spec), it might be more concise and also sufficient to say that signature keys must be authenticated by the AS, and cite the distribution of signature keys over a deniable channel as an example.

bifurcation (changes requested):
I strongly object to this PR.  This is WONTFIX in my view.

This is introducing new ambiguity in the spec that is not there now.  Right now, the spec is clear that the keys used for signatures are the ones in the credential.

IIUC the confusion here hinges on the following change from #437: Changing "private key corresponding to the credential" to "private key corresponding to the public key authenticated by the credential".

The current PR seems to assume that this allows the public key used to verify signatures to differ from the public key in the credential.  I don't see any plausible reading where that is the case.  Consider X.509 certificates -- the public key authenticated by the certificate is the public key in the certificate.

If other folks are unclear on that, we should clarify in the opposite direction, probably just by reverting #437.

kkohbrok:
#437 didn't happen in a vacuum, but was merged after a discussion on the mailing list with the intent to leave a gap to allow for "deniable" keys.

To make sure that we're not claiming anything that we can't hold, I've changed the PR to not explicitly talk about "deniable" keys, but instead about "alternative" keys, the properties of which depend on their mode of distribution.

Another addition (as suggested by @bifurcation oob) that I intend to make is that of an explicit group context extension that ensures that group members agree on the use of these types of keys.

kkohbrok:
I realize now that "alternative keys" is probably not the best name... but again, that can be changed if we agree that this is something we want.

bifurcation:
Closing in favor of #617 in light of mailing list discussion

PR #614, opened 2022-03-07
More editorial changes from # Ratchet Tree Ops through # Group Evolution

rohan-wire:


draft-ietf-mls-protocol.md
```diff
@@ -3085,12 +3090,13 @@ considered invalid.
 # Group Evolution
 
 Over the lifetime of a group, its membership can change, and existing members
-might want to change their keys in order to achieve post-compromise security. In
-MLS, each such change is accomplished by a two-step process:
+might want to change their keys in order to achieve post-compromise security.
+In MLS, changes can be made immediately using a Commit message with embedded
+Proposals, such changes can be accomplished via a two-step process:
```
bifurcation, line 3095:
```suggestion
In MLS, each such change is accomplished by a two-step process:
```

Since every change requires a proposal, stating this as Proposal/Commit first, then one-step below.

draft-ietf-mls-protocol.md
```diff
@@ -3085,12 +3090,13 @@ considered invalid.
...
 
 1. A proposal to make the change is broadcast to the group in a Proposal message
-2. A member of the group or a new member broadcasts a Commit message that causes one or more
-   proposed changes to enter into effect
+2. A member of the group or a new member broadcasts a Commit message that causes
+   one or more proposed changes to enter into effect
```
bifurcation, line 3099:
```suggestion
2. A member of the group or a new member broadcasts a Commit message that causes
   one or more proposed changes to enter into effect
   
In cases where the Proposal and Commit are sent by the same member, these two steps
can be combined by sending the proposals in the commit.
```

draft-ietf-mls-protocol.md
```diff
@@ -3387,13 +3393,10 @@ Proposals. It instructs group members to update their representation of the
...
-supplied by value are included directly in the Commit object.  Proposals
-supplied by reference are specified by including the hash of the MLSPlaintext in
-which the Proposal was sent (see {{hash-based-identifiers}}).  For proposals
-supplied by value, the sender of the proposal is the same as the sender of the
-Commit.  Conversely, proposals sent by people other than the committer MUST be
-included by reference.
+identifies the proposal to be applied by value or by reference.  Commits that
+refer to new Proposals from the committer can be included by value. Commits
+for previously sent proposals from anyone (including the committer) can be sent
+by reference.
```
bifurcation, line 3399:
```suggestion
identifies the proposal to be applied by value or by reference.  Commits that
refer to new Proposals from the committer can be included by value. Commits
for previously sent proposals from anyone (including the committer) can be sent
by reference.  Proposals sent by reference are specified by including the hash of
the MLSPlaintext in which the proposal was sent (see {{hash-based-identifiers}}).
```

draft-ietf-mls-protocol.md
```diff
@@ -3474,7 +3477,7 @@ The only proposal types defined in this document that do not require a path are:
 * `reinit`
 
 New proposal types MUST state whether they require a path. If any instance of a
-proposal type requires a path, then the proposal type requires a path. This
+proposal type requires a path, then the Commit requires a path. This
```
bifurcation, line 3480:
```suggestion
proposal type requires a path, then the proposal type requires a path. This
```

Reverting because the emphasis here is on the "requires path" attribute being at the level of proposal types.

draft-ietf-mls-protocol.md
```diff
@@ -3547,7 +3550,8 @@ message at the same time, by taking the following steps:
   field populated from Proposals received during the current epoch, and an empty
   `path` field.
 
-* Generate the provisional ratchet tree and GroupContext by applying the proposals
+* Generate the provisional ratchet tree and GroupContext by applying the valid,
+  compatible proposals
```
bifurcation, line 3554:
```suggestion
* Generate the provisional ratchet tree and GroupContext by applying the proposals
```

Reverting because the choice of proposals happens earlier.  All proposals included in the initial commit need to be applied.

draft-ietf-mls-protocol.md
```diff
@@ -3580,8 +3584,8 @@ message at the same time, by taking the following steps:
      `commit_secret` as the value `path_secret[n+1]` derived from the
      `path_secret[n]` value assigned to the root node.
 
-* If not populating the `path` field: Set the `path` field in the Commit to the
-  null optional.  Define `commit_secret` as the all-zero vector of length
+* If not populating the `path` field: Set the `path` field in the Commit to a
+  zero-length vector.  Define `commit_secret` as the all-zero vector of length
```
bifurcation, line 3588:
```suggestion
* If not populating the `path` field: Set the `path` field in the Commit to the
  null optional.  Define `commit_secret` as the all-zero vector of length
```

Reverting because the `path` field has type `optional<UpdatePath>`.

draft-ietf-mls-protocol.md
```diff
@@ -3634,8 +3638,8 @@ message at the same time, by taking the following steps:
   parameters:
   * `psktype`: `resumption`
   * `usage`: `reinit`
-  * `group_id`: The group ID for the current group
-  * `epoch`: The epoch that the group will be in after this Commit
+  * `group_id`: The group ID for the new group
+  * `epoch`: The epoch that the group will be in after this Commit (0x0)
```
bifurcation, line 3642:
```suggestion
  * `group_id`: The group ID for the current group
  * `epoch`: The epoch that the group will be in after this Commit
```

Reverting because the PSK here refers to the last epoch of the earlier group.  So it has to use the old group's information.

draft-ietf-mls-protocol.md
```diff
@@ -3797,14 +3801,18 @@ The full tree can be included via the `ratchet_tree` extension
 The `signature` on the GroupInfo struct authenticates the HPKE public key, so
 that the joiner knows that the public key was provided by a member of the group.
 The fields that are not signed are included in the key schedule via the
-GroupContext object. If the joiner is provided an inaccurate data for these
+GroupContext object. If the joiner has provided inaccurate data for these
```
bifurcation, line 3804: This is wrong, because we're talking about GroupInfo that the joiner consumes.  But this whole paragraph is wrong; will remove in a subsequent commit.

bifurcation (approved):

bifurcation (approved):

PR #617, opened 2022-03-09
Preconfigured senders extension and removal of signature key indirection

### LINKED ISSUE #616: Remove or refine preconfigured senders
As [Raphael pointed out on the mailing list](https://mailarchive.ietf.org/arch/msg/mls/VWzhGx2XwgBqHoHIE0fvuZS1sm4/), the current preconfigured sender mechanism isn't interoperably specified.  There is no confirmation that the members of a group agree on who the allowed preconfigured senders are, so different members might respond differently.  It seems like there are two options here:

1. Add the required details.  We would need a `preconfigured_senders` extension to confirm agreement to use the mechanism at all, and would probably need to have a list of senders there as well.

2. Remove the feature from the base protocol.  Since an extension is required in any case, there's no pressing need to have the feature in the base protocol.

Personally, I would lean toward (2).  Early deployment experience doesn't show a lot of need for this feature.  For example, in Webex, we just have the server forward a KeyPackage instead of synthesizing an Add proposal under a preconfigured signer; the TLS authentication is as good as a preconfigured signer signature.  As long as the SenderType enum is around, the extension should be a small follow-on spec if it is really needed.

kkohbrok:
This PR:
- consolidates instructions on which key should be used to compute (and verify) a signature on an MLS message depending on the sender type
- reverts #437 
- introduces a `preconfigured_senders` extension that ensures that group members agree on preconfigured signature keys, as discussed in #616 and on the mailing list

Fixes #616.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1490,7 +1490,9 @@ depending on the sender's `sender_type`:
   sender's `LeafNodeRef`
 * `external`: The signature key contained in the Credential at the index
   indicated by the `sender_index` in the `external_senders` group context
-  extension (see Section {{external-senders-extension}}).
+  extension (see Section {{external-senders-extension}}).  In this case, the
+  `content_type` of the message MUST NOT be `commit`, since only members
+  of the group or new joiners can send Commit messages.
 * `new_member`: The signature key depends on the `content_type`:
   * `proposal`: The signature key in the credential contained in KeyPackage in
     the Add proposal (see Section {{external-proposals}}).
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1378,7 +1378,7 @@ enum {
 enum {
     reserved(0),
     member(1),
-    preconfigured(2),
+    external(2),
     new_member(3),
     (255)
 } SenderType;
@@ -1388,8 +1388,8 @@ struct {
     switch (sender_type) {
         case member:
             LeafNodeRef member_ref;
-        case preconfigured:
-            u32 sender_id;
+        case external:
+            u32 sender_index;
         case new_member:
             struct{};
     }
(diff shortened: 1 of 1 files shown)


bifurcation (approved):
Thanks @kkohbrok, this looks good to me.

I wonder if it would be better to change "preconfigured" to "external". They're not really "preconfigured" in the sense of being "configured out of band", they're merely additional signers who are not in the group.

draft-ietf-mls-protocol.md
```diff
@@ -1477,14 +1477,25 @@ struct {
...
+current epoch, so that signatures are specific to a given group and epoch.
+
+The sender MUST use the private key corresponding to the following signature key
+depending on the sender's `sender_type`:
+
+* `member`: The signature key contained in the Credential at the leaf with the
+  sender's `LeafNodeRef`
+* `external`: The signature key contained in the Credential at the index
+  indicated by the `sender_index` in the `external_senders` group context
+  extension (see Section {{external-senders-extension}}).
```
bifurcation, line 1493:
```suggestion
* `external`: The signature key contained in the Credential at the index
  indicated by the `sender_index` in the `external_senders` group context
  extension (see Section {{external-senders-extension}}).  In this case, the
  `content_type` of the message MUST NOT be `commit`, since only members
  of the group or new joiners can send Commit messages.
```

bifurcation: Could also be phrased as "MUST be only `proposal` / `application`.  Or could be added below these bullets.

draft-ietf-mls-protocol.md
```diff
@@ -1388,8 +1388,8 @@ struct {
     switch (sender_type) {
         case member:
             LeafNodeRef member_ref;
-        case preconfigured:
-            opaque sender_id<V>;
+        case external:
+            u32 sender_index;
```
bifurcation, line 1392:
```suggestion
            uint32 sender_index;
```

PR #619, opened 2022-03-21
Clean up code blocks

bifurcation:
A few changes to make code blocks a little more consistent / usable / pretty:

* Fence every code block with exactly three tildes
* Mark each code block with a type (`aasvg` / `ascii-art` / `pseudocode` / `tls` / `python`)

Marking with `aasvg` generates slightly prettier pictures for the ladder diagrams in the introduction.  Marking with `tls` makes it easy to programmatically extract the TLS syntax definitions in case we want to produce a syntax summary appendix like [the one TLS has](https://datatracker.ietf.org/doc/html/rfc8446#appendix-B), or in case implementers want to quickly get a listing of all the structs they need.  

The only substantive changes I made here were:

1. Removing the `ReInit` lines from the branch and re-inject PSK illustrations
2. Splitting pseudocode from TLS syntax in a few cases.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -476,7 +476,7 @@ An optional value is encoded with a presence-signaling octet, followed by the
 value itself if present.  When decoding, a presence octet with a value other
 than 0 or 1 MUST be rejected as malformed.
 
-~~~
+~~~ tls
 struct {
     uint8 present;
     select (present) {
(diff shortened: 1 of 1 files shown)


TWal (approved):
Nice idea! It would also allow someone to compile the document with syntax coloration for TLS formats and Python code.

bifurcation:
Yeah, I specifically had the idea because I wanted to pull out the TLS syntax blocks to update my implementation :)

PR #620, opened 2022-03-25
Add make target to extract TLS stanzas

rohan-wire:
Added a short python program to extract the TLS presentation language stanzas into a single file.
Added extract-tls make target to run it.

An equivalent one liner option (only in the Makefile) is commented out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,3 +12,9 @@ endif
 
 fix-insecure-links:
 	sed -i'.bak' -e 's/http:/https:/g' draft-ietf-mls-protocol.html
+
+extract-tls:
+	cat draft-ietf-mls-protocol.md | python3 extract-tls.py > draft-ietf-mls-protocol.tls
+
+#extract-tls:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -12,3 +12,9 @@ endif
 
 fix-insecure-links:
 	sed -i'.bak' -e 's/http:/https:/g' draft-ietf-mls-protocol.html
+
+extract-tls:
+	cat draft-ietf-mls-protocol.md | python3 extract-tls.py > draft-ietf-mls-protocol.tls
+
+#extract-tls:
+#	cat draft-ietf-mls-protocol.md | python3 -c 'exec("""import sys, re\nmatch = False\nfor line in sys.stdin:\n if match:\n  if re.match("^~~~$", line):\n   match = False\n   print ("")\n  else:\n   print (line.rstrip())\n elif re.match("^~~~ tls$", line):\n  match = True\n""")' > draft-ietf-mls-protocol.tls
```
bifurcation, line 20:
```suggestion
```

Having a separate script is OK with me.

bifurcation (approved):

Issue #623, opened 2022-03-26
Subgroup branching bug


tomleavy:
There is currently a logical flaw in the subgroup branching flow.

In order to branch, you need to compose a `Welcome` message . A Welcome message includes `EncryptedGroupSecrets` which is defined as follows: 

```
struct {
  KeyPackageRef new_member;
  HPKECiphertext encrypted_group_secrets;
} EncryptedGroupSecrets;
```

The problem with this is that the `new_member` field of type `KeyPackageRef` cannot be made for existing members of the group because the leaf nodes are now `LeafNode` types. `KeyPackageRef` is now only used to add new members to the group, so we can't reuse this for sub group initialization.

kkohbrok:
I'm not sure I understand correctly. Are you suggesting when branching a group off of another group, one re-uses the leaves of the old group? If so, that's not how it's meant to work. The new group created in the process of subgroup branching should be independent from the "parent" group safe for the PSK, which all member have to provide upon creation of the group. In fact one should never use the same key material in the leaves of multiple groups except when the supply of key packages that a party provides via the DS is exhausted (see Section 17.4).

tomleavy:
That was part of my question. When there were key packages in the leaves, you theoretically could fork the group from a given point in time and reuse the current tree, but I wasn't sure if that was a proper solution or not. Sounds like it isn't, but either way, I think there should be clarification in the document about this. Let me write up a quick PR on it.

rohan-wire:
> I'm not sure I understand correctly. Are you suggesting when branching a group off of another group, one re-uses the leaves of the old group? If so, that's not how it's meant to work. The new group created in the process of subgroup branching should be independent from the "parent" group safe for the PSK, which all member have to provide upon creation of the group. In fact one should never use the same key material in the leaves of multiple groups except when the supply of key packages that a party provides via the DS is exhausted (see Section 17.4).

@kkohbrok Konrad, could you safely derive new keys for the members of the subgroup instead? If so, that would be an enormous increase in efficiency and (especially in the federated case) make group creation timing consistent enough to do dynamically.

kkohbrok:
[...]

PR #624, opened 2022-03-28
Sub-group branching requires fresh key packages for each member

tomleavy:
Based on #623 , it should be made explicit that sub-group branching requires a fresh key package for each member. I think there is still a conflict of wording here that I would like feedback on because of this line (which is what triggered me thinking there was a bug)

```
Each LeafNode in the new group's tree MUST be a leaf in the old group's tree at the epoch indicated in the PreSharedKey
```

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3145,7 +3145,7 @@ the referenced group.
 * The `version` and `ciphersuite` values in the Welcome MUST be the same as
   those used by the old group.
 * Each LeafNode in a new subgroup MUST match some LeafNode in the original
-  group. In this context, a pair of LeafNodes are said to "match" if the
+  group. In this context, a pair of LeafNodes is said to "match" if the
   identifiers presented by their respective credentials are considered
   equivalent by the application.
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3128,9 +3128,15 @@ considered invalid.
 ## Sub-group Branching
 
 A new group can be formed from a subset of an existing group's members, using
-the same parameters as the old group.  The creator of the group indicates this
-situation by including a PreSharedKey of type `resumption` with usage `branch`
-in the Welcome message that creates the branched subgroup.
+the same parameters as the old group.  
+
+A member can create a sub-group by performing the following steps:
+
+1. Determine a subset of existing members that should be a part of the sub-group. 
+2. Create a new tree for the sub-group by fetching a new KeyPackage for each 
+   existing member that should be included in the sub-group.
+3. Create a Welcome message that includes a PreSharedKey of type `resumption` with 
+   usage `branch`.
 
 A client receiving a Welcome including a PreSharedKey of type `resumption` with
 usage `branch` MUST verify that the new group reflects a subgroup branched from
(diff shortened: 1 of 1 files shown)


kkohbrok:
Thanks for the PR and thanks for digging into this. I think what the sentence is meant to say is that the receiver MUST check that the members of the new group are a subset of the members of the old group at that specific epoch.

Since we're using new LeafNodes in the new group and we don't have a fixed notion of identity (at least not within MLS), we have to refer to the application/the AS to ensure that the identity expressed by the credential in each LeafNode in the subgroup has a corresponding identity (i.e. one that represents the same party) expressed by a credential in a LeafNode in the original group.

Maybe something like:
```
The members of the new group MUST be a subset of the members of the old group. When comparing identifiers across the old and the new group, the identifiers of a member in the new group MUST be such that they would represent a valid set of identifiers for the corresponding member in the old group as described in Section 8.3.    
```
This leaves the possibility of a collision, where the identifiers in a given leaf in the new group would be valid updates to more than one member in the old group. This should probably be forbidden in general, although outside of this case it's not really a problem, because updates are always specific to one leaf. Alternatively, we could somehow include LeafNodeRefs that specifically indicate which member represents which other member. Or just leave it up to the application to sort this all out.

tomleavy:
Curious what you think about saying the the credential has to be equal? We can determine equality of the credential itself just by doing tls serialization.

kkohbrok:
Sure, requiring the credential to be exactly the same would be a simple way to figure out if it's the same party in both groups. However, I can imagine that some applications will want distinct credentials for each group. If we mandate that credentials are the same, this would force the application to use the same credential in every key package/LeafNode. Since subgroups can potentially be branched at any time and in any group, all key packages that a party publishes then have to have the same credential as all of the groups that party is in, so another party branching a subgroup can get hold of a KeyPackage with the same credential as the group the subgroup is branched off of.

Personally, I'd prefer asking the application to check that the identities of the new group are a subset of the identities of the original group.

tomleavy:
I can see that being an issue, although I am somewhat worried about us not being opinionated in same way about how to determine equality here, mainly because of interop type things, but we punt on how to validate the credentials as well in general so I guess this is in line with prior decisions

bifurcation (approved):

tomleavy:
Hey @kkohbrok and @bifurcation check the change I just posted. @psyoptix and I put that together, it's a bit tough to get a generic definition of equivalency.

draft-ietf-mls-protocol.md
```diff
@@ -3128,18 +3128,26 @@ considered invalid.
...
 A client receiving a Welcome including a PreSharedKey of type `resumption` with
 usage `branch` MUST verify that the new group reflects a subgroup branched from
 the referenced group.
 
 * The `version` and `ciphersuite` values in the Welcome MUST be the same as
   those used by the old group.
-* Each LeafNode in the new group's tree MUST be a leaf in the
-  old group's tree at the epoch indicated in the PreSharedKey.
+* Each LeafNode in a new subgroup MUST match some LeafNode in the original
+  group. In this context, a pair of LeafNodes are said to "match" if the
```
kkohbrok, line 3148:
```suggestion
  group. In this context, a pair of LeafNodes is said to "match" if the
```

kkohbrok (approved):
Well put!

PR #625, opened 2022-03-28
More aasvg usage

bifurcation:
Continuing to convert more figures to SVG.

The remaining `ascii-art` figures are that way for two reasons:

* Trees with blank nodes `_` render poorly with `aasvg`.  Replacements for the `_` character, like `-`, don't do much better.
* The diagram showing the reuse guard application benefits from the byte-by-byte markers

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1471,7 +1471,8 @@ struct {
     opaque signature<V>;
     select (MLSMessageContent.content_type) {
         case commit:
-            // MAC(confirmation_key, GroupContext.confirmed_transcript_hash)
+            // MAC(confirmation_key,
+            //     GroupContext.confirmed_transcript_hash)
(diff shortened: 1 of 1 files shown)


rohan-wire:
I think these look good. I didn't modify the trees for the reasons you mentioned. In general I have two concerns to keep an eye out for:
- Make sure the overall width of aasvg ASCII version does not exceed 72 chars
- ASCII version of the art shouldn't look substantially worse. Using . and + in some cases looks a bit worse but is still understandable

Is the Key Schedule the oblong blob or is it derived from the blob?  My read of that section was that it is derived from the blob.

bifurcation:
* All the `ascii-art` figures fit within 72 columns.  I trimmed some `tls` and `python` ones to fit
* I think the ASCII still looks OK with these.  Guess people can opine in WGLC if they don't like it :)
* Key Schedule is the blob (IMO)

PR #628, opened 2022-04-01
Strawman proposal to include BasicCredential in any Credential type

rohan-wire:
I was thinking about a possible proliferation of credentials. This could mean a lot more KeyPackages and it could lead to situations where some members are added to a group and use some fancy or rather esoteric credential scheme and then want to invite other new clients, but not all the new clients support that credential scheme.

This strikes me as potentially unnecessary. If all clients support BasicCredential, why not include a BasicCredential in every other Credential (it would be a copy of the identity and public key from your X.509Certificate or ProofCredential or JsonWebCredential or CborWebTokenCredential or whatever).

Maybe this makes it a lot easier to add new credential types later as in the ProofCredential discussion [1] because the new credential does not prevent communication with other other clients.

[1] https://github.com/mlswg/mls-protocol/pull/592

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1226,17 +1226,18 @@ used with:
   by the CipherSuite of the group
 * One or more identifiers of the holder of the private key
 
-Note that a Credential can provide multiple identifiers for the client.  If an
-application wishes to decided whether a credential represents the correct
-identifier for a participant in a given context, it is up to the application to
-decide what the correct value is and compare it to the credential.  For example,
-a certificate in an X509Credential may attest to several domain names or email
-addresses in its subjectAltName extension.  An application may decide to
-present all of these to a user, or if it knows a "desired" domain name or email
-address, it can check that the desired identifier is among those attested.
-Using the terminology from {{?RFC6125}}, a Credential provides "presented
-identifiers", and it is up to the application to supply a "reference identifier"
-for the authenticated client, if any.
+These required components are included in a BasicCredential structure regardless
+of credential type. Note that a Credential can provide multiple identifiers for
+the client.  If an application wishes to decide whether a credential represents
+the correct identifier for a participant in a given context, it is up to the
+application to decide what the correct value is and compare it to the
+credential.  For example, a certificate in an X509Credential may attest to
+several domain names or email addresses in its subjectAltName extension.
+An application may decide to present all of these to a user, or if it knows a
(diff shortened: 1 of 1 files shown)


raphaelrobert:
I like the general idea. Right now the spec effectively only allows X.509 certs besides BasicCredentials and that might be limiting (which is probably what motivated #592). Maybe the two ideas could be combined? Every KeyPackage would have a BasicCredential that can be used for the MLS-internal verification of signatures and additionally provide proof data that the AS can parse and validate. An X.509 certificate chain would just be one instance of such proof data. I think this would make for a nice separation between what MLS needs for its security guarantees/the stuff the AS needs to agree on.

raphaelrobert:
I agree with your concerns regarding duplication. The upside is however a clear split between what's relevant for the protocol (just the stuff in the BasicCredential) and the stuff for the AS. This would make for a much more modular composition of a generic protocol implementation like mlspp or OpenMLS and a more application-specific AS.

kkohbrok:
I also don't have a strong opinion on this, but we might as well simplify it even further and remove credential types altogether. MLS doesn't care about what's in `authentication_info`, so why even communicate the type? The application might want to know, but we could leave that up to the AS in the same way as we do with the content type in application messages.

This would leave us with the same drawbacks that @bifurcation listed, but gain us some more simplicity.

bifurcation:
Strong disagree on removing credential types @kkohbrok.  This protocol needs to be clear on what's in the `authentication_info`, even if that information is ultimately only processed by the application.

On thinking on this a bit further: I think the concerns about "Fail slow / fail open" are adequately addressed by #631.  And it's not clear to me that any format besides X.509 will end up with the duplication that causes the other issues.  (Though there will be some, since the `auth_info` will have to have some binding to the key.)

So I would be OK with this PR if we refactor to the "public key + auth_info" framework I sketched.

bifurcation:
@rohan-wire Note that I filed #637 as an alternative approach here.  But as discusssed there, I think any of these approaches where we factor the public key out of the credential risks invites bad design patterns in future credentials.

So I'm inclined to burn down this strawman, and close both this PR and #637.

rohan-wire:
> I think any of these approaches where we factor the public key out of the credential risks invites bad design patterns in future credentials.

I don't think that is happening at all here. We are putting the public key in a consistent place. The individual credential types can reference it or duplicate it as required by their syntax, but the public key is always there.

bifurcation:
Closing in favor of #637

PR #631, opened 2022-04-01
Specify compatibility behavior around new credentials.

Bren2010:
This PR adds a way for the members of a group to indicate the credential types they support, and how to negotiate which credential types may be used in a group.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -4236,7 +4236,12 @@ extensible.  As discussed in {{commit}}, Proposals with a non-default proposal
 type MUST NOT be included in a commit unless the proposal type is supported by
 all the members of the group that will process the Commit.
 
-## Credentials
+## Credential Extensibility
+
+In order to ensure that MLS provides meaningful authentication it is important
+that each member is able to authenticate some identity information for each
+other member.  Identity information is encoded in Credentials, so this property
+is assured by ensuring that members use compatible credential types.
 
 The types of credential that may be used in a group is restricted to what all
 members of the group support, as specified by the `capabilities` field of each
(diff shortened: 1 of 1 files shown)


bifurcation (approved):
This seems like a good idea, thanks @Bren2010!

rohan-wire (changes requested):
> The types of credential that may be used in a group is restricted to what all
> members of the group support...

I think this is too restrictive if we adopt PR #628 . I think it is between the client and its AS if all the clients need to support every credential type used in the group.

If you agree with the concept, I am happy to propose slight changes.

raphaelrobert (approved):
I think this makes sense.

draft-ietf-mls-protocol.md
```diff
@@ -4226,6 +4236,21 @@ extensible.  As discussed in {{commit}}, Proposals with a non-default proposal
 type MUST NOT be included in a commit unless the proposal type is supported by
 all the members of the group that will process the Commit.
 
+## Credentials
+
```
bifurcation, line 4240:
```suggestion
## Credential Extensibility

In order to ensure that MLS provides meaningful authentication it is important
that each member is able to authenticate some identity information for each
other member.  Identity information is encoded in Credentials, so this property
is assured by ensuring that members use compatible credential types.

```

draft-ietf-mls-protocol.md
```diff
@@ -4226,6 +4236,21 @@ extensible.  As discussed in {{commit}}, Proposals with a non-default proposal
...
+by choosing an unallocated identifier from the registry in
+{{mls-credential-types}} and indicating support for the credential type in
+published LeafNodes, whether in Update proposals to existing groups or
+KeyPackages that are added to new groups. Once all members in a group indicate
+support for the credential type, members can start using LeafNodes with the new
+credential. Application may enforce that certain credential types always remain
+supported by adding a `required_capabilities` extension to the group's
+GroupContext, which would prevent any member from being added to the group that
+doesn't support them.
+
```
bifurcation, line 4253:
```suggestion

In future extensions to MLS, it may be useful to allow a member to present more
than one credential.  For example, such credentials might present different
attributes attested by different authorities.  To be consistent with the general
principle stated at the beginning of this section, such an extension would need
to ensure that each member can authenticate some identity for each other member.
For each pair of members (Alice, Bob), Alice would need to present at least one
credential of a type that Bob supports.
```

bifurcation (commented):
I discussed this offline with @rohan-wire and we arrived at:

* We might need multiple credentials later (as I suspect)
* If we do that (say in an extension), we will need to update this text
* We should capture the general theory that covers both single- and multiple-credential cases

The proposed additions here just abstract the principle of universal compatibility (which we need in a single-credential world) to one of "at least something" compatibility (which is more compatible with a multi-credential world).

PR #635, opened 2022-04-19
Fix parent hash

mulmarta:
The current draft implements the original incorrect version of the tree-hash-for-parent-hash mechanism proposed in #527. That version [breaks in some border cases](https://github.com/mlswg/mls-protocol/pull/527#issuecomment-1020387612). This PR implements the corrected "full-extension + tree hash" mechanism proposed by @TWal [here](https://github.com/mlswg/mls-protocol/pull/527#issuecomment-1035122916).

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2230,65 +2230,49 @@ that the path from P to the root may contain some blank nodes that are not
 fixed by P's Parent Hash. However, for each node that has an HPKE key, this key
 is fixed by P's Parent Hash.
 
-Finally, `original_sibling_tree_hash` is the original tree hash of S. The
-original tree hash corresponds to the tree hash of S the last time P was
-updated. It can be computed as the tree hash of S in the ratchet tree modified
-the following way:
+Finally, `original_sibling_tree_hash` is the tree hash of S in the ratchet tree
+modified as follows:
 
-* reset the leaves in P.unmerged_leaves to blanks
-* remove P.unmerged_leaves from all unmerged_leaves lists
-* truncate the ratchet tree as described in {{remove}}
+* Extend the subtree of S until it is full (see {{adding-and-removing-leaves}}).
+* For each leaf L in `P.unmerged_leaves`, blank L and remove it from the
+`unmerged_leaves` sets of all parent nodes.
+
+Observe that `original_sibling_tree_hash` does not change between updates of P.
+This property is crucial for the correctness of the protocol.
 
 For example, in the following tree:
 
 ~~~ ascii-art
-              W [D, H]
+              W [F]
         ______|_____
        /             \
-      U [D]           Y [F, H]
+      U               Y [F]
     __|__           __|__
    /     \         /     \
-  T       _       X [F]   _
- / \     / \     / \     / \
-A   B   C   D   E   F   G   H
+  T       _       _      |
+ / \     / \     / \     |
+A   B   C   D   E   F    G
 ~~~
-{: #parent-hash-tree title="A full tree unmerged leaves that illustrate parent
-hash computations" }
+{: #parent-hash-tree title="A tree illustrating parent hash computations." }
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2230,65 +2230,49 @@ that the path from P to the root may contain some blank nodes that are not
...
         ______|_____
        /             \
-      U [D]           Y [F, H]
+      U               Y [F]
     __|__           __|__
    /     \         /     \
-  T       _       X [F]   _
- / \     / \     / \     / \
-A   B   C   D   E   F   G   H
+  T       _       _      |
```
bifurcation, line 2252:
```suggestion
  T       _       X[F]   |
```

I think this is what you mean if you want to get the result below?

draft-ietf-mls-protocol.md
```diff
@@ -2230,65 +2230,49 @@ that the path from P to the root may contain some blank nodes that are not
...
-original tree hash corresponds to the tree hash of S the last time P was
-updated. It can be computed as the tree hash of S in the ratchet tree modified
-the following way:
+Finally, `original_sibling_tree_hash` is the tree hash of S in the ratchet tree
+modified as follows:
 
-* reset the leaves in P.unmerged_leaves to blanks
-* remove P.unmerged_leaves from all unmerged_leaves lists
-* truncate the ratchet tree as described in {{remove}}
+* Extend the subtree of S until it is full (see {{adding-and-removing-leaves}}).
```
bifurcation, line 2236: Suggest "Extend the subtree of S by adding blank leaves".  Also, since we don't talk much about "full", a reminder might be helpful here, say "(i.e., until its size is a power of 2)".

draft-ietf-mls-protocol.md
```diff
@@ -4946,16 +4922,10 @@ To construct the tree in {{evolution-tree}}:
...
-* E sends an empty Commit, setting X, Y, and W
-* A adds a new member at F in a partial Commit, adding F as unmerged at X, Y,
-  and W
-* A removes D, resetting T, U, and W (in particular, F is no longer unmerged at
-  W)
-* A adds new memebers at D and H in a partial commit
-  * D is added as unmerged at U, W
-  * H is added as unmerged at Y, W
+* E sends an empty Commit, setting Y and W
+* A adds a new member at F in a partial Commit, adding F as unmerged at Y and W
```
bifurcation, line 4928:
```suggestion
* A adds a new member at F in a partial Commit, adding F as unmerged at X, Y, and W
```

bifurcation (approved):
Thanks for finding yet more parent-hash issues @mulmarta :) 

I was initially concerned about this approach, because it seemed like it would be expensive to fill out the tree.  However, (a) we only incur that cost for the `log(N)` nodes on the right edge of the tree, and (b) the extra nodes to the right are "virtual", in the sense that they don't take up any memory, they just cause some extra hashes.  I [prototyped this in MLSpp](https://github.com/cisco/mlspp/commit/dc35e790c5d6b22cb3307ed6973084d1315db741) and it was pretty straightforward (that also allowed me to verify that it fixes the pathological case that @mulmarta notes).

Overall, this looks good to merge once we get the noted nits fixed.

draft-ietf-mls-protocol.md
mulmarta: Actually, it's the result below that's wrong. This tree couldn't result from any execution. Reason: after the "omit redundant nodes" PR, X would never be inserted by E's commit if F was blank at the time. I changed the result.

draft-ietf-mls-protocol.md
mulmarta: Done. Hope it's better now.

draft-ietf-mls-protocol.md
mulmarta: I removed X from the tree, so the current version is correct again.

PR #637, opened 2022-04-28
Ensure that a signature public key is always intelligible

bifurcation:
This PR moves the signature key from `Credential` to `LeafNode`, to ensure that an MLS client always has a signature public key with which to verify messages, even if it does not support a member's credential type.

This is an alternative approach to #628, and IMO a bit cleaner.  We no longer need the notion of a SignatureScheme, since BasicCredential doesn't stand alone.  And since Credential only appears in LeafNode, common fields can be hoisted up into LeafNode. The Credential struct really only exists as a conceptual boundary, to make the exposition clearer; otherwise you could just shove the `select` into LeafNode.

Note that this does not obviate the need for #631.  This PR only ensures that the signatures can be processed; that PR helps ensure that the signatures are meaningful.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1173,6 +1173,19 @@ the same as the signature algorithm specified in the credentials in the leaves
 of the tree (including the leaf node information in KeyPackages used to add new
 members).
 
+Like HPKE public keys, signature public keys are represented as opaque values in
+a format defined by the cipher suite's signature scheme.
+
+~~~ tls
+opaque SignaturePublicKey<V>;
+~~~
+
+For ciphersuites using Ed25519 or Ed448 signature schemes, the public key is in
+the format specified in {{?RFC8032}}.  For ciphersuites using ECDSA with the
+NIST curves (P-256, P-384, or P-521), the public key is the output of the
+uncompressed Elliptic-Curve-Point-to-Octet-String conversion according to
+{{SECG}}.
+
 To disambiguate different signatures used in MLS, each signed value is prefixed
 by a label as shown below:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1230,19 +1230,29 @@ ciphersuite.
...
+~~~
+
+For ciphersuites using Ed25519 or Ed448 signature schemes, the public key is in
+the format specified in {{?RFC8032}}.  For ciphersuites using ECDSA with the
+NIST curves (P-256, P-384, or P-521), the public key is the output of the
+uncompressed Elliptic-Curve-Point-to-Octet-String conversion according to
+{{SECG}}.
+
+A Credential can provide multiple identifiers for the client.  It is up to the
+applciation to decide which identifier or identifiers to use at the application
```
kkohbrok, line 1254:
```suggestion
application to decide which identifier or identifiers to use at the application
```

kkohbrok (changes requested):
Can you elaborate on why we need this if we also mandate that all clients have to support the credential types of all other clients? I would think that if a client supports a credential type, it knows how to extract the signature public key from that credential.

bifurcation:
@kkohbrok - As I noted in the intro comment, I think you can view these two PRs as addressing different levels of interop -- the mechanics of verifying signatures vs. the ability to actually meaningfully interpret a credential.  Always having the key in the same place could also allow some minor simplification to implementation logic

It's also interesting to consider what this would imply for future credential types.  On the one hand, they wouldn't need to recite the public key, just provide a binding to it, like a hash or a signature where the public key was filled in from outside.  On the other hand, it would open up the door to credential types that aren't actually bound to the public key, which creates replay risk.

In any case, as the above might hint, I don't have strong feelings about this.  I mainly wanted to present an alternative to #628 that was more palatable to me.  Given the replay risk noted above, though, I think I'm trending toward closing both this PR and #628, and just keeping the current structure with the improved compatibilty bounds provided by #631.

rohan-wire (approved):

bifurcation:
Latest commit responds to some offline comments from @rohan-wire: 

* Having `public_key` and `signature_key` makes it a little confusing what `public_key` is about.  So changed to `encryption_key` and propagated that change throughout.
* The definition of `SignaturePublicKey` fits better in the ciphersuites section (next to `HPKEPublicKey`) than it does in the credentials section

Having done that, I'm more positive on this change.  The `encryption_key` change is a good clarification, and having the signature public key separate from the credential provides a stonger decoupling of the credential from the MLS crypto. The only thing the credential layer needs to know now is how to bind to the signature key.

PR #639, opened 2022-04-29
Clarify secret/key; change to epoch authenticator

### LINKED ISSUE #638: Keys and Secrets
Reopening #630 due to incomplete discussion and inaccuracy.

The claims presented are incorrect in response by @bifurcation demonstrate a misunderstanding of the roles of these keys and uses, and #636 constitutes a change to the fundamental uses thereof.

The authentication_secret is NOT a public value nor was it ever introduced as such. If an application is using it as a public value, that onus is on the developers, and I advise against using Cisco's implementation as a standard for MLS. As shown in the name and all original discussion, the authentication_secret is intended as a secret value (or else 'public' in the same way that the group secret is 'public). It is used for further derivation to support authentication.

This should be corrected, and such foundational changes without WG discussion should be avoided.

bifurcation:
Fixes #638

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2781,11 +2781,11 @@ by the other group members.
 This property can be used to construct defenses against impersonation attacks
 that are effective even if members' signature keys are compromised. As a trivial
 example, if the users of the clients in an MLS group were to meet in person and
-confirm that their epoch authenticator values were equal (using some suitable
-user interface), then each user would be assured that the others were not being
-impersonated in the current epoch. As soon as the epoch changed, though, they
-would need to re-do this confirmation. The state of the group would have
-changed, possibly introducing an attacker.
+reliably confirm that their epoch authenticator values were equal (using some
+suitable user interface), then each user would be assured that the others were
+not being impersonated in the current epoch. As soon as the epoch changed,
+though, they would need to re-do this confirmation. The state of the group would
+have changed, possibly introducing an attacker.
 
 More generally, in order for the members of an MLS group to obtain concrete
 authentication protections using the `epoch_authenticator`, they will need to
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -463,6 +463,12 @@ Application Message:
 Terminology specific to tree computations is described in
 {{ratchet-tree-terminology}}.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2769,14 +2771,27 @@ Some uses of resumption PSKs might call for the use of PSKs from historical
...
+The main MLS key schedule provides a per-epoch `epoch_authenticator`. If one
+member of the group is being impersonated by an active attacker, the
+`epoch_authenticator` computed by their client will differ from those computed
+by the other group members.
+
+This property can be used to construct defenses against impersonation attacks
+that are effective even if members' signature keys are compromised. As a
+trivial example, if the users of the clients in an MLS group were to meet in
+person and confirm that their epoch authenticator values were equal, then each
+user would be assured that the others were not being impersonated in the current
```
br-hale, line 2785: A slight suggested rewording: "...in an MLS group were to reliably confirm out-of-band that their epoch authenticator values were equal, then..."
[this is just because actually showing the epoch authenticators face-to-face depends on other device/display attributes as well, which we do not want to make claims on here]

draft-ietf-mls-protocol.md
```diff
@@ -2769,14 +2771,27 @@ Some uses of resumption PSKs might call for the use of PSKs from historical
...
+epoch. As soon as the epoch changed, though, they would need to re-do this
+confirmation. The state of the group would have changed, possibly
+introducing an attacker.
+
+More generally, in order for the members of an MLS group to obtain concrete
+authentication protections using the `epoch_authenticator`, they will need to
+use it in some secondary protocol (such as the face-to-face protocol above).
```
br-hale, line 2792: Corresponding to the above edit:
"...secondary protocol (such as a face-to-face protocol)..."

br-hale:
That looks like a good change and clarification. I added a couple small changes in in-line comments above.

draft-ietf-mls-protocol.md
bifurcation: Added a hand-wave "(using some suitable user interface)".  I would like to keep the in-person aspect, since it makes clear that this is really just a toy example, and if you did something real you would need to do real analysis.

bifurcation:
Thanks for the review @br-hale, and thanks for hashing through this.

br-hale:
The latest edit looks good and that works.

PR #654, opened 2022-05-05
Remove AppAck proposal.

Bren2010:


rohan-wire (commented):
One comment that renumbering the group_context_extensions proposal was probably a bad idea from a process perspective which we should avoid doing in the future. In this case it seems unlikely that many people have implemented it who could not make a breaking change.

PR #656, opened 2022-05-06
Editorial

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -566,18 +566,22 @@ ReadVarint(data):
...
   // remaining bytes.
   v = v & 0x3f
   repeat length-1 times:
     v = (v << 8) + data.next_byte()
-  return v
 
   // Check that the encoder used the minimum bits required
-  if length > 1 && v < (1 << (length - 1)):
-    raise an exception
+  if length > 1 && v < (1 << ((8*(length-1))-2)):
```
TWal, line 581: The fix is still buggy, here is the correct one:
```suggestion
  if prefix >= 1 && v < (1 << (8*(1 << (prefix-1))-2)):
```

Bren2010: What's the bug in the current code?

TWal: In your code, all numbers between `2^14` and `2^22` will be rejected: they are represented with `prefix = 2` (hence `length = 4`). In this case, the buggy check is `v < (1 << (8*3-2))`, the fixed check is `v < (1 << (8*2-2))`.

rohan-wire:
Some of these editorial comments I find less clear than the original. The code changes and varint are fine. Could you please split this into two more atomic PRs so we can discuss separately?

Bren2010:
Feel free to leave comments on the editorial portion of this PR to keep things moving

draft-ietf-mls-protocol.md
Bren2010: Ah, thank you!

draft-ietf-mls-protocol.md
```diff
@@ -683,46 +687,31 @@ The cryptographic state at the core of MLS is divided into three areas of respon
   subsets of the group.  Each epoch has a distinct ratchet tree. It seeds the
   _key schedule_.
 * A _key schedule_ that describes the chain of key derivations used to progress from
-  epoch to epoch (mainly using the _init_secret_ and _epoch_secret_); and to derive
+  epoch to epoch (mainly using the _init_secret_ and _epoch_secret_) and which derives
```
bifurcation, line 690:
```suggestion
  epoch to epoch (mainly using the _init_secret_ and _epoch_secret_), as well as the derivation of
```

draft-ietf-mls-protocol.md
```diff
@@ -683,46 +687,31 @@ The cryptographic state at the core of MLS is divided into three areas of respon
...
 
 Each new epoch is initiated with a Commit message.  The Commit instructs
 existing members of the group to update their views of the ratchet tree by applying
 a set of Proposals, and uses the updated ratchet tree to distribute fresh
 entropy to the group.  This fresh entropy is provided only to members in the new
-epoch, not to members who have been removed, so it maintains the confidentiality
-of the epoch secret (in other words, it provides post-compromise security with
+epoch and not to members who have been removed, which maintains the confidentiality
+of the epoch secret (in other words, the protocol provides post-compromise security with
 respect to those members).
```
bifurcation, line 707:
```suggestion
epoch and not to members who have been removed. Commits thus maintains the property that the
the epoch secret is confidential to the members in the current epoch.
```

draft-ietf-mls-protocol.md
```diff
@@ -683,46 +687,31 @@ The cryptographic state at the core of MLS is divided into three areas of respon
...
-  epoch.
-
-* A _confirmation key_ that is used to confirm that all members agree on the
-  shared state of the group.
-
-* A _resumption secret_ that members can use to prove their membership in the
-  group, e.g., in the case of branching a subgroup.
-
-Finally, an _init secret_ is derived that is used to initialize the next epoch.
-
```
bifurcation, line 725: These seem like useful examples?  Since they show how the key schedule actually gets put to use.  Might drop `init_secret` and replace `confirmation_key` with `exporter_secret` to make it more functional.

bifurcation (approved):
Couple of suggestions, generally looks good.

draft-ietf-mls-protocol.md
Bren2010: Sure, but it doesn't seem to fit in this section. The rest of the text is talking about messages and how they evolve the group state.

bifurcation: Maybe move them up under the `_key schedule_` bullet above?  "... derive a variety of secrets, for example: <sub-bullets>"

PR #659, opened 2022-05-12
Remove new_member Add proposals.

Bren2010:
Sending an Add proposal from outside the group is how we let people do "external add" functionality without needing to publish the group state. But you can't talk right away like you can with an external Commit. In that case, applications seem more inclined to build this functionality themselves so let's just remove it from the spec.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1502,8 +1502,20 @@ MLSMessageContent is authenticated using the MLSMessageAuth structure.
 
 ~~~ tls
 struct {
-    opaque mac_value<V>;
-} MAC;
+    ProtocolVersion version = mls10;
+    WireFormat wire_format;
+    MLSMessageContent content;
+    select (MLSMessageContentTBS.content.sender.sender_type) {
+        case member:
+        case new_member:
+            GroupContext context;
+
+        case external:
+            struct{};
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1500,49 +1512,28 @@ struct {
...
-            struct{};
-    }
-} MLSMessageContentTBS;
-~~~
+* `external`: The signature key at the index
+  indicated by `sender_index` in the `external_senders` group context
+  extension (see Section {{external-senders-extension}}). The
+  `content_type` of the message MUST be `proposal`.
+* `new_member`: The signature key in the LeafNode in
+    the Commit's path (see Section {{external-initialization}}). The
```
bifurcation, line 1528:
```suggestion
    the Commit's path (see Section {{joining-via-external-commits}}). The
```

bifurcation (approved):
Looks good, thanks for putting this together!

bifurcation:
Virtual interim 2022-05-19:

* External sender change is unrelated
    * @kkohbrok notes that using only SignaturePublicKey loses auth info
    * @Bren2010 to revert this part of the PR
* Clear to merge once conflicts are cleared and above is reverted

kkohbrok (approved):

PR #660, opened 2022-05-12
Remove clause about verifying identity changes.

Bren2010:
I don't think the "MUST" clause here is really weight-bearing, especially given that we don't provide any specific way to do the verification. We already say that applications "SHOULD require that a client present the same identifiers" and going against the "SHOULD" means the implementer is taking responsibility for any problems.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1212,6 +1212,10 @@ NIST curves (P-256, P-384, or P-521), the public key is the output of the
 uncompressed Elliptic-Curve-Point-to-Octet-String conversion according to
 {{SECG}}.
 
+The signatures used in this document are encoded as specified in {{!RFC8446}}.
+In particular, ECDSA signatures are DER-encoded and EdDSA signatures are defined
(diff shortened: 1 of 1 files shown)


kkohbrok:
This is again in the gray area of how much authentication do we want to deal with in MLS. I'm not attached to the MUST, but I think we should give people some guidance on what to look out for when deciding on this SHOULD. Maybe this should go in the architecture doc?

raphaelrobert:
There is some overlap with #678. I think it is important to be more specific in the protocol spec about the interaction with the AS.

seanturner:
Park this for now while we evolve #678.

kkohbrok:
After the discussion at the interim, I now lean towards keeping the MUST, but maybe we can phrase it a little differently. Instead of mandating that the application MUST do the verification, we could mandate that the implementation MUST present the credentials to the AS for verification of the credential succession. Then it's really up to the AS what it wants to do, but at least different implementations have a somewhat consistent interface with the AS (which we can discuss in #678).

PR #662, opened 2022-05-14
More wording improvements, updating sections that got out of date.

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -1637,7 +1632,7 @@ schedule it is; if this persistent state is lost or corrupted, a client might
 reuse a generation that has already been used, causing reuse of a key/nonce pair.
 
 To avoid this situation, the sender of a message MUST generate a fresh random
-4-byte "reuse guard" value and XOR it with the first four bytes of the nonce
+four byte "reuse guard" value and XOR it with the first four bytes of the nonce
```
bifurcation, line 1635:
```suggestion
four-byte "reuse guard" value and XOR it with the first four bytes of the nonce
```

draft-ietf-mls-protocol.md
```diff
@@ -1935,9 +1935,8 @@ The client verifies the validity of a LeafNode using the following steps:
     leaf node was proposed for addition, even if it is expired at these later
     points in the protocol.
 
-* Verify that the `leaf_node_source` field has the appropriate value for the
-  context in which the LeafNode is being validated (as defined in
-  {{leaf-node-contents}}).
+* If applicable, verify the `parent_hash` as specified in
+  {{verifying-parent-hashes}}.
```
bifurcation, line 1939: This doesn't make sense.  It's not meaningful to just verify the parent hash in a leaf.  For example, a leaf's parent hash might be invalid just because its sibling updated more recently.  Parent hashes should really only be validated (a) on validating a newly received tree, and (b) when validating a Commit UpdatePath.

draft-ietf-mls-protocol.md
```diff
@@ -2125,11 +2124,11 @@ the following information for each node in the filtered direct path of the
...
 
 * The public key for the node
-* Zero or more encrypted copies of the path secret corresponding to
+* One or more encrypted copies of the path secret corresponding to
   the node
 
-The path secret value for a given node is encrypted for the subtree
-corresponding to the parent's non-updated child, that is, the child
+The path secret value for a given node is encrypted to the subtree
+rooted at the parent's non-updated child. That is, the child
```
bifurcation, line 2131:
```suggestion
rooted at the parent's non-updated child, that is, the child
```

If there's a full stop here, the "That is..." part isn't a full sentence.  It might read more smoothly as "i.e."?

draft-ietf-mls-protocol.md
```diff
@@ -2157,10 +2156,10 @@ The recipient of an UpdatePath processes it with the following steps:
        each hash incorporates all the non-blank nodes above it. The root node
        always has a zero-length hash for this value.
    * For nodes where a path secret was recovered in step 1 ("Compute the
-     updated path secrets"), compute and store the node's updated private key.
+     updated path secrets"), store the node's updated private key.
```
bifurcation, line 2159: The private key needs to computed from the path secret at some point.  Do you just mean for that to be implicit?

draft-ietf-mls-protocol.md
```diff
@@ -2366,13 +2365,11 @@ chain from the committer's leaf to the root.
...
-The path contains a public key and encrypted secret value for all
-intermediate nodes in the filtered direct path from the leaf to the
-root. The path is ordered
-from the closest node to the leaf to the root; each node MUST be the
-parent of its predecessor.
+As described in {{commit}}, each Commit message may optionally contain an
+UpdatePath, with a new LeafNode and set of parent nodes for the sender's
+filtered direct path. For each intermediate node, the UpdatePath contains a new
+public key and encrypted path secret. The nodes are ordered such that each node
+is the ancestor of its predecessor.
```
bifurcation, line 2372:
```suggestion
is the next non-blank ancestor of its predecessor in the filtered direct path.
```

Bren2010: I don't think this is right, the filtered direct path doesn't have blank nodes.

bifurcation: I can see how it's confusing to say both.  How about either "next non-blank ancestor" or "next ancestor in the filtered direct path"?

Alternatively, the filtered direct path already has an order defined:

> The direct path of a root is the empty list, and of any other node is the concatenation of that node's parent along with the parent's direct path.

> The filtered direct path of a node is obtained from the node's direct path by removing all nodes ... [note this doesn't change the order]

So we could just say that the nodes in the UpdatePath correspond to the nodes in the filtered direct path.

Issue #672, opened 2022-05-16
Minor section reordering


bifurcation:
* “Adding and Removing Leaves” after “Synchronizing Views of the Tree”
* Consider merging “Ratchet Tree Evolution” and “Synchronizing Views of the Tree” and “Update Paths”
* Swap “Joining via External Commits” and “Joining via Welcome Message”
* Restore section break for “Extensions”

bifurcation:
Assuming people are OK with this, we should probably do it at a "pause" in PRs -- where we've merged a round and there are no new ones in flight -- in order to avoid merge conflicts.

bifurcation:
If folks are OK with this, I will just implement it before shipping the new post-LC draft.

rohan-wire:
> * Consider merging “Ratchet Tree Evolution” and “Synchronizing Views of the Tree” and “Update Paths”

I would skip this one. I've now interviewed 4 developers and one product manager who are reading the spec for the first time. I think it is a lot clearer divided up.

PR #675, opened 2022-05-17
More clarity around validating proposals

### LINKED ISSUE #657: Undefined behavior around removed Committer
It's not clear to me from reading the spec what should happen when a Commit includes a Remove proposal for the committer.

mulmarta:
The text in the RFC around validating proposals, choosing which proposals go into a commit and how to validate a received commit could do with a bit more work.

The root issue is that Section 13.2 states “The sender of a Commit MUST include all valid proposals that it has received during the current epoch. Invalid proposals include, for example, proposals with an invalid signature or proposals that are semantically invalid, such as an Add when the sender does not have the application-level permission to add new users.” There are two intertwined issues with this text:
1. The qualifier “MUST”
2. The text is too vague about what a “valid Proposal” is. E.g. it could be interpreted to mean a Proposal is “valid” relative to the current group state: Can it be applied to the group as it stands now? Then it’s “valid”.

Put simply, it doesn’t make sense to only validate proposals individually. Instead one should validate if the full list of proposals can be committed. Indeed, other parts of the RFC enforce validation steps that apply to the full list. E.g.
* A ReInit proposal MUST be the only one in the commit, and other existing proposals SHOULD be preferred (13.2.1).
* Out of multiple proposals affecting the same leaf, the committer MUST choose an arbitrary one (13.2, “If there are multiple (…)”).
* Out of multiple PSK proposals with the same ID, the committer MUST choose an arbitrary one (13.2).

In fact, once extensions come into play there are likely many more similar situations, even ones affecting lists of otherwise benign proposals (e.g. Add/Remove). Suppose a chat app is built on MLS with a custom room moderator roles extension. The app wants MLS to enforce two rules: Only mods can Add/Remove. Rooms should always have at least 1 Mod. Here’s how the RFC’s current text can become a problem: A and B are the only mods in a room. A makes proposal P1 that removes B. Meanwhile, B makes prop P2 that removes A. C gets both proposals. What should C do?
* Option 1: validate & cache both proposals. → Problem: C can no longer issue a commit as it “MUST include all valid proposals”. (Alternatively, it does commit to both as instructed by the RFC which leaves the room without a mod.)
* Option 2: validate & cache the first prop received (say, P1), but reject P2 as validating both means C can’t commit. → Problem: If someone else commits to P2 by ref. C can’t process the commit as it doesn’t “know” P2.

So, to clean all this up a bit more, this PR proposes the following main changes:
1. Add a section "Validating Proposals" that defines a procedure for validating a *list* of proposals. The procedure starts with checking that each individual proposal is valid, as described in the "Proposals" section. It also has a note that applications may add more rules.
2. Clean up a bit the validation rules in the "Proposals" section and remove redundancies. Now for each proposal, we precisely define when it is valid.
3. In the "Commit" section, add a requirement that both a committer and a receiver of a commit MUST verify that the committed proposal list is valid according to section "Validating Proposals".
4. Downgrade the requirement that “The sender of a Commit MUST include all valid proposals that it has received“ to "The sender of a Commit SHOULD include all valid proposals that it has received, as long as this does not make the proposal list invalid."

Note that the validation in section "Proposals" can be done when a commit is received. The final check in "Validating Proposals" is done when a commit is created and when it is received.

This also fixes #657.


draft-ietf-mls-protocol.md
```diff
@@ -4213,6 +4209,61 @@ Regardless of how the client obtains the tree, the client MUST verify that the
...
+procedures, depending on whether the commit is external or not.
+
+For a regular, i.e. not external, commit the list is invalid if it contains any of the
+following:
+
+* An individual proposal that is invalid as specified in {{proposals}}.
+
+* An Update proposal generated by the committer.
+
+* A Rmove proposal that removes the committer.
```
raphaelrobert, line 4225:
```suggestion
* A Remove proposal that removes the committer.
```

seanturner:
Discussed 19 May. Will leave it open for further review from @bifurcation @raphaelrobert review.

bifurcation:
Virtual interim 2022-05-19:

* `SHOULD` include all valid proposals could use explanation of why it's not a `MUST`
   * Proposals can contradict each other
   * Application policy might declare some policy invalid
* Agreement in principle, @raphaelrobert and @bifurcation to review and merge when ready

draft-ietf-mls-protocol.md
```diff
@@ -1898,10 +1898,11 @@ The validity of a LeafNode needs to be verified at a few stages:
...
 * When a LeafNode is received by a group member in an Add, Update, or Commit
   message
-* When a client joining a group receives LeafNode objects for the other members
-  of the group in the group's ratchet tree
+* When a client joining a group receives LeafNode objects for the leaves of
+  of the group's ratchet tree
 
-The client verifies the validity of a LeafNode using the following steps:
+To verify that a LeafNode is valid regardless of its use, the client starts
+with the following steps:
```
bifurcation, line 1905: I'm not clear why this says "regardless of its use".  Maybe you mean "In all these cases...", but I would prefer leaving it unadorned to not imply that there's a restriction to those cases (in case another one comes up).

If you're going to say "starts with..." you need to have more steps that follow.

draft-ietf-mls-protocol.md
```diff
@@ -1910,46 +1911,45 @@ The client verifies the validity of a LeafNode using the following steps:
   another LeafNode, the authentication service MUST additionally validate that
   the set of identities attested by the credential in the new LeafNode is
   acceptable relative to the identities attested by the old credential.
+  E.g. sending an Update proposal updates the sender's old LeafNode to a new
+  one, or joining via external commit updates the joiner's old LeafNode
+  included in the committed Remove proposal to a new one.
```
bifurcation, line 1916:
```suggestion
  For example:
    * An Update proposal updates the sender's old LeafNode to a new one
    * A "resync" external commit removes the joiner's old LeafNode via a Remove proposal and replaces it with a new one
```

draft-ietf-mls-protocol.md
```diff
@@ -1910,46 +1911,45 @@ The client verifies the validity of a LeafNode using the following steps:
...
-* Verify that the `leaf_node_source` field has the appropriate value for the
-  context in which the LeafNode is being validated (as defined in
-  {{leaf-node-contents}}).
-
-* Verify that the following fields in the LeafNode are unique among the
-  members of the group (including any other members added in the same
-  Commit):
-
-    * `encryption_key`
-    * `signature_key`
```
bifurcation, line 1947: Why are you removing this uniqueness requirement?

draft-ietf-mls-protocol.md
```diff
@@ -1910,46 +1911,45 @@ The client verifies the validity of a LeafNode using the following steps:
...
+  * If the LeafNode appears in a message being sent by the client, e.g., a
+    proposal or a commit, then the client MUST verify that the current time is within
+    the range of the `lifetime` field.
+  * If instead the LeafNode appears in a message being received by the client, e.g.,
+    a proposal, a commit, or a ratchet tree of the group the client is joining, it is
+    RECOMMENDED that the client verifies that the current time is within the range
+    of the `lifetime` field.
```
bifurcation, line 1938: This description seems less clear to me than before.  Are you trying to be more generic?

Perhaps we should just have a generic RECOMMENDED here, and have the MUST in the Add section.

draft-ietf-mls-protocol.md
```diff
@@ -3249,32 +3249,42 @@ struct {
 ~~~
 
 On receiving an MLSMessageContent containing a Proposal, a client MUST verify the
-signature inside MLSMessageAuth.  If the signature verifies
-successfully, then the Proposal should be cached in such a way that it can be
-retrieved by hash (as a ProposalOrRef object) in a later Commit message.
+signature inside MLSMessageAuth and that the `epoch` field of the enclosing
```
bifurcation, line 3252: We should delete these requirements here, and make them generic. It seems like you should always check the signature on MLSMessageContentAuth before using the content, regardless of the content type.

draft-ietf-mls-protocol.md
bifurcation: I see you have added it below.  That seems like a lot of repetition and risk of decoherence.  I would prefer to have it up here.

If the concern is the "added in the same Commit" part -- it seems like this requirement could be applied iteratively, in the sense that if you're updating/adding a bunch of leaves in a Commit and you do this check relative to the current tree before you apply an update/add, then that will catch a duplication.

draft-ietf-mls-protocol.md
```diff
@@ -3551,34 +3571,16 @@ a Commit message before sending application data. This ensures, for example,
...
-client or a client already in the group (for example, identical KeyPackages or
-KeyPackages sharing the same Credential), the committer again chooses one to
-include and considers the rest invalid. The committer MUST consider invalid any
-Add or Update proposal if the Credential in the contained KeyPackage shares the
-same signature key with a Credential in any leaf of the group, or if the
-LeafNode in the KeyPackage shares the same `encryption_key` with another LeafNode in
-the group.
+A sender and a receiver of a Commit MUST verify that the committed list of
+proposals is valid as specified in {{validating-proposals}}. A list is invalid if, for example,
+it includes a proposal with an invalid signature, a proposal sent within a different epoch.
```
bifurcation, line 3576: Suggest using a more interesting example here, such as including an Update and Remove for the same member.

draft-ietf-mls-protocol.md
```diff
@@ -3551,34 +3571,16 @@ a Commit message before sending application data. This ensures, for example,
...
+or an Add when the sender does not have the application-level permission to add new users.
 
-The Commit MUST NOT combine proposals sent within different epochs. Due to the
-asynchronous nature of proposals, receivers of a Commit SHOULD NOT enforce that
-all valid proposals sent within the current epoch are referenced by the next
+The sender of a Commit SHOULD include all valid proposals that it has received
+during the current epoch, as long as this does not make the proposal list invalid.
```
bifurcation, line 3580:
```suggestion
The sender of a Commit SHOULD include all proposals that it has received
during the current epoch, that are valid according to the rules for their
proposal types and according to application policy, as long as this results in
a valid proposal list.
```

draft-ietf-mls-protocol.md
```diff
@@ -3551,34 +3571,16 @@ a Commit message before sending application data. This ensures, for example,
...
+
+Due to the asynchronous nature of proposals, receivers of a Commit SHOULD NOT enforce
+that all valid proposals sent within the current epoch are referenced by the next
```
bifurcation, line 3583: FWIW I think "valid proposals" is a fine shorthand even with the above edit.

draft-ietf-mls-protocol.md
```diff
@@ -4213,6 +4209,61 @@ Regardless of how the client obtains the tree, the client MUST verify that the
...
+
+* A proposal with a non-default proposal type that is not supported by some
+  members of the group that will process the Commit (i.e., members being added
+  or removed by the Commit do not need to support the proposal type).
+
+An application may extend the above procedure by additional rules, for example,
+requiring application-level permissions to add members, or rules concerning
+non-default proposal types.
+
+For an external commit, the list is valid if:
```
bifurcation, line 4254: It's slightly confusing that the above is a negative check, and this is a positive check.  But that might be unavoidable.  I might rephrase as something like:

... the list is valid if it contains only the following proposals (not necessarily in this order):

* Exactly one ExternalInit proposal
* At most one Remove proposal...
* Zero or more PSK proposals...
* No other proposals

(The last bullet is a little redundant, but it bears repeating.)

bifurcation: (And maybe we should require order here!)

bifurcation (changes requested):

draft-ietf-mls-protocol.md
mulmarta: Thanks for pointing this out! (We used to have a earlier version with more cases and more steps.)

draft-ietf-mls-protocol.md
mulmarta: Thanks for pointing this out! (We used to have a earlier version with more cases and more steps.)

draft-ietf-mls-protocol.md
mulmarta: On second thought, I think this check belongs here.

Re. "added in the same Commit", I think a fix for this would be to change the "Validating Proposals" procedure to check individual validity at the end, after applying all committed proposals. I.e. the list is invalid if "After applying the proposal list, an individual proposal is invalid as specified in {{proposals}}."

Re. applying the check iteratively, the way I read the current draft, it seems that all checks are applied to the current state. For example, its valid to commit to A removing B and B removing A. But if we applied checks iteratively, it would be invalid.

draft-ietf-mls-protocol.md
```diff
@@ -1910,46 +1910,45 @@ The client verifies the validity of a LeafNode using the following steps:
   another LeafNode, the authentication service MUST additionally validate that
   the set of identities attested by the credential in the new LeafNode is
   acceptable relative to the identities attested by the old credential.
+  For example:
+    * An Update proposal updates the sender's old LeafNode to a new one
+    * A "resync" external commit removes the joiner's old LeafNode via a Remove proposal and replaces it with a new one
```
Bren2010, line 1915: Also checking if identities are the same in a resync is a good point, but isn't it undermined by doing the external join and remove separately?

draft-ietf-mls-protocol.md
```diff
@@ -3249,32 +3248,42 @@ struct {
...
+* The KeyPackage is invalid according to {{keypackage-validation}}.
+
+* The LeafNode in the KeyPackage shares the same `signature_key` with another
+  LeafNode in the group.
+
+* The LeafNode in the KeyPackage shares the same `encryption_key` with another
+  LeafNode in the group.
+
+* The Credential in the KeyPackage represents a client already in the
+  group according to the application.
```
Bren2010, line 3278: I don't think this was agreed on

Bren2010:
I opened a PR here that resolves the merge conflicts and makes some additional suggested changes: https://github.com/mulmarta/mls-protocol/pull/1

draft-ietf-mls-protocol.md
kkohbrok: I think this is covered in the External Commits section, but mentioning it here or including a reference probably doesn't hurt.

mulmarta: Currently this requirement is in the External Commit section:
>  There MAY be a single Remove proposal, where the LeafNode in the path field MUST meet the same criteria as the LeafNode in an Update for the removed leaf (see Section 13.1.2). In particular, the credential in the LeafNode MUST present a set of identifiers that is acceptable to the application for the removed participant.

The PR moves it here (so it's only in one place).

This PR is editorial, so I'd rather not remove the restriction here but instead discuss this as a separate issue. I've also been wondering why only one Remove is allowed? E.g. I may want to remove a couple of my old devices.

draft-ietf-mls-protocol.md
mulmarta: This is how we interpreted the text
> If there are multiple Add proposals containing KeyPackages that the committer considers to represent the same client or a client already in the group (for example, identical KeyPackages or KeyPackages sharing the same Credential), the committer again chooses one to include and considers the rest invalid.

If this was not what was meant, the part "or a client already in the group" would be redundant. But of course I'll remove it if we were wrong.

draft-ietf-mls-protocol.md
bifurcation: Agree that we should handle the revalidation separately, in #660.

On "why only one" -- The idea is to *only* cover the resync case, where a device is replacing *its own* state.  Other devices can be cleaned up after the new member has joined.

Bren2010:
@mulmarta Please update to resolve merge conflicts.

draft-ietf-mls-protocol.md
```diff
@@ -1940,45 +1940,48 @@ The client verifies the validity of a LeafNode using the following steps:
...
+    of the `lifetime` field.
 
 * Verify that the extensions in the LeafNode are supported by checking that the
   ID for each extension in the `extensions` field is listed in the
   `capabilities.extensions` field of the LeafNode.
 
+* Verify the `leaf_node_source` field:
+  * If the LeafNode appears in a KeyPackage in an Add proposal, or in a downloaded
+    KeyPackage used to create an Add proposal, verify that `leaf_node_source` is
+    set to `key_package`.
```
bifurcation, line 1975:
```suggestion
  * If the LeafNode appears in a KeyPackage, verify that `leaf_node_source` is
    set to `key_package`.
```

draft-ietf-mls-protocol.md
```diff
@@ -3363,13 +3370,11 @@ struct {
 } Update;
 ~~~
 
-A member of the group applies an Update message by taking the following steps:
+An Update proposal is invalid if any of the following is true:
 
-* Validate the LeafNode as specified in {{leaf-node-validation}}.  The
-  `leaf_node_source` field MUST be set to `update`.
+* The LeafNode is invalid for an Update proposal according to {{leaf-node-validation}}.
```
bifurcation, line 3375:
```suggestion
```

draft-ietf-mls-protocol.md
```diff
@@ -3673,6 +3670,9 @@ their associated GroupContexts are used:
 A member of the group creates a Commit message and the corresponding Welcome
 message at the same time, by taking the following steps:
 
+* Verify that the committed list of proposals is valid as specified in
+  {{proposal-list-validation}}.
```
bifurcation, line 3674:
```suggestion
* Verify that the list of proposals to be committed is valid as specified in
  {{proposal-list-validation}}.
```

draft-ietf-mls-protocol.md
```diff
@@ -4214,6 +4205,67 @@ Regardless of how the client obtains the tree, the client MUST verify that the
 root hash of the ratchet tree matches the `tree_hash` of the GroupContext before
 using the tree for MLS operations.
 
+## Proposal-List Validation
```
bifurcation, line 4208:
```suggestion
## Proposal List Validation
```

bifurcation: I would put this above, probably as a subsection of {{proposals}}.  But we can handle that in a section reordering PR; I have added this to #672

bifurcation:
@mulmarta I proactively merged my editorial nits (with review from @Bren2010) in the interest of getting this merged.  If I've screwed anything up, obviously feel free to file a follow up to fix.

PR #682, opened 2022-05-20
Update external senders extension.

### LINKED ISSUE #681: External Senders Extension is incomplete
The External Senders Extension allows clients to agree on the entities that are allowed to send external proposals, as well as the signature public keys required to verify them. It contains a list of credentials that is meant to achieve this goal. However, due to a recent change, Credentials don't necessarily contain signature public keys anymore. @bifurcation suggested to have it contain a signature public key, as well as a corresponding Credential, which allows members to verify the binding between signature public key and the owning entity's identity, which I think would be a good solution.

Bren2010:
Closes #681

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3536,11 +3536,16 @@ will not have the keys necessary to construct an MLSCiphertext object.
 #### External Senders Extension
 
 The `external_senders` extension is a group context extension that contains
-credentials of senders that are permitted to send external proposals to the
-group.
+the credentials and signature keys of senders that are permitted to send
+external proposals to the group.
 
 ~~~~~
-Credential external_senders<V>;
+struct {
+  SignaturePublicKey signature_key;
+  Credential credential;
+} ExternalSender;
+
+ExternalSender external_senders<V>;
 ~~~~~
 
 ## Commit


bifurcation (approved):
Thanks @Bren2010 I like that this captures both the signature and authentication aspects.

@kkohbrok wdyt?

draft-ietf-mls-protocol.md
```diff
@@ -3536,11 +3536,16 @@ will not have the keys necessary to construct an MLSCiphertext object.
...
-credentials of senders that are permitted to send external proposals to the
-group.
+the credentials and signature keys of senders that are permitted to send
+external proposals to the group.
 
 ~~~~~
-Credential external_senders<V>;
+struct {
+  SignaturePublicKey signature_key;
+  Credential credential;
```
tomleavy, line 3545: Should the combo of SignaturePublicKey and Credential just be independently defined as a struct like Identity, and then used in LeafNode as well as in the extension?

kkohbrok: Not sure I'd call it "identity", but I think this is a good idea.

kkohbrok (approved):
Looks good to me! @tomleavy's idea might be a good additional improvement.

rohan-wire (approved):

rohan-wire (approved):

Issue #669, opened 2022-05-16
More pictures


bifurcation:
* Add illustration of filtered direct path
* Add illustration of parent hash validation, including:
    * Leaves with no parent hash
    * Nodes with invalid parent hash
    * Skipping blank nodes
* Add corresponding derivations to Appendix A
* Move figure in Section 10.2 up to Section 10 as an overview

bifurcation:
@Bren2010 Note that I have taken a stab at parts of this in #689 because it came along with some other tree stuff.

Bren2010:
Duplicate #644

Issue #644, opened 2022-05-04
Illustrations for “filtered direct path”


bifurcation:
In #587, we introduced the idea of a "filtered direct path".  It would be helpful to have an illustration of how a filtered direct path differs from a direct path.

PR #685, opened 2022-05-21
Partial work on #669

Bren2010:


Bren2010:
Unfortunately I just realized this overlaps with #644. I'm happy to remove that part of the PR if you want, @bifurcation

bifurcation:
@Bren2010 I assume you mean #689?  Yeah, I think we can make that one responsible for the direct path / parent hash parts of #669, and this can just hit the secret tree part.

bifurcation:
Sorry for the conflict btw.  It ended up being a natural part of reorganizing things.

PR #686, opened 2022-05-21
Recommend that group ids are generated randomly.

### LINKED ISSUE #661: Specify that clients MUST ensure the group id is unique and fixed for each group.
Group ids are signed to bind certain information to specific groups. But I can imagine a case where a user mis-uses the group id to be something like a "room title" which lets an adversary control the group id and introduces potential problems.

- If we can prove that it's fine for a group id to be reused, then should we get rid of group ids?
- If not (as I expect), then group ids need to be guaranteed to be unique [globally | per client | per signing key] and we need to specify that.

Bren2010:
Closes #661

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3172,6 +3172,13 @@ The creator of a group MUST take the following steps to initialize the group:
 
 * Transmit the Welcome message to the other new members
 
+Group IDs SHOULD be constructed in such a way that there's an overwhelmingly low
+probability of honest group creators generating the same group ID, even without
+assistance from the Delivery Service. For example, by making the group ID a
+freshly generated random value of size `KDF.Nh`. The Delivery Service MAY
+attempt to ensure that group IDs are globally unique by rejecting the creation
+of new groups with a previously used ID.
+
 The recipient of a Welcome message processes it as described in
 {{joining-via-welcome-message}}.  If application context informs the recipient that
 the Welcome should reflect the creation of a new group (for example, due to a
@@ -3863,8 +3870,11 @@ struct {
 } GroupInfo;
 ~~~
 
-New members MUST verify the `signature` using the public key taken from the
-credential in the leaf node of the ratchet tree with leaf index `signer`. The
+New members MUST verify that `group_id` is unique among the groups they're
+currently participating in.
+
+New members also MUST verify the `signature` using the public key taken from the
+leaf node of the ratchet tree with leaf index `signer`. The
 signature covers the following structure, comprising all the fields in the
 GroupInfo above `signature`:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -3818,8 +3818,13 @@ struct {
 } GroupInfo;
 ~~~
 
-New members MUST verify the `signature` using the public key taken from the
-credential in the leaf node of the ratchet tree with leaf index `signer`. The
+New members MUST verify that `group_id` is unique among the groups they're
+currently participating in. The Delivery Service MAY attempt to ensure that
```
bifurcation, line 3822: I may be over-thinking this, but: The group ID isn't necessarily exposed to the DS at group creation time, since the act that creates the group is a Welcome message, which encrypts the group ID.  The group ID is visible in MLSPlaintext and MLSCiphertext, though, so the DS can pick it up there.

bifurcation (approved):

draft-ietf-mls-protocol.md
```diff
@@ -3106,7 +3106,7 @@ The creator of a group MUST take the following steps to initialize the group:
 * Initialize a one-member group with the following initial values:
   * Ratchet tree: A tree with a single node, a leaf containing an HPKE public
     key and credential for the creator
-  * Group ID: A value set by the creator
+  * Group ID: A fresh random value of size `AEAD.Nk`
```
rohan-wire, line 3109: There was never a requirement before that the group id had to have a specific internal structure, format, or length. Why does the protocol suddenly care now? It just needs to be unique enough, and you have specified that in the rest of the diff below. 

s/fresh random value of size `AEAD.Nk`/a unique value set by the creator/

rohan-wire (changes requested):

draft-ietf-mls-protocol.md
raphaelrobert: I agree that it would be prohibitive to require a special structure for group IDs. This would potentially make it difficult to use MLS in existing environments, where group IDs already exist.

It's also not clear to me why `AEAD.Nk` is a good reference for entropy. It depends on the ciphersuite and is therefore orthogonal to the requirements of uniqueness in a system. Moreover, a system that supports several cipehrsuites would see a conflict here.

bifurcation: I would probably use `KDF.Nh`, since that's what we use for other unique values (e.g., PSK nonces).

seanturner:
Brendan to revise based on Rohan suggestion.

bifurcation:
Interim 2022-05-26:

* Concerns around making the group_id totally random
    * Malicious group creator can choose it malicious
    * Federated systems get worse randomness
* Using AEAD.Nk depends on ciphersuite, could be awkward 
* @Bren2010  will revise to say:
    * Here are the requirements that group ID must satisfy
    * If you don’t have an alternative plan, do it randomly

bifurcation (approved):

PR #687, opened 2022-05-21
Add a marker byte to tree hash input structs.

### LINKED ISSUE #663: Leaf/parent indicator bit in tree hash?
When computing the tree hash, should there be an indicator bit for whether the value being hashed represents a leaf or parent node?

Bren2010:
Closes #663. I'm not sure if this is actually necessary, but there usually seem to be issues in Merkle trees when the leaves are not explicitly distinct from the parents.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2218,6 +2218,13 @@ objects contain the `ParentNode` (if any) and the tree hash of the node's left
 and right children.
 
 ~~~ tls
+enum {
+    reserved(0),
+    leaf(1),
+    parent(2),
+    (255)
+} NodeType;
+
 struct {
   NodeType node_type;
   select (TreeHashInput.node_type) {
@@ -4113,13 +4120,6 @@ the whole public state of the ratchet tree can be provided in an extension of
 type `ratchet_tree`, containing a `ratchet_tree` object of the following form:
 
 ~~~ tls
-enum {
-    reserved(0),
-    leaf(1),
-    parent(2),
-    (255)
-} NodeType;
-
 struct {
     NodeType node_type;
     select (Node.node_type) {
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2219,12 +2219,10 @@ and right children.
 
 ~~~ tls
 struct {
-  uint8 leaf_or_parent_marker;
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2217,6 +2217,7 @@ might include a `LeafNode` object depending on whether or not it is blank.
 
 ~~~ tls
 struct {
+    uint8 marker = 0;
```
bifurcation, line 2220: Nit: I might call this `leaf_or_parent_marker`.

bifurcation (approved):
I am also not totally clear on the necessity here.  It seems like the issues with Merkle trees are probably about an adversary being able to choose a leaf value that has the same form as a parent.  Assuming that's the case, we probably wouldn't have similar situations here, given the extra structure.  But having a marker saves us having to make that assumption or do the analysis to check it.

Seems low-cost and unlikely to hurt in any case.

TWal:
I would find it nicer to describe things with a `select` like in other parts of the protocol:

```
struct {
  uint8 leaf_or_parent_marker;
  select (leaf_or_parent_marker) {
    case 0:
      LeafNodeHashInput hash_input;
    case 1:
      ParentNodeHashInput hash_input;
  }
} TreeHashInput;
```

It is equivalent to your modification. Stating things that way ensures there is no ambiguity: TreeHashInput is parseable, which is I think the property you want to ensure with this PR.

seanturner:
Use enum instead and then ready for merge.

bifurcation:
Interim 2022-05-26:

* Nicer with an `enum`
* @Bren2010 to update to `enum`, then merge

draft-ietf-mls-protocol.md
```diff
@@ -2208,33 +2208,38 @@ specifically:
...
+The tree hash of an individual node is the hash of the node's `TreeHashInput`
+object, which may contain either a `LeafNodeHashInput` or a
+`ParentNodeHashInput` depending on the type of node. `LeafNodeHashInput` objects
+contain the `leaf_index` and the `LeafNode` (if any). `ParentNodeHashInput`
+objects contain the `ParentNode` (if any) and the tree hash of the node's left
+and right children.
 
 ~~~ tls
+struct {
+  NodeType node_type;
```
bifurcation, line 2222: This is defined all the way down in the ratchet tree extension.  Please move it up to this section.

bifurcation (approved):

PR #688, opened 2022-05-21
Use a hash function for hash identifiers.

### LINKED ISSUE #640: Weak transcript binding of proposals by reference
In #510 we changed from using full-size hash references to using truncated hashes.  This is fine for cases where the reference is only acting as a reference, so unambiguous identification is the only concern.

For proposal references, however, the hash-based reference is also the way that the proposal is incorporated into the transcript.  In this case, we end up with a looser binding with truncated hashes, since a malicious transcript need only match a subset of the bits of the hash.

This doesn’t seem like a huge issue in practice, since if the hash function is broken enough to allow 128-bit collisions, we likely have deeper problems.  But it might mean that transcript integrity fails before other aspects of the protocol.

To fix this, we could either restore all hash-based identifiers to full size (basically, revert #510), or we could revert just the proposal references.  The former would be more uniform; the latter would be more parsimonious.  I’m pretty ambivalent between the two.

Bren2010:
Closes #640.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1243,18 +1243,24 @@ messages refer to KeyPackages for the members being welcomed, and Commits refer
 to Proposals they cover.  These identifiers are computed as follows:
 
 ~~~ tls
-opaque HashReference[16];
+opaque HashReference[Hash.length];
 
 HashReference KeyPackageRef;
 HashReference ProposalRef;
 ~~~
 
 ~~~ pseudocode
-MakeKeyPackageRef(value) = KDF.expand(
-  KDF.extract("", value), "MLS 1.0 KeyPackage Reference", 16)
+MakeKeyPackageRef(value) = RefHash("MLS 1.0 KeyPackage Reference", value)
+MakeProposalRef(value)   = RefHash("MLS 1.0 Proposal Reference", value)
 
-MakeProposalRef(value) = KDF.expand(
-  KDF.extract("", value), "MLS 1.0 Proposal Reference", 16)
+RefHash(label, value) = Hash(RefHashInput)
+
+Where RefHashInput is defined as:
+
+struct {
+  opaque label<V> = label;
+  opaque value<V> = value;
+} RefHashInput;
 ~~~
 
 For a KeyPackageRef, the `value` input is the encoded KeyPackage, and the


draft-ietf-mls-protocol.md
```diff
@@ -1243,18 +1243,24 @@ messages refer to KeyPackages for the members being welcomed, and Commits refer
 to Proposals they cover.  These identifiers are computed as follows:
 
 ~~~ tls
-opaque HashReference[16];
+opaque HashReference[Hash.length];
```
bifurcation, line 1246:
```suggestion
opaque HashReference<V>;
```

Elsewhere, where we have values whose size depends on the ciphersuite, we use vectors instead of arrays.  See, e.g., HPKEPubicKey.  It costs a byte, but it means the parsing/serialization logic doesn't have to know about ciphersuites.

draft-ietf-mls-protocol.md
```diff
@@ -1243,18 +1243,24 @@ messages refer to KeyPackages for the members being welcomed, and Commits refer
...
 
 ~~~ pseudocode
-MakeKeyPackageRef(value) = KDF.expand(
-  KDF.extract("", value), "MLS 1.0 KeyPackage Reference", 16)
+MakeKeyPackageRef(value) = RefHash("MLS 1.0 KeyPackage Reference", value)
+MakeProposalRef(value)   = RefHash("MLS 1.0 Proposal Reference", value)
 
-MakeProposalRef(value) = KDF.expand(
-  KDF.extract("", value), "MLS 1.0 Proposal Reference", 16)
+RefHash(label, value) = Hash(RefHashInput)
```
bifurcation, line 1256: Paging @kkohbrok @TWal et al. for crypto review.  @Bren2010 fwiw, I think the more minimal change would be:

``` suggestion
RefHash(label, value) = KDF.expand(KDF.extract("", value), label, KDF.Nh)
```

(That is, just change `16` to `KDF.Nh`.)

From an implementation POV, I don't think there's much difference either way.  Extract/expand is 4 hash invocations instead of 1, but none of this is done at much scale.

bifurcation (changes requested):

draft-ietf-mls-protocol.md
Bren2010: Using a hash ensures that we're really relying on the ciphersuite's understanding of what a secure hash function is. If KDF.Nh is the size of the output of KDF.Extract (which outputs keys, not collision resistant hashes), then it may again be too small to be collision-secure.

kkohbrok: At least for the key derivation proofs, we're already assuming that KDFs are collision resistant, which I believe is fairly standard. That being said increasing the length is probably a good idea. 

Also, I don't think that it makes much of a difference if we're using a hash function or a KDF, since we don't rely on any pseudorandomness properties of the hash references.

mulmarta: I lean towards using the hash. A hash function is "the" collision resistant primitive and since collision resistance is what we want here, hash would be the right primitive. The main purpose of extract is pseudo-randomness, which is unrelated to collision resistance. (Even if we have to assume collision resistance of extract for some property, using hash here makes it less ambiguous which property is needed.)

bifurcation: Thanks @kkohbrok @mulmarta, clearing my objection.

bifurcation (approved):

Issue #673, opened 2022-05-16
Streamline structs


bifurcation:
* The `MLSMessage*` collection of structs could be less chatty.  Suggest
    * `MLSMessageContent => MessageContent`
    * `MLSMessageContentAuth => AuthenticatedContent`
    * `MLSMessageCommitContent => CommitContent`
    * `MLSMessageCommitAuthData => CommitAuthData`
* PreSharedKeys is only used once, can be inlined in GroupSecrets
* Use GroupContext for the internals of GroupInfo

bifurcation:
Note:
* `PreSharedKeys` was fixed in #691 
* In #695, we did:
    * `MLSMessageCommitContent => ConfirmedTranscriptHashInput`
    * `MLSMessageAuthData => InterimTranscriptHashInput`

So only the following remain:
* `MLSMessageContent => MessageContent`
* `MLSMessageContentAuth => AuthenticatedContent`

PR #691, opened 2022-05-22
Clean up PSK provisions

### LINKED ISSUE #666: Clean up PSK specification
* `psk_nonce` currently described at time of use.  In the "Pre-Shared Keys" section, we should just specify that it must always be a fresh value of length `KDF.Nh`
* The ReInit provisions require `epoch=1`, but those for branching do not.  Presumably these should be the same?
* Currently, the processing rules for PreSharedKey proposal says “MUST be external”.  This seems overly restrictive; for example, it rules out the scenario in Figure 8.  Suggest “MUST NOT be branch/reinit” instead.

bifurcation:
Closes #666 

Also covers the PSK part of #673

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2722,11 +2722,13 @@ struct {
   }
   opaque psk_nonce<V>;
 } PreSharedKeyID;
-
-struct {
-    PreSharedKeyID psks<V>;
-} PreSharedKeys;
 ~~~
+
+Each time a client injects a PSK into a group, the `psk_nonce` of its
+PreSharedKeyID MUST be set to a fresh random value of length `KDF.Nh`, where
+`KDF` is the KDF for the ciphersuite of the group into which the PSK is being
+injected. This ensures that even when a PSK is used multiple times, the value
+used as an input into the key schedule is different each time.
 
 On receiving a Commit with a `PreSharedKey` proposal or a GroupSecrets object
 with the `psks` field set, the receiving Client includes them in the key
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -3366,8 +3363,8 @@ struct {
 } PreSharedKey;
 ~~~
 
-The `psktype` of the pre-shared key MUST be `external` and the `psk_nonce` MUST
-be a randomly sampled nonce of length `KDF.Nh`. When processing a Commit message
+A PreSharedKey proposal MUST NOT contain a PSK of type `resumption` and usage
```
Bren2010, line 3366:
```suggestion
A PreSharedKey proposal MUST NOT contain a PSK of type `resumption` that has usage
```

kkohbrok (approved):

draft-ietf-mls-protocol.md
```diff
@@ -3178,13 +3180,10 @@ The members of a group reinitialize it using the following steps:
     * The `group_id`, `version`, and `cipher_suite` fields in the Welcome
       message MUST be the same as the corresponding fields in the ReInit
       proposal.
-    * The `epoch` in the Welcome message MUST be 1
+    * The `epoch` in the Welcome message MUST be 1.
     * The Welcome MUST specify a PreSharedKey of type `resumption` with usage
       `reinit`.  The `group_id` must match the old group, and the `epoch` must
       indicate the epoch after the Commit covering the ReInit.
```
Bren2010, line 3186: The above line about "epoch must be 1" conflicts with this line that "epoch must be +1 the last Commit"

draft-ietf-mls-protocol.md
```diff
@@ -3178,13 +3180,10 @@ The members of a group reinitialize it using the following steps:
     * The `group_id`, `version`, and `cipher_suite` fields in the Welcome
       message MUST be the same as the corresponding fields in the ReInit
       proposal.
-    * The `epoch` in the Welcome message MUST be 1
+    * The `epoch` in the Welcome message MUST be 1.
```
bifurcation, line 3183:
```suggestion
```

draft-ietf-mls-protocol.md
bifurcation: Good point, reverted.

bifurcation:
Interim 2022-05-26:

* Ready to go once merge conflicts are resolved.

PR #692, opened 2022-05-24
Use new tree for context in path secret encryption

### LINKED ISSUE #648: Clarify order of operations in processing external join
In #511, we removed the requirement for an external Commit to have an Add proposal, under the theory that the required `path` already has a leaf node for the new joiner.  When I went to implement this, however, I noticed some ambiguities. 

First, the Commit processing section doesn't mention when the new joiner is added.  The obvious approach is to do the add as part of the `path` processing, but that processing presumes that the sender is already in the tree (for example, there is no provision for extending the tree).

Second, the UpdatePath for an external commit needs to be done in a specific order in order to avoid a circular dependency.  The TreeKEM encapsulation done by the joiner uses the GroupContext a context parameter, including the hash of the tree being used.  But the LeafNode in the `path.leaf_node` has to have a parent hash in it, which can only be computed after the encapsulation operation.  This circularity can be resolved, but only by preparing the UpdatePath with the following specific steps:

1. Compute the path secrets
2. Compute the parent hash and the new leaf node
3. Compute the new tree hash
4. Encrypt the path secrets to the rest of the group

That rules out some implementations that compute path secrets and encrypt them in one traversal of the direct path (this is what MLSpp does now).

The options here are basically:

1. Re-add the Add proposal, thus having a throw-away temporary LeafNode for the joiner
2. Specify that the UpdatePath must be computed in the specific order above

To me, (2) seems like a pretty clear winner.

bifurcation:
Closes #648 via option (3)

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2427,12 +2427,17 @@ ciphertext = context.Seal("", path_secret)
 ~~~
 
 where `node_public_key` is the public key of the node that the path
-secret is being encrypted for, group_context is the current GroupContext object
+secret is being encrypted for, group_context is the provisional GroupContext object
 for the group, and the functions `SetupBaseS` and
 `Seal` are defined according to {{!RFC9180}}.
 
 Decryption is performed in the corresponding way, using the private
 key of the resolution node.
+
+~~~ pseudocode
+context = SetupBaseR(kem_output, node_private_key, group_context)
+path_secret = context.Open("", ciphertext)
+~~~
 
 # Key Schedule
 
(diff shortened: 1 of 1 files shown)


mulmarta:
Bringing up the issue mentioned in the comments to #648: this can be inconvenient for implementations because this way the update path can't be generated and encrypted at the same time.

From security perspective, I think it can't break anything, since we include strictly more in the context. On the other hand, it's not clear to me what it buys.

bifurcation:
I sympathize with the idea that this constraints applications.  MLSpp does exactly what you're talking about, updating and encrypting at the same time.  But I'm not really seeing why that refactoring matters all that much.  It's not any more net work, aside from maybe computing a filtered direct path twice.

draft-ietf-mls-protocol.md
```diff
@@ -3607,23 +3612,12 @@ uses:
...
-   the proposals and the UpdatePath (if any).  The new GroupContext uses the
-   epoch number for the new epoch, and the new confirmed transcript hash.  This
-   is used when deriving the new epoch secrets, and is the only GroupContext
-   that newly-added members will have.
+When creating or processing a Commit, a client updates the ratchet tree and
+GroupContext for the group.  These values advance from an "old" state reflecting
+the current epoch to a "new" state reflecting the new epoch initiated by the
+Commit.  When the Commit includes an UpdatePath, a "provisional" group context
+is constructed that reflects changes due to the proposals and UpdatePath, but
+with the old confirmed transcript hash.
```
rohan-wire, line 3620: If you want to keep the mental models, you could add:
"Conceptually, these three states can be thought of as three separate ratchet trees."

PR #689, opened 2022-05-21
Reorganize discussion of ratchet tree concepts

bifurcation:
Closes #670 
Closes #644 
Closes #669 

This PR reorganizes the exposition of tree concept to make them read more clearly.

* A new Paths through the Ratchet Tree section is introduced, which covers direct paths, copaths, and filtered direct paths
* The Ratchet Tree Nodes section is moved forward, so that the definition of filtered direct paths can refer to the notion of resolutions (see @Bren2010's note on #670)
* The path example expanded so that there are blanks in the tree, and so that it shows all three path types (direct, copath, and filtered direct) for every leaf.
* The section on verifying parent hashes uses the example tree from the path section to illustrate the concepts noted in #669

PR #694, opened 2022-05-24
Always use a full tree

### LINKED ISSUE #645: Always operate on a full tree
In order to use tree hashes to represent siblings in parent hashes, we had to compute a different tree hash on the right edge of the tree, pretending that the tree is full, in the sense of having `2^n` leaves.  Thus we actually have two tree hashes in play, one used for group agreement on the tree and one used for parent hash inputs.

We should consider specifying that the logical tree used by MLS is always full.  This change to the specification would not change the computations done by the protocol much, if at all.  Thanks to the fact that we no longer generate redundant nodes (#587), there is never anything non-blank to the right of the rightmost non-blank leaf anyway.  So the main impact is to insert some blank nodes between a parent node on the right edge of the tree and its right child -- and all of the algorithms in the specification already skip over such nodes.

Likewise, implementations' internal represntations of the tree wouldn't have to change much, since the extra nodes could be "virtual".  The only time they would be noticed is when computing tree hashes.  We would have to update the description of the `ratchet_tree` extension to say that it only covers the tree up to the rightmost non-blank leaf.

It would simplify the specification in a few ways, for example: 
* The "original sibling tree hash" would just entail blanking unmerged leaves, not structure changes.  
* The description of extending and truncating the tree would be much simpler.

The main cost would be extra hashes when joining, or when updating a member close to the right edge of the tree.  If you have a tree with `2^n + 1` leaves, then you have to compute tree hashes over `2^n - 1` blank leaves and their `2^n - 2` blank parents.

bifurcation:
Based on #689 

Closes #645

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1028,7 +1028,6 @@ in {{tree-hashes}}.
 
 The _resolution_ of a node is an ordered list of non-blank nodes
 that collectively cover all non-blank descendants of the node.  The resolution
-of a non-blank node with no unmerged leaves is just the node itself. More generally, the resolution
 of a node is effectively a depth-first, left-first enumeration of the nearest
 non-blank nodes below the node:
 
@@ -1108,7 +1107,7 @@ nodes are as follows:
 | B    | U, V, X     | A, W, Z  | U, X                 |
 | E    | Y, Z, X     | F, G, V  | Y, Z, X              |
 | F    | Y, Z, X     | E, G, V  | Y, Z, X              |
-| G    | U, V, X     | B, W, Z  | U, X                 |
+| G    | Z, X        | Y, V     | Z, X                 |
 
 ## Views of a Ratchet Tree {#views}
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -979,29 +979,18 @@ number of leaf nodes it contains.  For a given parent node, its _left
 subtree_ is the subtree with its left child as head (respectively
 _right subtree_).
 
-All trees used in this protocol are left-balanced binary trees. A
-binary tree is _full_ (and _balanced_) if its size is a power of
-two and for any parent node in the tree, its left and right subtrees
-have the same size.
-
-A binary tree is _left-balanced_ if for every
-parent, either the parent is balanced, or the left subtree of that
-parent is the largest full subtree that could be constructed from
-the leaves present in the parent's own subtree.
-Given a list of `n` items, there is a unique left-balanced
-binary tree structure with these elements as leaves.
-
-(Note that left-balanced binary trees are the same structure that is
-used for the Merkle trees in the Certificate Transparency protocol
-{{?I-D.ietf-trans-rfc6962-bis}}.)
+Every tree used in this protocol is a perfect binary tree, that is, a complete
+balanced binary tree with `2^d` leaves all at the same depth `d`.  This
+structure is unique for a given depth `d`.
 
 There are multiple ways that an implementation might represent a ratchet tree in
-memory.  For example, left-balanced binary trees can be represented as an array
-of nodes, with node relationships computed based on nodes' indices in the array.
-Or a more traditional representation of linked node objects may be used.
-{{array-based-trees}} and {{link-based-trees}} provide some details on how to
-implement the tree operations required for MLS in these representations.
-MLS places no requirements on implementations' internal representations
+memory.  A convenient property of the complete trees used here is that they can
+be represented as an array of nodes, with node relationships computed based on
+the nodes indices in the array.  A more traditional representation based on
+linked node object may also be used.  {{array-based-trees}} and
+{{link-based-trees}} provide some details on how to implement the tree
+operations required for MLS in these representations.  MLS places no
+requirements on implementations' internal representations
 of ratchet trees.  An implementation MAY use any tree representation and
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -979,68 +979,124 @@ number of leaf nodes it contains.  For a given parent node, its _left
...
+  node was changed (see {{parent-hash}}).
+
+Every node, regardless of whether the node is blank or populated, has
+a corresponding _hash_ that summarizes the contents of the subtree
+below that node.  The rules for computing these hashes are described
+in {{tree-hashes}}.
+
+The _resolution_ of a node is an ordered list of non-blank nodes
+that collectively cover all non-blank descendants of the node.  The resolution
+of a node is effectively a depth-first, left-first enumeration of the nearest
```
TWal, line 1020:
```suggestion
of a node is effectively a left-first enumeration of the nearest
```

I find the "depth-first" misleading, e.g. in the following tree:
```
      Y
    __|__
   /     \
  _       Z
 / \     / \
A   B   C   D
```
the resolution of Y is A, B, Z and not Z, A, B.

Thinking more about it, this high-level description of what is a "resolution" is a bit weird because it doesn't takes into account unmerged leaves.
I think a better high-level description of the "resolution" would be something like "The resolution of a node contains the smallest set of keys required to encrypt a secret to its subtree".

What do you think?

draft-ietf-mls-protocol.md
```diff
@@ -4993,11 +5028,11 @@ parent=01x => left=00x, right=10x
...
 
 Since node relationships are implicit, the algorithms for adding and removing
-nodes at the right edge of the tree are quite simple:
+nodes at the right edge of the tree are quite simple.  If there are `N` nodes in
+the array:
 
-* Add: Append a blank parent node to the array of nodes, then append the new
-  leaf node
-* Remove: Remove the rightmost two nodes from the array of nodes
+* Add: Append `N - 1` blank values to the end of the array.
```
TWal, line 5034:
```suggestion
* Add: Append `N + 1` blank values to the end of the array.
```

I think it's N+1 because you need the new right tree of the root (N nodes) + 1 node for the root

TWal (commented):
Using full tree + filtered direct path seems like a fine combo!

As you said in #645 it not only faster, but also makes things a lot easier to understand in various places of the protocol.

I also like a lot the two examples you added (the one for filtered direct path, and the one for parent-hash link), they have a nice pedagogical value!

bifurcation:
Interim 2022-05-26:

* This will require some optimized implementations
* Have people done performance evaulations?
    * @bifurcation and @TWal did evaluation on parent hash, this should be easier
* How does this affect the secret tree?
    * Would change the structure to be a full tree
    * Always `log(N)` derivations to get to a leaf, instead of sometimes less
* TODO: Hold this for next week, @bifurcation and @TWal to do performance evaluation

draft-ietf-mls-protocol.md
bifurcation: The point of the "depth-first, left-first" language is to clarify the order in which the nodes in the resolution are listed.  As I read it, the language is correct, or maybe better as "left-first, depth first", in the sense that you visit the left child before the right child, and if you find a blank left child, you descend before going to the right child.  Following those rules, you do indeed end up at [A, B, Z] in your example -- though of course the resolution of Y should be [Y] ;)

In any case, I will handle this in a follow-on, since this text is an effect of #689, not this PR.

bifurcation:
Mailing list post with performance metrics:
https://mailarchive.ietf.org/arch/msg/mls/42tmYeZIJRAAZTxnSb6pzXKZGa0/

draft-ietf-mls-protocol.md
```diff
@@ -4954,33 +4993,29 @@ To construct the tree in {{parent-hash-tree}}:
 
 # Array-Based Trees
 
-One benefit of using left-balanced trees is that they admit a simple
```
Bren2010, line 4957: I would've leaned towards leaving this section as it was, since left-balanced and complete are supposed to be functionally equivalent but left-balanced is more memory efficient.

bifurcation:
Interim 2022-06-02:

* @bifurcation presented performance data showing +/- ~20% on join
* @Bren2010 asked about removing leaf index from LeafNodeHashInput in order to allow for precomputing blank subtree hashes
    * Even with leaf indices, can pre-generate, they just have to be in-place
    * No action
* Clear to merge once editorial comments from @Bren2010 are addressed

draft-ietf-mls-protocol.md
```diff
@@ -979,68 +979,124 @@ number of leaf nodes it contains.  For a given parent node, its _left
...
+operations required for MLS in these representations.  MLS places no
+requirements on implementations' internal representations
+of ratchet trees.  An implementation MAY use any tree representation and
+associated algorithms, as long as they produce correct protocol messages.
+
+### Ratchet Tree Nodes
+
+Each leaf node in a ratchet tree is given an _index_ (or _leaf index_), starting
+at `0` from the left to `n-1` at the right (for a tree with `n` leaves). A tree
+with `n` leaves has `2*n - 1` nodes, including parent nodes.
```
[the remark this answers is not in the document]
bifurcation:
```suggestion
with `2^d` leaves has `2^(d+1) - 1` nodes, including parent nodes.
```

This is still accurate, but we can phrase it more accurately now that `n` is always a power of 2.

PR #698, opened 2022-05-25
Editorial

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -3314,10 +3309,9 @@ struct {
 } Proposal;
 ~~~
 
-On receiving an MLSMessageContent containing a Proposal, a client MUST verify the
-signature inside MLSMessageAuth.  If the signature verifies
-successfully, then the Proposal should be cached in such a way that it can be
-retrieved by hash (as a ProposalOrRef object) in a later Commit message.
+Upon receiving an MLSMessageContent containing a Proposal which has been
+successfully decrypted and/or authenticated, the Proposal is cached in such a
+way that it can be easily referenced by a ProposalRef in a later Commit message.
```
bifurcation, line 3314: I would prefer to keep the explicit requirement to verify the signature.

bifurcation (approved):
Couple of minor things, otherwise lgtm.

draft-ietf-mls-protocol.md
```diff
@@ -4662,14 +4657,9 @@ in an asynchronous application, it may be the case that all members that could
...
 accepted by the group, and the resulting state possibly used as the basis for
 further Commits.  When the affected members come back online, they will reject
-the first commit, and thus be unable to catch up with the group.
-
-Applications can address this risk by requiring certain members of the group to
-acknowledge successful processing of a Commit before the group regards the
-Commit as accepted.  The minimum set of acknowledgements necessary to verify
-that a Commit is well-formed comprises an acknowledgement from one member per
-node in the UpdatePath, that is, one member from each subtree rooted in the
-copath node corresponding to the node in the UpdatePath.
```
raphaelrobert, line 4672: Why remove this? It sounds like a viable strategy in some cases.

Bren2010: I thought about this more and I still don't like this section:
- It says you need one acknowledgement per node in the UpdatePath, but that's not really true. You would need one acknowledgement per *ciphertext* in the UpdatePath since the ciphertexts are independent.
- That's assuming that everybody votes honestly. If the attacker colludes with someone to vote dishonestly, then the same attack still works. So I need acknowledgement from *everybody* in the group.

Bren2010: There's really not a reasonable way to prevent this issue. Can only provide reporting mechanism and make it right when it happens

bifurcation (approved):

PR #699, opened 2022-05-25
Clarify some parts of Application Messages section.

### LINKED ISSUE #679: Content encryption: Clarify generation sequence
The generation numbers should be increased sequentially, gaps should be avoided to avoid unnecessary expensive operations on clients. Clients might also abort if the gap is too large to prevent attacks from malicious insiders.

Bren2010:
Closes #679

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -4486,7 +4486,7 @@ steps that clients will move the ratchet forward in response to a new message. M
 with an unusually high generation counter would then be rejected to avoid
 causing a denial-of-service attack by requiring the recipient to perform an
 excessive number of key derivations.  For example, a malicious group member
-could send a message with `generation = 0xffffffff` at the beginning of a new 
+could send a message with `generation = 0xffffffff` at the beginning of a new
 epoch, forcing recipients to perform billions of key derivations.
 
 # Security Considerations
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -4420,54 +4420,37 @@ for some clients if they keep failing to get their proposal accepted.
 
 # Application Messages
 
-The primary purpose of the Handshake protocol is to provide an
-authenticated group key exchange to clients. In order to protect
-Application messages sent among the members of a group, the Application
-secret provided by the Handshake key schedule is used to derive nonces
-and encryption keys for the Message Protection Layer according to
-the Application Key Schedule. That is, each epoch is equipped with
-a fresh Application Key Schedule which consist of a tree of Application
-Secrets as well as one symmetric ratchet per group member.
+The primary purpose of handshake messages are to provide an authenticated group
+key exchange to clients. In order to protect application messages sent among the
+members of a group, the `encryption_secret` provided by the key schedule is used
+to derive a sequence of nonces and keys for message encryption. Every epoch
+moves the key schedule forward which triggers the creation of a new secret
+tree, as described in {{secret-tree}}, along with a new set of symmetric
+ratchets of nonces and keys for each member.
 
-Each client maintains their own local copy of the Application Key
-Schedule for each epoch during which they are a group member. They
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -4488,16 +4471,20 @@ state compromise.
...
+are received out of order, the client moves the sender ratchet forward to match
+the received generation counter. Any unused nonce and key pairs from the ratchet
+are potentially stored so that they can be used to decrypt the messages which
+were delayed or reordered.
+
+Applications SHOULD define a policy on how long to keep unused nonce and key
+pairs for a sender, and the maximum number to keep. This is in addition to
+ensuring that these nonce and key pairs are promptly deleted when the epoch
+ends. Applications SHOULD also define a policy limiting the maximum number of
+steps that clients will move the ratchet forward in one step. Messages received
```
bifurcation, line 4485:
```suggestion
steps that clients will move the ratchet forward in response to a new message. Messages received
```

draft-ietf-mls-protocol.md
```diff
@@ -4488,16 +4471,20 @@ state compromise.
...
+with an unusually high generation counter would then be rejected to avoid
+causing a denial-of-service attack.
```
bifurcation, line 4487:
```suggestion
causing a denial-of-service attack by requiring the recipient to perform an
excessive number of key derivations.  For example, a malicious group member
could send a message with `generation = 0xffffffff` at the beginning of a new 
epoch, forcing recipients to perform billions of key derivations.
```

bifurcation (approved):
Couple of friendly amendments, otherwise lgtm.

I am going to hold this as slightly more than editorial, given the number of MUSTs that are affected.  But I would note for others that the MUSTs that are removed here are all redundant with requirements above.

draft-ietf-mls-protocol.md
```diff
@@ -4488,16 +4471,23 @@ state compromise.
...
+
+Applications SHOULD define a policy on how long to keep unused nonce and key
+pairs for a sender, and the maximum number to keep. This is in addition to
+ensuring that these nonce and key pairs are promptly deleted when the epoch
+ends. Applications SHOULD also define a policy limiting the maximum number of
+steps that clients will move the ratchet forward in response to a new message. Messages received
+with an unusually high generation counter would then be rejected to avoid
+causing a denial-of-service attack by requiring the recipient to perform an
+excessive number of key derivations.  For example, a malicious group member
+could send a message with `generation = 0xffffffff` at the beginning of a new
```
Bren2010, line 4489:
```suggestion
could send a message with `generation = 0xffffffff` at the beginning of a new
```

draft-ietf-mls-protocol.md
```diff
@@ -4488,16 +4471,23 @@ state compromise.
...
+the received generation counter. Any unused nonce and key pairs from the ratchet
+are potentially stored so that they can be used to decrypt the messages which
+were delayed or reordered.
+
+Applications SHOULD define a policy on how long to keep unused nonce and key
+pairs for a sender, and the maximum number to keep. This is in addition to
+ensuring that these nonce and key pairs are promptly deleted when the epoch
+ends. Applications SHOULD also define a policy limiting the maximum number of
+steps that clients will move the ratchet forward in response to a new message. Messages received
+with an unusually high generation counter would then be rejected to avoid
```
raphaelrobert, line 4486: "unusually high" is a bit vague. Maybe we can be more specific here by saying that clients can enforce a maximum difference for the generation counter between two subsequent messages.

raphaelrobert (approved):
Thanks for picking this up! Just one nit, but looks good otherwise.

bifurcation:
Interim 2022-05-26:

* Intended to be mostly editorial
* Leaving open for now
* Clear to merge in between meetings if folks are comfortable

bifurcation:
Interim 2022-06-02:

* Still clear from last week

PR #703, opened 2022-05-27
Be more precise about credential validation

### LINKED ISSUE #678: Be more specific about Credential verification
While this is spread out throughout the document, it would be beneficial to spell out in which scenarios a Credential MUST be verified (with the AS):

 - When fetching a KeyPackage to add a new member
 - When discovering a KeyPackage in an AddProposal from another member
 - When joining a new group (verify all other members' Credentials), both when receiving a Welcome message and when joining through an External Commit
 - When a member updates its leaf
 - ... anywhere else?

bifurcation:
Closes #678 
Replaces #660 

This PR attempts to add precision to the credential validation requirements in the document, along the lines @raphaelrobert suggests in #678, in the spirit of what @Bren2010 wrote up in #660 and drawing inspiration from what RFC 6125 does in the context of TLS.  It seems like this gives a clearer definition of (a) what the AS's job is and (b) when it needs to be done.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1354,6 +1354,8 @@ the AS.  In particular, at the following events in the protocol:
   for the member.
 * When a member receives a Commit with an UpdatePath whose LeafNode has a new
   credential for the committer.
+* When an `external_senders` extension is added to the group, or an existing
+  `external_senders` extension is updated.
 
 In cases where a member's credential is being replaced, such as Update and
 Commit cases above, the AS MUST also verify that the set of presented
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1218,6 +1218,10 @@ NIST curves (P-256, P-384, or P-521), the public key is the output of the
 uncompressed Elliptic-Curve-Point-to-Octet-String conversion according to
 {{SECG}}.
 
+The signatures used in this document are encoded as specified in {{!RFC8446}}.
+In particular, ECDSA signatures are DER-encoded and EdDSA signatures are defined
+as the concatenation of `r` and `s` as specified in {{?RFC8032}}.
+
 To disambiguate different signatures used in MLS, each signed value is prefixed
 by a label as shown below:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1321,18 +1325,41 @@ the issuer of the previous certificate.  The public key encoded in the
...
-client presents a new credential, the application MUST verify that the set
-of identifiers in the credential is acceptable to the application for this
-client.
+### Credential Validation
+
+The application using MLS is responsible for specifying which identifiers it
+finds acceptable for each member in a group.  In other words, following the
+model that {{?RFC6125}} describes for TLS, the application maintains a list of
+"reference identifiers" for the members of a group, and the credentials provide
+"presented identifiers".  A member of a group are authenticated by first
```
Bren2010, line 1334:
```suggestion
"presented identifiers".  A member of a group is authenticated by first
```

draft-ietf-mls-protocol.md
```diff
@@ -1321,18 +1325,41 @@ the issuer of the previous certificate.  The public key encoded in the
...
+
+* When a member receives a KeyPackage that it will use in an Add proposal to add
+  a new member to the group.
+* When a member receives a GroupInfo object that it will use to join a group,
+  either via a Welcome or via an external Commit
+* When a member receives an Add proposal adding a member to the group.
+* When a member receives an Update proposal whose LeafNode has a new credential
+  for the member.
+* When a member receives a Commit with an UpdatePath whose LeafNode has a new
+  credential for the committer.
```
Bren2010, line 1356: In each of these cases, where does the reference identifier come from? The reference/presented identifier distinction makes sense in TLS because you have a domain the user has typed in. Not as clear in MLS.

Bren2010 (approved):

draft-ietf-mls-protocol.md
```diff
@@ -1321,18 +1325,41 @@ the issuer of the previous certificate.  The public key encoded in the
...
+the credential's presented identifiers, and verifies that those identifiers
+match the reference identifiers for the member.
+
+Whenever a new credential is introduced in the group, it MUST be validated with
+the AS.  In particular, at the following events in the protocol:
+
+* When a member receives a KeyPackage that it will use in an Add proposal to add
+  a new member to the group.
+* When a member receives a GroupInfo object that it will use to join a group,
+  either via a Welcome or via an external Commit
```
raphaelrobert, line 1351:
```suggestion
  either via a Welcome or via an External Commit
```

raphaelrobert (approved):
Thanks for picking it up!

kkohbrok (commented):
Looks good in general, thanks for the PR!

draft-ietf-mls-protocol.md
kkohbrok, line 1356: We might want to also do this check whenever a new credential shows up in the External Senders extension.

bifurcation: It's not so different in messaging applications.  When you add someone to the group, you type in some identifier for them.  And there's typically a way that the application tells you who is in the group, independent of the crypto.  E.g., if you were mapping this on to XMPP, you might compare the credentials to [the JIDs in the roster](https://xmpp.org/rfcs/rfc6121.html#roster-syntax-items-jid).

PR #704, opened 2022-05-28
Minimize sequencing section.

Bren2010:


draft-ietf-mls-protocol.md
```diff
@@ -4441,72 +4441,26 @@ current state of the group.  In practice, however, there is a risk
...
+Commit out of several sent in an epoch will be canonical. The approach chosen
+MUST minimize the amount of time that forked or previous group states are kept
+in memory, and promptly delete them once they're no longer necessary to ensure
+forward secrecy.
+
+The generation of Commit messages MUST NOT modify a client's state, since the
+client doesn't know at that time whether the changes implied by the Commit
+message will conflict with another Commit or not. Similarly, a client MUST NOT
+send Welcome messages corresponding to a Commit until it's clear that the Commit
+has been accepted.
```
bifurcation, line 4456: Interim 2022-06-02:

* Probably too tight.  E.g., client could send Commit+Welcome to DS, then DS could decide whether to send Welcome based on whether Commit is accepted.  Proposed text:

> The Welcome message corresponding to a Commit MUST NOT be delivered to a new joiner until it's clear that the Commit has been accepted.

* Otherwise clear to merge

Bren2010, line 4456:
```suggestion
message will conflict with another Commit or not. Similarly, the Welcome
message corresponding to a Commit MUST NOT be delivered to a new
joiner until it's clear that the Commit has been accepted.
```

PR #705, opened 2022-05-28
Clarify rules around Add proposals.

### LINKED ISSUE #700: Not allowing the same member to join the group multiple times?
In #675, it was pointed out that we previously had a clause:

> If there are multiple Add proposals containing KeyPackages that the committer considers to represent the same client or a client already in the group (for example, identical KeyPackages or KeyPackages sharing the same Credential), the committer again chooses one to include and considers the rest invalid.

But this was amplified in #675 to say that we also can't send Add proposals for users that are already in the group. I think we should talk about if this is the actual desired behavior or not.

Bren2010:
Closes #700

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3340,12 +3340,8 @@ struct {
 } Add;
 ~~~
 
-An Add proposal is invalid if any of the following is true:
-
-* The KeyPackage is invalid according to {{keypackage-validation}}.
-
-* The Credential in the KeyPackage represents a client already in the
-  group according to the application.
+An Add proposal is invalid if the KeyPackage is invalid according to
+{{keypackage-validation}}.
 
 An Add is applied after being included in a Commit message.  The position of the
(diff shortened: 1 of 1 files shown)


bifurcation (approved):

draft-ietf-mls-protocol.md
```diff
@@ -4262,8 +4258,11 @@ occurs:
   received, or the most recent Update if there are no Removes.
 
 * It contains multiple Add proposals that contain KeyPackages that represent the same
-  client according to the application (for example, identical KeyPackages or KeyPackages
-  sharing the same Credential).
+  client according to the application (for example, identical signature keys).
+
+* It contains an Add proposal with a KeyPackage that represents a client already
+  in the group according to the application, which isn't being removed by
+  another proposal in the same Commit.
```
bifurcation, line 4265:
```suggestion
  in the group according to the application, unless there is a Remove proposal
  in the list removing the matching client from the group.
```

Clarifies what it means for a client to be "removed by another proposal", and keeps it scoped to the proposal list.

PR #707, opened 2022-05-29
Derive the commit secret from the end of the UpdatePath, not the root

### LINKED ISSUE #690: Commit secret is ambiguous when root is blank
As I was pondering pathological cases related to #645, it occurred to me that we already have a slightly odd situation that can arise as a result of the suppression of redundant nodes.  When a Commit has an UpdatePath, the `commit_secret` is defined as follows:

> Define commit_secret as the value path_secret[n+1] derived from the path_secret[n] value assigned to the root node.

However, there are situations where the root node is not assigned a value, namely when the left half of the tree is blank.  For example, suppose A creates a group with B, C, D:

```
      Y
    __|__
   /     \
  X       _ = Z
 / \     / \
A   B   C   D
```

Suppose that then C removes A and B, leaving C and D as the only leaves in the tree.  After removing A and B, the tree on which C computes an UpdatePath is as follows:

```
      _ = Y
    __|__
   /     \
  _       _ = Z
 / \     / \
_   _   C   D  
```

C's direct path is (Z, Y), but it's filtered direct path is just (Z) because the copath subtree of Y is all blank.  So C will not generate a path_secret for Y, which means that the commit secret is not really defined.

It seems like there are two paths to fix this:

1. Keep the current definition of `commit_secret` and require that a path secret always be generated for the root node
    * Presumably the root node path secret would be generated directly from the path secret in the UpdatePath
    * This would be simple to implement by just forcing the root node to be in the filtered direct path
2. Keep the current path secret generation and specify that the `commit_secret` is derived from the last path secret in the UpdatePath 
    * In the above example, the last path secret from the path secret for Z

I lean slightly toward (2) because it makes the computation the same for two neighbors regardless of where they are in the tree. It seems like the two options primarily differ in whether there are one or two invocations of the key derivation.

bifurcation:
Fixes #690

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3721,7 +3721,7 @@ message at the same time, by taking the following steps:
   * Generate path secrets for the parent nodes along the sender's filtered
     direct path, as described in {{synchronizing-views-of-the-tree}}.  Define
     `commit_secret` as the value `path_secret[n+1]` derived from the
-    `path_secret[n]` value assigned to the root node.
+    last path secret value (`path_secret[n]`) derived for the UpdatePath.
 
   * Update the new ratchet tree by setting the parent nodes on the sender's
     filtered direct path based on the path secrets. Compute parent hashes for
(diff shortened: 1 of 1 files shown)


Bren2010 (approved):

bifurcation:
Interim 2022-06-02

* As @TWal notes in his comments, tree malleability is not worth blocking this for because you won't block the tree-extension attack anyway
* @mulmarta notes the exception case where the UpdatePath has no nodes in it
    * @bifurcation suggests we should just forbid that case
    * @bifurcation will file a new PR to forbid that case

PR #708, opened 2022-05-29
Describe requests to join and directed KeyPackages

### LINKED ISSUE #693: External Add proposals revisited
@rohan-wire raised a few points today, which I'm trying to capture in this issue with some thoughts on proposed solutions.

In #659, we removed the usage of `new_member` to send Add proposals, in order to allow for the signature on an external commit to cover the GroupContext (which is not necessarily possible for an external Add sender).  In the discussion, the point was made that an Add proposal only really differs from a KeyPackage in that the KeyPackage doesn't have `group_id` or `epoch` fields.  Part of the resolution in #659 was a decision that folks didn't think `group_id` was a significant loss.  And I think there's pretty broad agreement that `epoch` is actively undesirable, since it means that the new joiner has to be sufficiently in sync with the group to know the epoch hasn't changed before sending its proposal.

After talking with @rohan-wire, there are two other issues to highlight:

* Losing `group_id` is a bigger deal than it might first seem in a system that has both "prepublished KeyPackages" and "KeyPackage as request to be added".  Someone could go grab a bunch of prepublished KeyPackages that they don't own, and send them in as requests to be added.  This isn't an immediate security risk, since the attacker doesn't necessarily have the private keys that go with the KeyPackages (if they do, they have other attack avenues).  But it would mean, for example, that a client would have an even less certain idea of all the groups it was a member of.

* When "requests to be added" were expressed as Proposals, the "request" semantic followed from being a Proposal, which tied into the recommendations for what should be included in a Commit.  Expressing these requests in a different way, means that the "request" semantic rests outside of the core MLS logic.

A few possible avenues come to mind for addressing these issues:

1. Don't change the protocol, but document the operational considerations.  For example:
    * The "unintended join" issue with abuse of prepublished KPs could be mitigated by a DS that supports proactive joins enforcing that the sender owns the signature key (e.g., checking that `kp.leaf_node.credential` authenticates the same identity that the DS has for the client).
    * The request semantic could be captured in the architecture document / protocol overview as a thing that applications could implement if they choose. 

2. Re-add joiner-sent Add proposals, using a different SenderType this time to avoid the conflict with external commits.
     * Say `new_member_add_proposal` and `new_member_commit`
     * This would still have the epoch problem.

3. Add a field to KeyPackage indicating whether the KeyPackage is general or directed to a specific group
    * Something like `optional<GroupID> target_group_id`
    * Require that if the field is set in an Add proposal, then it MUST match the group's ID.
    * Specify that only directed KPs can be used to request to be added
    * Add parallel text to proposals that recommend the the committer commit Add proposals for any directed KPs it has received.

4. Define a new top level message (an entry in the MLSMessage enum) that expresses precisely the semantic we want, something like:

```
struct {
    opaque group_id<V>;
    KeyPackage key_package;
    // SignWithLabel(., "JoinRequestTBS", JoinRequestTBS)
    opaque signature<V>;
} JoinRequest;
```

bifurcation:
Closes #693 

(At least, this is one approach to closing #693)

This PR implements the "directed KeyPackage" approach discussed in #693.

It uses a KeyPackage extension to carry the target group ID.  This could also be an optional field on KeyPackage. I mainly shied away from that because the mechanics of the TLS presentation syntax would require us to define something like `struct { opaque group_id<V>; } GroupID;` (so that we could have `optional<GroupID>`), which seemed ugly.

I've also added a subsection to the protocol overview that discusses requests to be added.  I would suggest we add such a subsection regardless of how we resolve #693, to make sure we provide good context  for how the mechanism we define is used.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -870,6 +870,36 @@ A              B     ...      Z          Directory       Channel
 ~~~
 {: title="Client Z removes client B from the group"}
 
+## Requesting to Join
+
+In addition to the directory-based distribution of KeyPackages illustrated
+above, applications may allow a prospective joiner to provide a KeyPackage
+directly to a member of the group, as a way of asking to be added to the group.
+
+~~~ aasvg
+                                                          Group
+A              B     ...      Z          Directory       Channel
+|  KeyPackageB |              |              |              |
+|<-------------+              |              |              |
+|              |              |              |              |
+| Welcome(B)   |              |              |              |
++------------->|              |              |              |
+|              |              |              |              |
+| Add(B)       |              |              |              |
+| Commit(Add)  |              |              |              |
++---------------------------------------------------------->|
+|              |              |              |              |
+|              |              |              | Add(B)       |
+|              |              |              | Commit(Add)  |
+|<----------------------------------------------------------+
+|              |              |<----------------------------+
+|              |              |              |              |
+~~~
+
+For applications that allow both directory-based KeyPackage distribution and
+requests to join, some additional security considerations are discussed in
+{{directed-keypackages}}.
+
 ## Relationships Between Epochs
 
 A group has a single linear sequence of epochs. Groups and epochs are generally
@@ -3133,6 +3163,31 @@ set to `key_package`.
 Extension included in the `extensions` or `leaf_node.extensions` fields MUST be
 included in the `leaf_node.capabilities` field.
 
+## Directed KeyPackages
+
+Before a new member can be added to the group, the member that proposes their
+addition needs to get a KeyPackage for the new member. As discussed in
+{{protocol-overview}}, this can be done either by pre-publishing KeyPackages
+to a directory or by having the joining member send them proactively.
+
+In the latter case, the sender of the KeyPackage can limit the scope of the
+KeyPackage by including an extension of type `target_group_id`, of the following
+form:
+
+```
+opaque group_id<V>;
+```
+
+As defined in {{keypackage-validation}}, when this extension is present, the
+members of the group will verify that the KeyPackage is intended for the group.
+
+In applications where both pre-published and directed KeyPackage distribution
+are supported, the application SHOULD require that this extension be present in
+a KeyPackage sent proactively by a joiner, and absent in pre-published
+KeyPackages. This prevents pre-published KeyPackages from being fetched by a
+third party and used to request membership in a group (since the pre-published
+KeyPackages will not have this extension).
+
 ## KeyPackage Validation
 
 The validity of a KeyPackage needs to be verified at a few stages:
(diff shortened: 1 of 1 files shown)


raphaelrobert:
I think that's an excellent candidate for the MLS Extensions document, given that it's a pretty niche feature and that it is written as an extension.

draft-ietf-mls-protocol.md
```diff
@@ -3133,6 +3163,31 @@ set to `key_package`.
...
 
+## Directed KeyPackages
+
+Before a new member can be added to the group, the member that proposes their
+addition needs to get a KeyPackage for the new member. As discussed in
+{{protocol-overview}}, this can be done either by pre-publishing KeyPackages
+to a directory or by having the joining member send them proactively.
+
+In the latter case, the sender of the KeyPackage can limit the scope of the
+KeyPackage by including an extension of type `target_group_id`, of the following
```
Bren2010, line 3174:
```suggestion
KeyPackage by including an extension of type `group_id`, of the following
```

draft-ietf-mls-protocol.md
```diff
@@ -3155,6 +3210,10 @@ The client verifies the validity of a KeyPackage using the following steps:
 * Verify that the value of `leaf_node.encryption_key` is different from the value of
   the `init_key` field.
 
+* If the `extensions` field contains an extension of type `target_group_id`,
```
Bren2010, line 3213:
```suggestion
* If the `extensions` field contains an extension of type `group_id`,
```

draft-ietf-mls-protocol.md
```diff
@@ -3133,6 +3163,31 @@ set to `key_package`.
...
+
+As defined in {{keypackage-validation}}, when this extension is present, the
+members of the group will verify that the KeyPackage is intended for the group.
+
+In applications where both pre-published and directed KeyPackage distribution
+are supported, the application SHOULD require that this extension be present in
+a KeyPackage sent proactively by a joiner, and absent in pre-published
+KeyPackages. This prevents pre-published KeyPackages from being fetched by a
+third party and used to request membership in a group (since the pre-published
+KeyPackages will not have this extension).
```
Bren2010, line 3189: I don't understand the value of having this extension. If pre-published KeyPackages already exist without this extension, then I can just use them to do whatever nefarious thing you're trying to stop from happening with the directed KeyPackages.

Bren2010:
Making this peer-to-peer (B requests a join from A) seems like an anti-pattern given that MLS is supposed to be async. Might make more sense to have the dominant example be: B requests a join from pre-configured external sender service. Phrasing it that way actually answers my question from before about why we need this extension. The pre-configured external sender service needs the group_id extension, while normal members of the group don't.

bifurcation:
Replying to two points from @Bren2010 

* Re: The value of this extension - The way you get a security property here is if the application knows which channel (prepublished or proactive) the KeyPackage came through, and rejects non-directed KeyPackages in the proactive channe.  It also give the client posting the KeyPackage some control: If my client only ever joins proactively, I can make it so that my KeyPackages can't be replayed from one group to another.

* Re: Async - This pattern doesn't break the async property any more than prepublication does.  The overall message flow is the same as the prepublicaiton case; the only difference is whether the KeyPackage is pulled or pushed.

PR #713, opened 2022-06-01
Improve parent hash guarantees

TWal:
# A nice security property for parent-hash

The goal of #527 was to have a property similar to:

> If there is a parent-hash chain from the leaf L to a node N, then the signature of L binds the tree Canonicalize(N)

(where Canonicalize(N) is N with its unmerged leaves removed, i.e. the subtree of N the last time a Commit went through it)

However, this is currently not true, there are a few places where we can do (hopefully benign) modifications on N.

Example 1:
```
              W
        ______|_____
       /             \
      _               Y
    __|__           __|__
   /     \         /     \
  T       _       X       Z
 / \     / \     / \     / \
A   B   _   _   E   F   G   H
```

with the parent-hash chain A-T-W (happening with filtered update path).

We can modify the blank subtree:

```
              W
        ______|_____
       /             \
      _               Y
    __|__           __|__
   /     \         /     \
  T       V       X       Z
 / \     / \     / \     / \
A   B   C   D   E   F   G   H
```

and the parent-hash chain A-T-W is still valid

Example 2:

```
              W
        ______|_____
       /             \
      U               Y
    __|__           __|__
   /     \         /     \
  T       V       X       Z
 / \     / \     / \     / \
A   B   _   D   E   F   G   H
```

with the parent-hash chain A-T-U-W.

We can add unmerged leaves:

```
              W
        ______|_____
       /             \
      U[C]            Y
    __|__           __|__
   /     \         /     \
  T       V[C]    X       Z
 / \     / \     / \     / \
A   B   C   D   E   F   G   H
```

and the parent-hash chain A-T-U-W is still valid.

# How to get this property

These examples are not possible if we require the parent-hash check to verify other invariants that exists in MLS:
- if there is a parent-hash link from U to P and there are blank nodes between U and P, these blank nodes are here because of the filtered update path so their sibling have empty resolution (this invariant forbids example 1)
- if there is a parent-hash link from U to P, then there can't be a leaf which is merged for U but not for P, because U and P were last modified by a commit at the same time (this invariant forbids example 2)

This PR add these checks to the parent hash validation procedure.

# Proof that we do have the property with this PR

## Definitions

Un-adding leaves:
Let N be a node, and Ls a list of leaves.
Then UnAdd(N, Ls) is a tree identical to N, except that the leaves of Ls are blanked and removed from every node's unmerged leaves list.
Informally, this operation is the inverse of adding leaves Ls in N by participant not in N.

Canonicalization:
We define Canonicalize(N) = UnAdd(N, N.unmerged_leaves).

Tree equivalence:
We say T1 ~= T2 when Canonicalize(T1) = Canonicalize(T2).

Parent-hash link:
We say C ~> P when "the parent hash in a node C is valid with respect to a parent node P" as described in the "Verifying Parent Hashes" section (modified by this PR)

# Key property on ~= and ~>

We prove the following property:
If C ~> P
and C' ~> P'
and C ~= C'
then P ~= P'.

This is better understood with a nice square:
```
If we have:

C ~= C'
~    ~
V    V
P    P'

Then there is a ~= between P and P'
```

Proof of this property:

Let S (resp. S') be the sibling in the C ~> P (resp. C' ~> P') relation. Because leaf indices are inside the tree hash, S and S' are both on the same side of P and P'.

Fact 1: UnAdd(S, P.unmerged_leaves) = UnAdd(S', P'.unmerged_leaves).
Proof: By definition of the parent hash and injectivity of tree-hash.

Fact 2: UnAdd(C, P.unmerged_leaves) = UnAdd(C', P'.unmerged_leaves).
Proof:

```
  UnAdd(C, P.unmerged_leaves)
= UnAdd(C, Intersection(P.unmerged_leaves, Leaves(C))     by a simple lemma on UnAdd
= UnAdd(C, C.unmerged_leaves)                             by C ~> P (we use the second new invariant here!)
= UnAdd(C', C'.unmerged_leaves)                           by C ~= C'
= UnAdd(C', Intersection(P'.unmerged_leaves, Leaves(C'))) by C' ~> P' (we use the second new invariant here!)
= UnAdd(C', P'.unmerged_leaves)                           by a simple lemma on UnAdd
```

With Fact 1 and Fact 2, we deduce that P ~= P' (we use the first new invariant here!).

# Proof of signature binding

A leaf signature binds the existence of nodes Mi suh that L ~> M1 ~> ... ~> Mm = root

When there is a parent-hash-link path to a node P starting from a leaf L like L ~> N1 ~> ... ~> Nn ~> P, by applying inductively the property on ~= and ~> you get P ~= Mi for some i, hence L's signature binds Canonicalize(P) = Canonicalize(Mi).

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2387,16 +2387,21 @@ Parent hashes are verified at two points in the protocol: When joining a group
 and when processing a Commit.
 
 The parent hash in a node U is valid with respect to a parent node P if the
-following criteria hold:
+following criteria hold.  Here C and S are the children of P (for "child" and
+"sibling"), with C being the child that is on the direct path of U (possibly U
+itself) and S the other child:
 
-* U is a descendant of P in the tree
-* The nodes between U and P in the tree are all blank
+* U is a descendant of P in the tree.
+
+* The resolution of C is equal to U added to the intersection of P's
+  `unmerged_leaves` with the subtree under C.
+
 * The `parent_hash` field of U is equal to the parent hash of P with copath
-  child S, where S is the child of P that is not on the path from U to P
-* The sibling of every node with a blank parent between U (included) and P
-  (excluded) has an empty resolution
-* `U.unmerged_leaves` is equal to the intersection between `P.unmerged_leaves`
-  and the leaves under U.
+  child S.
+
+These checks verify that U and P were updated at the same time (in the same
+UpdatePath), and that they were neighbors in the UpdatePath because the nodes in
+between them would have omitted from the filtered direct path.
 
 A parent node P is "parent-hash valid" if it can be chained back to a leaf node
 in this way.  That is, if there is leaf node L and a sequence of parent nodes
@@ -4111,16 +4116,15 @@ welcome_key = KDF.Expand(welcome_secret, "key", AEAD.Nk)
   * Verify that the tree hash of the ratchet tree matches the `tree_hash` field
     in GroupInfo.
 
-  * For each non-empty parent node, verify that exactly one of the node's
-    children are non-empty and have the hash of this node set as their
-    `parent_hash` value (if the child is another parent) or has a `parent_hash`
-    field in the LeafNode containing the same value (if the child is a
-    leaf). If either of the node's children is empty, and in particular does not
-    have a parent hash, then its respective children's `parent_hash` values have
-    to be considered instead.
+  * For each non-empty parent node, verify that it is "parent-hash valid",
+    as described in {{verifying-parent-hashes}}.
 
   * For each non-empty leaf node, validate the LeafNode as described in
     {{leaf-node-validation}}.
+
+  * For each non-empty parent node, verify that each entry in the node's
+    `unmerged_leaves` represents a non-blank leaf node that is a descendant of
+    the parent node.
 
 * Identify a leaf whose LeafNode is
   identical to the one in the KeyPackage.  If no such field exists, return an
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2393,8 +2393,8 @@ following criteria hold:
 * The nodes between U and P in the tree are all blank
 * The `parent_hash` field of U is equal to the parent hash of P with copath
   child S, where S is the child of P that is not on the path from U to P
-* The sibling of every node between U (included) and P (excluded) with a blank
-  parent has an empty resolution
+* The sibling of every node with a blank parent between U (included) and P
+  (excluded) has an empty resolution
 * `U.unmerged_leaves` is equal to the intersection between `P.unmerged_leaves`
   and the leaves under U.
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2392,7 +2392,11 @@ following criteria hold:
 * U is a descendant of P in the tree
 * The nodes between U and P in the tree are all blank
 * The `parent_hash` field of U is equal to the parent hash of P with copath
-  child S, where S is the child of P that is not on the path from U to P.
+  child S, where S is the child of P that is not on the path from U to P
+* The sibling of every node between U (included) and P (excluded) with a blank
+  parent has an empty resolution
+* `U.unmerged_leaves` is equal to the intersection between `P.unmerged_leaves`
+  and the leaves under U.
 
 A parent node P is "parent-hash valid" if it can be chained back to a leaf node
 in this way.  That is, if there is leaf node L and a sequence of parent nodes
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2393,11 +2393,11 @@ itself) and S the other child:
 
 * U is a descendant of P in the tree.
 
-* The resolution of C is equal to U added to the intersection of P's
-  `unmerged_leaves` with the subtree under C.
-
 * The `parent_hash` field of U is equal to the parent hash of P with copath
   child S.
+
+* U is in the resolution of C, and the intersection of P's `unmerged_leaves`
+  with the subtree under C is equal to the resolution of C with U removed.
 
 These checks verify that U and P were updated at the same time (in the same
 UpdatePath), and that they were neighbors in the UpdatePath because the nodes in


draft-ietf-mls-protocol.md
```diff
@@ -2392,7 +2392,11 @@ following criteria hold:
 * U is a descendant of P in the tree
 * The nodes between U and P in the tree are all blank
 * The `parent_hash` field of U is equal to the parent hash of P with copath
-  child S, where S is the child of P that is not on the path from U to P.
+  child S, where S is the child of P that is not on the path from U to P
+* The sibling of every node with a blank parent between U (included) and P
+  (excluded) has an empty resolution
```
bifurcation, line 2397: I would merge this with the bullet above about the intermediate nodes being blanks.  So something like:

> Each node between U and P in the tree is blank, and the "copath child" of P (its child that is not in the direct path of U) has an empty resolution

bifurcation (approved):
Thanks for the thorough write-up @TWal.  This makes sense to me, especially given that it's a pretty small change.

mulmarta:
I haven’t had tome to look at the proofs, but I have questions:

Re. Example 1 (and the first new check) what about the following scenario: We start with this tree. 

~~~ aasvg
                ________________R
               /                |
              W                 |
        ______|_____            |
       /             \          |
      _               Y         |
    __|__           __|__       |
   /     \         /     \      |
  T       _       X       Z     |
 / \     / \     / \     / \    |
A   B   _   _   E   F   G   H   I
~~~

Then `A` commits, setting the A-T-W path then `I` commits adding `C` and we have

~~~ aasvg
                ________________R
               /                |
              W[C]              |
        ______|_____            |
       /             \          |
      _               Y         |
    __|__           __|__       |
   /     \         /     \      |
  T       _       X       Z     |
 / \     / \     / \     / \    |
A   B   C   _   E   F   G   H   I
~~~

According to your rules, verification of the parent hash of W no longer works, because the blank sibling of T doesn’t have an empty resolution — it contains C.


Re. Example 2 (and the second new check), consider analogous scenario: We start with this tree:

~~~ aasvg               
              W                 
        ______|_____            
       /             \          
      U               Y         
    __|__           __|__       
   /     \         /     \      
  T       V       X       Z     
 / \     / \     / \     / \   
A   B   _   D   E   F   G   H
~~~

Then `A` commits setting the A-T-U path and then `G` commits adding `C` and we get

~~~ aasv
              W              
        ______|_____            
       /             \          
      U[C]            Y         
    __|__           __|__       
   /     \         /     \      
  T       V[C]    X       Z     
 / \     / \     / \     / \    
A   B   C   D   E   F   G   H   
~~~

The result look the same as the “bad case” in Example 2. On the other hand, the second new check still passes. So why do you think it prevents the example?

mulmarta:
I think that the quick fix for rule 1 may work. Rule 2 shouldn't hurt correctness either. Nevertheless, I'd feel more confident after seeing a thorough proof of correctness.

Re. example 2, you're right and thanks for explaining!

TWal:
Here is a quick proof:

Given a parent-hash link C ~> P, we write Trees(C ~> P) the list of subtrees of siblings of nodes whose parents are blank between C and P.

When a parent-hash link C ~> P is created, it because there is a Commit going through C and P, the blank nodes between C and P are here because of the filtered direct path, so every tree in Trees(C ~> P) have empty resolution, hence they are included in P.unmerged_leaves.

Given a parent-hash link C ~> P, let's see what happens if a tree Ti of Trees(C ~> P) is modified.
- If the operation is a Remove, then P is blanked and the parent-hash link C ~> P don't exist in the new tree
- If the operation is an Add, and the new leaf is L, then we have (before the operation) Resolution(Ti) included in P.unmerged_leaves, so Resolution(Ti) + L included in P.unmerged_leaves + L, and Resolution(NewTi) included in Resolution(Ti) + L, and NewP.unmerged_leaves = P.unmerged_leaves + L. So Resolution(NewTi) is included in NewP.unmerged_leaves.
- If the operation is a UpdatePath, then P has a new parent-hash link.

In the case analysis I consider that Adds are always add-only commit, because an Add in a full commit (in this part of the protocol) is equivalent to an add-only Add followed by an UpdatePath.

This proof isn't super well-written (did it quickly) but I hope it's convincing enough.

bifurcation:
I would note that one of the trees y'all are talking about is impossible:

```
              W                 
        ______|_____            
       /             \          
      U               Y         
    __|__           __|__       
   /     \         /     \      
  T       V       X       Z     
 / \     / \     / \     / \   
A   B   _   D   E   F   G   H
```

V would not be generated, because it's redundant.  I don't think that really changes the analysis, though.

bifurcation:
I wonder if a convenient way to summarize the property would be as follows.  Suppose there is the following relationship between U and P, where (a) the links might be either left or right, and (b) C is either U or a blank node:


```
            P
          __|__
         /     \
        C      ...
       _|_
      /   \
 (blanks) ...
    /
   U
  / \
... ...


intersect(P.unmerged_leaves, subtree(C)) = resolution(C) - D
```

That seems like it captures the two properties without the issues @mulmarta notes.

* A leaf that is unmerged at P is either unmerged at U or not within U's subtree
* The only non-blank nodes outside of U's subtree are leaves that are unmerged at P (and thus were not present when U was last updated)

It also has a nice symmetry in that both children of P are verified -- the copath child is verified by the original sibling tree hash, and the direct path child is verified with the unmerged leaves calculation.

Suggested text:

```
The parent hash in a node U is valid with respect to a parent node P if the
following criteria hold.  Here C and S are the children of P (for "child" and
"sibling"), with C being the child that is on the direct path of U (possibly U
itself) and S the other child:

* U is a descendant of P in the tree.

* Any nodes between U and P in the tree are blank.

* The intersection of P's `unmerged_leaves` with the subtree under C is equal to
  the resolution of C with U removed.

* The `parent_hash` field of U is equal to the parent hash of P with copath
  child S.

These checks verify that U and P were updated at the same time (in the same
UpdatePath), and that they were neighbors in the UpdatePath because the nodes in
between them would have omitted from the filtered direct path.
```

TWal:
Hmm, that looks plausible yes! I have to think more about it.

I started doing formal proofs in F* that this new version of parent hash is correct, and found an interesting weird scenario.
It exists because we don't enforce unmerged leaves to be under the subtree in which they are listed.

Suppose we have the following (impossible) tree:

```
              W
        ______|_____
       /             \
      U               Y
    __|__           __|__
   /     \         /     \
  T       V[0]    X       Z
 / \     / \     / \     / \
_   1   2   3   4   5   6   7
```
Last commit is from 1, so we have the parent-hash chain 1 - T - U - W.
In the parent-hash link T ~> U, the sibling is:
```
  V[0]
 / \  
2   3 
```

7 adds 0:
```
              W
        ______|_____
       /             \
      U[0]            Y
    __|__           __|__
   /     \         /     \
  T       V[0]    X       Z
 / \     / \     / \     / \
0   1   2   3   4   5   6   7
```

In the parent-hash link T ~> U, the sibling is:
```
  V
 / \  
2   3 
```

It is different, and the link is not valid anymore!
I think this should be an invariant checked when joining the group?

TWal:
Your property implies that all nodes between U and P are blank.
It implies that U is in the resolution of C and all nodes except U in the resolution in C are leaves.
If U is a parent node, we directly deduce that all nodes between U and P are blank
If U is a leaf node, then it isn't there because it's an unmerged leaf of some parent node in the resolution of C, because the resolution of C only contains leaves, so we have the same conclusion.

Your property implies my first invariant, i.e. "the sibling of every node with a blank parent between U (included) and P (excluded) has a resolution included in P.unmerged_leaves"
The tree looks like this:
```
          P
         / \
      C=_   ...
       / \
     ...  Sn
     /
    _
   / \
  _   S2
 / \
U   S1
```
We have resolution(C) - U = resolution(S1) + ... + resolution(Sn).
Since resolution(C) - U is included in P.unmerged_leaves, then we get my first invariant.

Your property implies my second invariant.
Given the equality
```
intersect(P.unmerged_leaves, subtree(C)) = resolution(C) - U
```
we apply the intersection with subtree(U) on the equality:
```
intersect(intersect(P.unmerged_leaves, subtree(C)), subtree(U)) = intersect(resolution(C) - U, subtree(U))
```
the left hand side is equal to intersect(P.unmerged_leaves, subtree(U)) by the inclusion subtree(U) it subtree(C)
and the right hand side is equal to U.unmerged_leaves, since resolution(C) = U + U.unmerged_leaves + resolution(S1) + ... + resolution(Sn)
therefore we get:
```
intersect(P.unmerged_leaves, subtree(U)) = U.unmerged_leaves
```
which is my second invariant.

So I think it works!
We just need to be sure that it's actually an invariant preserved in MLS.
Also I think we need to explicitly say that U is in the resolution of C.

TWal:
Another quick analysis, this time about correction:

If an Add of leaf L is done below U:
We have
```
intersect(oldP.unmerged_leaves, subtree(oldC)) = resolution(oldC) - U
```
We also have these equalities:
```
newP.unmerged_leaves = oldP.unmerged_leaves + L
subtree(newC) = subtree(oldC) + L
resolution(newC) = resolution(oldC) + L
```
from which we deduce:
```
intersect(newP.unmerged_leaves, subtree(newC)) = resolution(newC) - U
```

If an Add of leaf L is done below P, not below U:
We have
```
intersect(oldP.unmerged_leaves, subtree(oldC)) = resolution(oldC) - U
```
We also have these equalities:
```
newP.unmerged_leaves = oldP.unmerged_leaves + L
subtree(newC) = subtree(oldC) + L
resolution(newC) = resolution(oldC) + L
```
from which we deduce:
```
intersect(newP.unmerged_leaves, subtree(newC)) = resolution(newC) - U
```

It looks like a copy-paste of the previous case, in fact it is!
The only proof change is in the reason why `resolution(newC) = resolution(oldC) + L`.
In the first case, it's because it's added to the unmerged_leaves of U.
In the second case, it's because the recursion in the resolution definition goes all the way down to L.

The equality is true when creating the U ~> P link: both hand-side are empty.

TWal:
I think the following equation is better:
```
resolution(C) = U + intersect(P.unmerged_leaves, subtree(C))
```
It is equivalent to the previous one if U is in resolution(C), while also enforcing this fact.
I guess it also feels less "ninja" and it is easier to have intuition on.

TWal:
I started doing formal proofs with the equation I gave in the last message, it's actually quite nice to use for proofs!

I found another weird scenario, where an unmerged leaf is actually blank:

```
              W
        ______|_____
       /             \
      U               Y[4]
    __|__           __|__
   /     \         /     \
  T       V       _       Z
 / \     / \     / \     / \
0   1   2   3   _   5   6   7
```

Last update is Commit is from 0. We have the U ~> W parent-hash-link which contains the sibling

```
      Y[4]
    __|__
   /     \
  _       Z
 / \     / \
_   5   6   7
```

0 add 4, without UpdatePath:

```
              W[4]
        ______|_____
       /             \
      U               Y[4]
    __|__           __|__
   /     \         /     \
  T       V       _       Z
 / \     / \     / \     / \
0   1   2   3   4   5   6   7
```

Now the parent-hash computation for the U ~> W link uses this tree:

```
      Y
    __|__
   /     \
  _       Z
 / \     / \
_   5   6   7
```

Therefore the link is broken

bifurcation:
> It exists because we don't enforce unmerged leaves to be under the subtree in which they are listed. ...
> I think this should be an invariant checked when joining the group?

Yes, we should add this requirement.  Since we're going to be updating this PR anyway, I think we can just add it here.

Seems like it should go under "Verify the integrity of the ratchet tree." in the "Joining via ..." sections.  BTW, while I was looking at that, I notice that the parent hash description there still refers to children, so that needs to be fixed as well.

bifurcation:
> an unmerged leaf is actually blank

This seems like something we could also validate easily, along with the above.  "For each non-blank parent node, verify that each entry in the node's `unmerged_leaves` represents a non-blank leaf node that is a descendant of the parent node."

TWal:
With the checks in latest commit on this PR, I have a formal proof in F* that these checks are actually a invariants of the following operations:
- extending and truncating the tree
- adding a leaf in an empty slot, marking it unmerged up to the root
- removing a leaf, blanking every node up to the root
- apply an UpdatePath

In the next days I'll also prove in F* that the checks actually implies the property I mentioned in the first message, to be sure that we don't do all of this for nothing :)

I dropped the sentence "The nodes between U and P in the tree are all blank" because it's implied by the equation on "resolution(C)".
I thought about dropping the sentence about "U is a descendant of P in the tree" because it's also implied by the equation on "resolution(C)", but the problem is that "C" is defined with "U" so that's not possible…

TWal:
While doing the formal security proof, I found a problem:

Consider the following case.

```
      P1
     /
 C1=_
   / \
  U1  _
```

with the U1 ~> P1 parent hash link,

and the same tree, but U is unmerged

```
      P2[U2]
     /
 C2=_
   / \
  U2  _
```

(with the U2 ~> P2 link)

We have U1 ~= U2, however P1 ~= P2 is not true.

Two possible (equivalent) fixes:
- Use the equation `Resolution(C) = U + Intersection(P.unmerged_leaves, SubTree(C))` and furthermore say that the `+` is disjoint (or, equivalently, if U is a leaf then it is not in P.unmerged_leaves)
- Use the equation `Intersection(P.unmerged_leaves, SubTree(C)) = Resolution(C) - U`, but we have to say that `U` is in `Resolution(C)`

Which one has the nicest formulation to you?

bifurcation:
Interim 2022-06-09:

* @TWal reviewed the change and his proofs of correctness and security
* @mulmarta noted that she had implemented it and it passed fuzz testing (~300 commits with random operations)
* Agreement to merge

PR #715, opened 2022-06-01
Clarify tree hash and parent hash

### LINKED ISSUE #655: Tree hash and Parent hash definitions are confusing
Regarding the Tree Hash
1. Including the parent_node in the tree hash which is supposed to be computed from the leaves, makes the definition appear circular to a new reader. Since the situation when you have blank nodes in the middle of a tree but not at the root is a rather complicated, the situation when you could calculate a tree hash using the parent_node is non-obvious.
2. In the description of Tree Hash the document says that the ParentNode object is optional "depending on whether the node is blank or not." This isn't clear specification language. Saying the parent_node object is present if the node is blank or present if the node is not blank would be clear. 
3. What if the parent node is also blank? What if the root is blank? 
4. I think an example with and without blank nodes would be appropriate.
5. Explaining the reason why the parent_node is included will help the reader establish a better mental model.

Regarding the Parent Hash
I still find this section super confusing. A few points
1. Using a tree hash (sibling_tree_hash) inside the parent hash is non-obvious. There should be some explanation about what property this is providing and that it is a tree hash and not a parent hash. (Honestly it was the third time I read the section that I realized that it was really the tree hash and not the parent hash of the sibling.) 
2. Using P, V, and S for Parent, target node, and sibling node is introduced here and never used anywhere else in the document. P for parent and S for sibling make sense. I can't tell what V is supposed to stand for, and it unfortunately could cause people to confuse the <V> vector notation used throughout, but shortly after introducing V for the target node. I propose using "N" for node or "X" instead.
3. There is one reference to the filtered direct path of U. Perhaps this is V, or a child of V?
4. The calculations are sufficiently complicated that I think this warrants a complete, non-trivial, worked example (possibly in an appendix). At some point we should consider providing test vectors in a future document, and this would be an ideal candidate.

bifurcation:
Closes #655 

Depends on #694 
Conflicts with #713 

The diff here is messy because it includes #694.  The biggest change is the extended example, for which you can just scroll to the bottom.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2244,7 +2244,8 @@ object, which may contain either a `LeafNodeHashInput` or a
 `ParentNodeHashInput` depending on the type of node. `LeafNodeHashInput` objects
 contain the `leaf_index` and the `LeafNode` (if any). `ParentNodeHashInput`
 objects contain the `ParentNode` (if any) and the tree hash of the node's left
-and right children.
+and right children.  For both parent and leaf nodes, the optional node value
+MUST be absent if the node is blank and present if the node contains a value.
 
 ~~~ tls
 enum {
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2422,22 +2422,22 @@ computation.
 
 ### Using Parent Hashes
 
-The Parent Hash of P appears in three types of structs. If V is itself a parent node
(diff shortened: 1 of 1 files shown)


bifurcation (commented):
Interim 2022-06-02:

* Hold for next week, should be easier to review after rebase.

rohan-wire (approved):
This is a huge improvement. While I still prefer "path hash" to "parent hash", the new language in this PR is clear and I think will be straightforward for new implementers to read.

PR #716, opened 2022-06-02
Add new new_proposal_member SenderType

rohan-wire:
Add new new_proposal_member SenderType and rename new_member to new_commit_member. 

This restores the External Proposal Add functionality which was present in the protocol for some years but clarifies how the authorization and signing takes place in this case.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1433,7 +1433,8 @@ enum {
     reserved(0),
     member(1),
     external(2),
-    new_member(3),
+    new_commit_member(3),
+    new_proposal_member(4),
     (255)
 } SenderType;
 
@@ -1444,8 +1445,10 @@ struct {
             uint32 leaf_index;
         case external:
             uint32 sender_index;
-        case new_member:
+        case new_commit_member:
             struct{};
+        case new_proposal_member:
+            KeyPackageID proposed_member;
     }
 } Sender;
 
@@ -1556,11 +1559,13 @@ struct {
     MLSMessageContent content;
     select (MLSMessageContentTBS.content.sender.sender_type) {
         case member:
-        case new_member:
+        case new_commit_member:
             GroupContext context;
-
         case external:
             struct{};
+        case new_proposal_member:
+            // The KeyPackage included in embedded Add proposal
+            KeyPackage key_package;
     }
 } MLSMessageContentTBS;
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1433,7 +1433,8 @@ enum {
     reserved(0),
     member(1),
     external(2),
-    new_member(3),
+    new_commit_member(3),
+    new_proposal_member(4),
```
bifurcation, line 1437: Nit: I would flip these around, `new_member_commit` and `new_member_proposal`.

draft-ietf-mls-protocol.md
```diff
@@ -1444,8 +1445,10 @@ struct {
             uint32 leaf_index;
         case external:
             uint32 sender_index;
-        case new_member:
+        case new_commit_member:
             struct{};
+        case new_proposal_member:
+            KeyPackageID proposed_member;
```
bifurcation, line 1451: This can just be `struct{}`, since because the KeyPackage you apply is the one in the Add proposal.

draft-ietf-mls-protocol.md
```diff
@@ -1556,11 +1559,13 @@ struct {
...
         case member:
-        case new_member:
+        case new_commit_member:
             GroupContext context;
-
         case external:
             struct{};
+        case new_proposal_member:
+            // The KeyPackage included in embedded Add proposal
+            KeyPackage key_package;
```
bifurcation, line 1568: This doesn't seem all that useful, since the KeyPackage is already being signed via the Add proposal.  I would just make this `struct{}`.

bifurcation:
Interim 2022-06-02:

* Could punt to extensions, but this is partly just because we moved a little too fast on the removal
* No objections to reverting 
* Once conflicts and comments resolved, good to merge

bifurcation:
Closes #693

draft-ietf-mls-protocol.md
```diff
@@ -1596,9 +1598,12 @@ depending on the sender's `sender_type`:
   indicated by `sender_index` in the `external_senders` group context
   extension (see {{external-senders-extension}}). The
   `content_type` of the message MUST be `proposal`.
-* `new_member`: The signature key in the LeafNode in
+* `new_commit_member`: The signature key in the LeafNode in
     the Commit's path (see {{joining-via-external-commits}}). The
     `content_type` of the message MUST be `commit`.
+* `new_proposal_member`: The signature key in the LeafNode in
+    the KeyPackage embedded in an External Add Proposal. The
+    `content_type` of the message MUST be `proposal`.
```
bifurcation, line 1606:
```suggestion
* `new_proposal_member`: The signature key in the LeafNode in
    the KeyPackage embedded in an External Add Proposal. The
    `content_type` of the message MUST be `proposal` and the 
    `proposal_type` of the Proposal MUST be `add`.
```

bifurcation (approved):
LGTM, though would still prefer `new_member_proposal` etc.  One minor suggestion to lock down the content.

PR #718, opened 2022-06-03
Fix resolution example

bifurcation:
The current resolution example tree can't be created by the protocol.  If C were not blank, then D would not populate Z in its filtered direct path.  (Note that this implies that there are never unmerged leaves at the first level of the tree.)  This PR expands the example tree so that it can have unmerged leaves while still having a blank top node.  This makes the example a little more complicated (you have to follow the recursion down two levels), but it doesn't seem too bad.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1032,25 +1032,27 @@ represents a blank node and unmerged leaves are indicated in square
 brackets:
 
 ~~~ ascii-art
-       ...
-       /
-      _
-      |
-    .-+-.
-   /     \
-  _       Z[C]
- / \     / \
-A   _   C   D
+               ...
+               /
+              _
+        ______|______
+       /             \
+      X[C]            _
+    __|__           __|__
+   /     \         /     \
+  _       _       Y       _
+ / \     / \     / \     / \
+A   _   C   D   E   F   _   H
 
-0   1   2   3
+0   1   2   3   4   5   6   7
 ~~~
 {: #resolution-tree title="A tree with blanks and unmerged leaves" }
 
 In this tree, we can see all of the above rules in play:
 
-* The resolution of node Z is the list \[Z, C\]
-* The resolution of leaf 1 is the empty list \[\]
-* The resolution of top node is the list \[A, Z, C\]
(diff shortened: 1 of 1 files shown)


bifurcation:
cc @mulmarta

draft-ietf-mls-protocol.md
```diff
@@ -5129,11 +5131,12 @@ To construct the tree in {{full-tree}}:
...
 To construct the tree in {{resolution-tree}}:
 
-* A creates a group with B, C, D, as well as some members outside this subtree
-* D removes C, setting Z and the top node (as well as any further nodes in the
-  direct path)
-* A member outside this subtree removes B, blanking B's direct path
-* A adds a new member at C with a partial Commit, adding it as unmerged at Z
+* A creates a group with B, ..., H, as well as some members outside this subtree
+* F sends an empty Commit, setting Y and its ancestors
+* D removes B, C, and F, with the following effects:
```
mulmarta, line 5136:
```suggestion
* D removes B, C, and G, with the following effects:
```

draft-ietf-mls-protocol.md
```diff
@@ -1032,25 +1032,27 @@ represents a blank node and unmerged leaves are indicated in square
...
-      _
-      |
-    .-+-.
-   /     \
-  _       Z[C]
- / \     / \
-A   _   C   D
+               ...
+               /
+              _
```
mulmarta, line 1037: Shouldn't this node be set to `Z[C]`? I think D will set it as "the top node"

mulmarta:
Good catch! It's kind of subtle in this case.

draft-ietf-mls-protocol.md
bifurcation: Ah, yes, as it is now.  I need to have someone outside the subtree remove G in order to clear it.  I'll update the description.

bifurcation:
@mulmarta - I think I have this patched up, would appreciate a quick re-review.

rohan-wire (approved):
I walked through the example tree on a whiteboard with different colored markers and it looks correct to me.

matedura300-spec (commented):
Exelente

PR #723, opened 2022-06-10
Add `external_senders` to MLS Extension Types registry

tomleavy:
`external_senders` was missing from the Extension Types table so it didn't have an extension type value assigned


rohan-wire (approved):
Thanks

bifurcation (approved):

PR #724, opened 2022-06-11
Fix #719 (contributors section)

### LINKED ISSUE #719: Document compilation drop accents
Hello,

In the ["contributors" section](https://github.com/mlswg/mls-protocol/blob/main/draft-ietf-mls-protocol.md?plain=1#L5101), my name is correctly written with the "é" accent, however when it is compiled, the accent is dropped, e.g. [here](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#section-18-1.14) or [here](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.txt).

It looks like this letter is the only non-ascii character in the whole document, is this the expected behavior when compiling an RFC, or is it an issue in [martinthomson/i-d-template](https://github.com/martinthomson/i-d-template) that I should track?

Thanks!

cabo:
To check whether there actually is a problem, I made the fix for #719, and it seems to work as designed.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -71,8 +71,7 @@ contributor:
    org:  Amazon
    email:  tomleavy@amazon.com
  - name: Brendan McMillion
-   org:  brendanmcmillion@gmail.
-   email:
+   email: brendanmcmillion@gmail.com
  - name: Marta Mularczyk
    org:  Amazon
    email:  mulmarta@amazon.com
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -42,6 +42,53 @@ author:
     organization: University of Oxford
     email: me@katriel.co.uk
 
+contributor:
+ - name: Joel Alwen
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -42,6 +42,53 @@ author:
...
+   email:  konrad.kohbrok@datashrine.de
+ - name: Albert Kwon
+   org:  MIT
+   email:  kwonal@mit.edu
+ - name: Tom Leavy
+   org:  Amazon
+   email:  tomleavy@amazon.com
+ - name: Brendan McMillion
+   org:  brendanmcmillion@gmail.
+   email:
```
Bren2010, line 75:
```suggestion
   email: brendanmcmillion@gmail.com
```

bifurcation:
Thanks @cabo, I didn't know about this feature!  If you'd like to do the same on [the architecture doc](https://github.com/mlswg/mls-architecture/), I would appreciate it.

PR #706, opened 2022-05-29
Clean up struct names and references to them

bifurcation:
Closes #673 
Closes #674 

With regard to #674: I searched for references to MLSPlaintext and Credential that were obsolete in light of the change to the MLSMessage framework.  (In particular, we used to have the idea that you would encode things as MLSPlaintext, then tranform them to MLSCiphertext if needed; now you encode to AuthenticatedContent and then convert that to MLSPlaintext/MLSCiphertext before sending.)  There were a few fixes for MLSPlaintext; I didn't find anything that needed fixing for Credentia, largely thanks to some PRs that have landed since I did the review that resulted in #674.

PR #728, opened 2022-06-21
Fix GroupInfoTBS

TWal:
The `confirmation_tag` was missing. The error was introduced in #706.

This kind of error makes me wonder if the redundancy between the `...` and the `...TBS` really are necessary.

In my code, I use a pattern like this:
```
struct {
    GroupContext group_context;
    Extension extensions<V>;
    MAC confirmation_tag;
    uint32 signer;
} GroupInfoTBS;

struct {
    GroupInfoTBS tbs;
    // SignWithLabel(., "GroupInfoTBS", GroupInfoTBS)
    opaque signature<V>;
} GroupInfo;
```

No copy-paste error possible! Looks a bit weird, but if we replace "tbs" by "data", it makes more sense.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -4197,6 +4197,7 @@ GroupInfo above `signature`:
 struct {
     GroupContext group_context;
     Extension extensions<V>;
+    MAC confirmation_tag;
     uint32 signer;
 } GroupInfoTBS;
 ~~~


Bren2010 (approved):

MartaMularczyk:
> Thanks, @TWal. FWIW, I would be disinclined to make the change you propose at this point. It's a fine idea (it's what [X.509 does](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1)), but we're late enough in the process that I would like to avoid changes of that magnitude.

To me it seems that this doesn't introduce a new change, but instead reverts a change made by mistake a month ago. The tag has always been signed, also in the version of MLS we analyzed.

bifurcation:
@MartaMularczyk - Sorry, I meant the bigger change that @TWal proposed (e.g., putting GroupInfoTBS inside GroupInfo). Signing the tag is obviously correct.

PR #740, opened 2022-08-26
Add limit to ResumptionPSKUsage

franziskuskiefer:
- The `ResumptionPSKUsage` enum was missing the `u8` limit.
- The `GroupContextExtensions` wasn't using a TLS environment.


seanturner (approved):
Seems reasonable to me.

PR #747, opened 2022-09-23
Address TSV-ART review comments

bifurcation:
Thanks to Gorry Fairhurst:
https://mailarchive.ietf.org/arch/msg/mls/ju2j3jsdttOfC06OedFFvZN73l8/

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -4746,7 +4746,7 @@ chosen at the time of message encryption by the sender. Senders may use padding
 to reduce the ability of attackers outside the group to infer the size of the
 encrypted content.  Note, however, that the transports used to carry MLS
 messages may have maximum message sizes, so padding schemes SHOULD avoid
-increasing message size beyond these any such limits that exist in a given
+increasing message size beyond any such limits that exist in a given
 deployment scenario.
 
 ## Restrictions {#restrictions}
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -135,10 +135,6 @@ in size ranging from two to thousands.
 
 # Introduction
 
-DISCLAIMER: This is a work-in-progress draft of MLS and has not yet
-seen significant security analysis. It should not be used as a basis
-for building production systems.
-
 RFC EDITOR: PLEASE REMOVE THE FOLLOWING PARAGRAPH The source for
 this draft is maintained in GitHub. Suggested changes should be
 submitted as pull requests at https://github.com/mlswg/mls-protocol.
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -911,7 +907,9 @@ an attacker that had compromised the sender's prior leaf secret.
 Update messages SHOULD be sent at regular intervals of time as long as the group
 is active, and members that don't update SHOULD eventually be removed from the
 group. It's left to the application to determine an appropriate amount of time
-between Updates.
+between Updates.  In general, however, applications should take care that they
+do not send MLS messages at a rate that overwhelms the transport over which
+messages are being sent.
```
gorryfair, line 912: /transport/ or /network path/

draft-ietf-mls-protocol.md
```diff
@@ -4744,7 +4744,10 @@ to an adversary solely by the ciphertext length.
 The length of the `padding` field in `MLSCiphertextContent` can be
 chosen at the time of message encryption by the sender. Senders may use padding
 to reduce the ability of attackers outside the group to infer the size of the
-encrypted content.
+encrypted content.  Note, however, that the transports used to carry MLS
+messages may have maximum message sizes, so padding schemes SHOULD avoid
+increasing message size beyond these any such limits that exist in a given
```
gorryfair, line 4749: Personally, I think it is necessary that the the resulting IP packets do not exceed the actual PMTU. This will "black hole" the data, the network will drop these packets. So, saying "SHOULD" kind of looks like there might be a way around that exists - whereas really the way around is to make sure this does not happen.

 I wonder in this case where it might be better to say "needs to avoid sending IP packets that are larger than this size" - this is of course often handled by lower protocol layers, and you may wish to also note that

Bren2010, line 4749:
```suggestion
increasing message size beyond any such limits that exist in a given
```

draft-ietf-mls-protocol.md
bifurcation: Here we are using "transport" in a higher-level sense.  As @Bren2010 points out, the "transport" could be something like WebSockets or Tor cells.  It is possible in principle that you might use something like UDP or raw IP packets, but that would require a lot of invention.

raphaelrobert: I think the name "update message" is perhaps causing some confusion. They are not some kind of keepalive packets or high-frequency state synchronization messages. In many applications, the interval between two update messages can range from hours to months.
Their primary purpose is to limit the "window of compromise" by providing post-compromise security, I doubt applications will require the window to be < 1 min.

raphaelrobert (approved):

Bren2010 (approved):

PR #751, opened 2022-10-04
State update/commit epoch in updated LeafNode

### LINKED ISSUE #748: Include last updated epoch in leaf nodes?
Inspired by #731 -- it seems like when someone issues an Update or Commit, they know the epoch in which it will be committed, and could include that under the signature for their LeafNode.  Especially in the Commit case, this could improve validation of parent hashes, since the leaf node to which a parent node chains via the parent hash should be the leaf node in that parent node's subtree that most recently updated.  A PR would have roughly the following form:

* Update LeafNode to include the following (plus corresponding updates to `LeafNodeTBS`):

```diff
     LeafNodeSource leaf_node_source;
     select (LeafNode.leaf_node_source) {
         case key_package:
             Lifetime lifetime;
 
         case update:
-            struct{};
+            uint64 update_epoch;
 
         case commit:
+            uint64 commit_epoch;
             opaque parent_hash<V>;
     }
```

* In `{{leaf-node-validation}}`, add the following checks:
    * When verifying a Update or Commit message, check that `update_epoch` / `commit_epoch` is the current epoch
    * When verifying a tree, check that for all leaves, `update_epoch` / `commit_epoch` is before the current epoch if present

(Note that we cannot just add it on the signature side, as with #731, since a new joiner won't know the right values.)

The main arguments that come to mind against making this change are: (1) The above rational isn't super strong, in that it's just a check on the ordering that the parent hash already provides.  (2) This could be cleanly done in a LeafNode extension.

bifurcation:
Fixes #748

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2670,10 +2670,17 @@ UpdatePath), and that they were neighbors in the UpdatePath because the nodes in
 between them would have omitted from the filtered direct path.
 
 A parent node P is "parent-hash valid" if it can be chained back to a leaf node
-in this way.  That is, if there is leaf node L and a sequence of parent nodes
-P\_1, ..., P\_N such that P\_N = P and each step in the chain is authenticated
-by a parent hash: L's parent hash is valid with respect to P\_1, P\_1's parent
-hash is valid with respect to P\_2, and so on.
+in this way, and if that leaf node was the last updated leaf under that parent
+node.  That is, P is parent-hash valid if:
+
+1. There is leaf node L and a sequence of parent nodes P\_1, ..., P\_N such that
+   P\_N = P and each step in the chain is authenticated by a parent hash: L's
+   parent hash is valid with respect to P\_1, P\_1's parent hash is valid with
+   respect to P\_2, and so on.
+
+2. The leaf node L has `leaf_node_source` set to `commit` and its `commit_epoch`
+   field indicates a epoch greater than or equal to any `update_epoch` or
+   `commit_epoch` value present among the leaf nodes that are descendants of P.
 
 When joining a group, the new member MUST authenticate that each non-blank
 parent node P is parent-hash valid.  This can be done "bottom up" by building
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -2665,10 +2670,17 @@ UpdatePath), and that they were neighbors in the UpdatePath because the nodes in
...
 
 A parent node P is "parent-hash valid" if it can be chained back to a leaf node
-in this way.  That is, if there is leaf node L and a sequence of parent nodes
-P\_1, ..., P\_N such that P\_N = P and each step in the chain is authenticated
-by a parent hash: L's parent hash is valid with respect to P\_1, P\_1's parent
-hash is valid with respect to P\_2, and so on.
+in this way, and if that leaf node was the last updated leaf under that parent
+node.  That is, P is parent-hash valid if all of the following hold:
+
+1. There is leaf node L and a sequence of parent nodes P\_1, ..., P\_N such that
```
Bren2010, line 2676:
```suggestion
1. There is a leaf node L and a sequence of parent nodes P\_1, ..., P\_N such that
```

draft-ietf-mls-protocol.md
```diff
@@ -2665,10 +2670,17 @@ UpdatePath), and that they were neighbors in the UpdatePath because the nodes in
...
+   P\_N = P and each step in the chain is authenticated by a parent hash: L's
+   parent hash is valid with respect to P\_1, P\_1's parent hash is valid with
+   respect to P\_2, and so on.
+
+2. The leaf node L has `leaf_node_source` set to `commit` and its `commit_epoch`
+   field indicates a epoch greater than or equal to any `update_epoch` or
```
Bren2010, line 2682:
```suggestion
   field indicates an epoch greater than or equal to any `update_epoch` or
```

Bren2010 (approved):

psyoptix:
What is the motivation for this change?

I ask because I think there is a bit of a down side to this change. It amounts to putting more metadata (about the activity of group members) into the ratchet tree which, in many implementations, will end up in the clear on a server. So I'm a bit hesitant at the moment to mandate this as opposed to, say, making it an optional extension.

PR #754, opened 2022-10-20
Allow multiple welcomes per commit

### LINKED ISSUE #741: ResumptionPSKUsage presentation syntax
@bifurcation shouldn't ResumptionPSKUsage include (255)?

    enum {
      reserved(0),
      application(1),
      reinit(2),
      branch(3),
      (255)
    } ResumptionPSKUsage;

### LINKED ISSUE #753: allow sending a different Welcome message to each user
From my reading, it looks like the creator of an epoch must send the exact same `Welcome` message to each new user.  However, in some systems, it is more efficient to send a different message to each user, only containing the `EncryptedGroupSecrets` intended for them, rather than the `EncryptedGroupSecrets` for all new members.  As far as I can tell, no changes are required on the recipient side.

I'm running some performance testing, trying to add thousands of members at a time, and the `Welcome` messages are hundreds of kilobytes large.  In a federated environment, this could be copied to multiple servers, with the vast majority of that being unnecessary data.

bifurcation:
Fixes #753 

This initial effort tweaks the sections that talk about generating and processing Welcome messages.  There might be other references to a single Welcome per Commit that should be amended for clarity.  Someone should audit for those before we merge this PR.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -4070,10 +4070,14 @@ message at the same time, by taking the following steps:
   * Compute an EncryptedGroupSecrets object that encapsulates the `init_secret`
     for the current epoch and the path secret (if present).
 
-* Construct a Welcome message from the encrypted GroupInfo object, the encrypted
-  key packages, and any PSKs for which a proposal was included in the Commit. The
-  order of the `psks` MUST be the same as the order of PreSharedKey proposals in the
-  `proposals` vector.
+* Construct one or more Welcome messages from the encrypted GroupInfo object,
+  the encrypted key packages, and any PSKs for which a proposal was included in
+  the Commit. The order of the `psks` MUST be the same as the order of
+  PreSharedKey proposals in the `proposals` vector.  As discussed on
+  {{joining-via-welcome-message}}, the committer is free to choose how many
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -4070,10 +4070,14 @@ message at the same time, by taking the following steps:
...
-  order of the `psks` MUST be the same as the order of PreSharedKey proposals in the
-  `proposals` vector.
+* Construct one or more Welcome messages from the encrypted GroupInfo object,
+  the encrypted key packages, and any PSKs for which a proposal was included in
+  the Commit. The order of the `psks` MUST be the same as the order of
+  PreSharedKey proposals in the `proposals` vector.  As discussed on
+  {{joining-via-welcome-message}}, the committer is free to choose how many
+  Welcome messages to construct.  However, the set of Welcome messages produced
+  in this step MUST include one Welcome encrypted for each new member added in
+  the Commit.
```
TWal, line 4080:
```suggestion
  Welcome messages to construct.  However, the set of Welcome messages produced
  in this step MUST cover every new member added in the Commit.
```
I found this sentence quite hard to understand (basically it means that every new member must be able to decrypt one of the welcome messages?), I tried a reformulation

PR #755, opened 2022-10-27
BasicCredential public key must be same as in LeafNode

franziskuskiefer:
Unless I'm missing something there's currently no connection between the basic credential and the signature public key in the leaf node. A sentence similar to the x.509 option should be added.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1401,6 +1401,8 @@ struct {
 
 A BasicCredential is a bare assertion of an identity, without any additional
 information.  The format of the encoded identity is defined by the application.
+Public key material associated with the BasicCredential MUST be identical to the
+`signature_key` in the LeafNode containing this credential.
 
 For an X.509 credential, each entry in the chain represents a single DER-encoded
 X.509 certificate. The chain is ordered such that the first entry (chain[0]) is


bifurcation:
I don't think this is necessary.  There is no key material in a BasicCredential, just a (non-cryptographic) identity.  Note that this is a relatively recent change, dating to when we pulled the `signature_key` out of the credential into the LeafNode.  There used to be a signature public key in `BasicCredential`, but now there is not.  In this regard, X509Credential is actually the outlier, since it duplicatively lists the signature public key!

franziskuskiefer:
Well, the basic credential is an odd thing right now. Because it's entirely defined by the application at could be anything.
The only connection between the two right now is "The credential field contains information authenticating both the member's identity and the provided signing key" in 8.2. But the basic credential could be a key(pair) itself. How should it then relate to the signature key?
I don't think we need this exact sentence here. But there needs to be a little more clarity around what the "basic" credential is supposed to be now, esp. in relation to the signature key.

kkohbrok:
I don't have a strong opinion about the basic credential one way or another, but the credential (regardless of its type) really must contain either the signature key in the LeafNode or at least a commitment to it. Otherwise the AS-validation of the credential is pretty much meaningless, as it isn't at all connected to the LeafNode it is in. I guess another possibility is for the credential to contain a signature over the LeafNode, but that seems a bit backwards.

PR #756, opened 2022-10-27
Remove reference to BasicCredential.

Bren2010:
#755 made me realize that BasicCredential actually doesn't exist anymore!

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1399,7 +1399,7 @@ struct {
 } Credential;
 ~~~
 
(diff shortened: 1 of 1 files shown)


franziskuskiefer (approved):
👍🏻

rohan-wire:
> #755 made me realize that BasicCredential actually doesn't exist anymore!

Well, "basic" is listed as a credential type in the IANA registry defined in [Section 17.4](https://www.ietf.org/archive/id/draft-ietf-mls-protocol-16.html#mls-credential-types). That said, your proposed text seems perfectly consistent with the registry and the rest of the document.

PR #757, opened 2022-10-28
Clarify external join into first epoch

mulmarta:
I realized that the external joiner deals with the epoch 0 a bit differently. This clarifies the special case. I hope this was the intended behavior.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -4426,6 +4426,11 @@ has to meet a specific set of requirements:
   init secret as described in {{external-initialization}}.
 * The sender type for the MLSAuthenticatedContent encapsulating the External Commit MUST be
   `new_member_commit`.
+* Normally, the joiner computes the interim transcript hash using the confirmed transcript
+  hash and the confirmation tag from the GroupInfo. The only exception is when the GroupInfo
+  was sent in an epoch with epoch ID 0 (i.e., before any members were added); in this case,
+  the confirmation tag in GroupInfo can be arbitrary and the joiner uses the empty interim
+  transcript hash.
 
 External Commits come in two "flavors" -- a "join" commit that
 adds the sender to the group or a "resync" commit that replaces a member's prior


draft-ietf-mls-protocol.md
```diff
@@ -4426,6 +4426,11 @@ has to meet a specific set of requirements:
   init secret as described in {{external-initialization}}.
 * The sender type for the MLSAuthenticatedContent encapsulating the External Commit MUST be
   `new_member_commit`.
+* Normally, the joiner computes the interim transcript hash using the confirmed transcript
+  hash and the confirmation tag from the GroupInfo. The only exception is when the GroupInfo
+  was sent in an epoch with epoch ID 0 (i.e., before any members were added); in this case,
```
bifurcation, line 4431: > epoch ID 0 (i.e., before any members were added)

This doesn't make sense.  In epoch 0, there is a member, the creator.

draft-ietf-mls-protocol.md
```diff
@@ -4426,6 +4426,11 @@ has to meet a specific set of requirements:
...
+  the confirmation tag in GroupInfo can be arbitrary and the joiner uses the empty interim
+  transcript hash.
```
bifurcation, line 4433: I think this is correct given the way group creation is specified right now.

Another way we could fix this is to make the 0->1 transition a little more like other transitions, in the following sense:

* Have the group creator choose an epoch secret (not just an init secret)
* Set the initial interim transcript hash to the MAC computed with the empty confirmed transcript hash and the confirmation key from the initial epoch secret
* Then the external joiner computes the interim transcript hash as normal.

I have a slight preference for doing this latter fix.  It makes things more consistent, but would be a slightly larger change.

bifurcation:
Interim 2012-12-08:
* Epoch 0 GroupInfo has empty confirmation tag
* Initial Interim transcript hash = confirmed + empty confirmation tag
* OK to merge with those changes 
* RLB to file an issue to update Section 12 to clarify that you can have one-member groups

PR #821, opened 2022-12-06
Make ProtocolVersion two bytes

### LINKED ISSUE #766: ENUMs limited to 1 octet?
_Comment by @paulwouters_

~~~ 
 enum {
       reserved(0),
       mls10(1),
       (255)
   } ProtocolVersion;
~~~
Would it really do harm to turn this into two octets? I don't think message size is a real concern here. Same for ContentType and SenderType. Note: thank you for skipping 0 as a valid entry!! Implementers applaud you :)

bifurcation:
Fixes #766

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1506,7 +1506,7 @@ Delivery Service to examine such messages.
 enum {
     reserved(0),
     mls10(1),
-    (255)
+    (65535)
 } ProtocolVersion;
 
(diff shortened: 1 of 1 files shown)


rohan-wire:
I think this is unnecessary and capricious.

bifurcation:
Interim 2022-12-08:
* Make ProtocolVersion and WireFormat two bytes
* Leave other enums as-is
* Define an IANA registry for WireFormat

PR #823, opened 2022-12-07
Minor edits for AD comments

### LINKED ISSUE #780: Prospective member or member ?
_Comment by @paulwouters_

~~~ 
   It also generates a Welcome message and sends this directly to the
  new member
~~~
Should this be "prospective member" or "invitee" ? They haven't joined yet, have they?

### LINKED ISSUE #781: "server" is ambiguous here
_Comment by @paulwouters_

~~~ 
   Only after A
   has received its Commit message back from the server
~~~
What server? The Delivery Service ?

### LINKED ISSUE #797: May implement correctly :-)
_Comment by @paulwouters_

~~~ 
   An implementation MAY use
   any tree representation and associated algorithms, as long as they
   produce correct protocol messages.
~~~
This MAY is super weird. Is it allowed to implement in such a way that they to NOT produce correct protocol messages? :-)

Remove the entire sentence or lowercase the MAY?

### LINKED ISSUE #798: 7 or 8 members
_Comment by @paulwouters_

Figure 10 claims to have seven members, but I count 8, namely Y,T,X,A,B,E,F and G  (not counting W, the root)

### LINKED ISSUE #799: Setion 8.7 MAY
_Comment by @paulwouters_

~~~ 
   An implementation MAY use any
   algorithm that produces the correct tree in its internal
   representation.
~~~
Another weird MAY here. Again, I would just lowercase it as this is just a pointer to a human decision, not a 2119 based protocol enforcement.

### LINKED ISSUE #801: Section 10.1 encryption size
_Comment by @paulwouters_

~~~ 
Each key/nonce pair MUST NOT be used to encrypt more than one message.
~~~
Isn't there technically also limits to how much one AEAD key and one nonce can be used to encrypt a (very very) large message? I guess this could not happen, but is there some formal message size limit in MLS ? (looking at the vectors at the start, limits could be higher than 2^64 ?)

### LINKED ISSUE #813: Discuss metadata privacy
_Comment by @paulwouters_

That is quite some information that must be publicly available to allow external Add requests. This leaks information, eg same epoch means the group membership did not change, etc etc. I hope this will appear in the Security Considerations later :)

### LINKED ISSUE #814: Section 15.3
_Comment by @paulwouters_

~~~ 
This is in addition to ensuring that these nonce and key pairs are promptly
   deleted when the epoch ends.
~~~
This sentence with a hidden MUST is stuck between a sentence with MAY and a sentence with SHOULD. Perhaps make the MUST explicit ?

### LINKED ISSUE #817: Section 17.1
_Comment by @paulwouters_

~~~ 
   Additionally clients that run predominantly on mobile
   processors can choose ChaCha20Poly1305 over AES-GCM for performance
   reasons.
~~~
Citation needed? I think most phones have AES(_GCM) in hardware too ? Also compared to HPKE operations and the times for epoch/group reinit, does this performance really matter? (eg waking up from sleep to increment the epoch would make the symmetric encryption cost pretty irrelevant. I would remove this sentence.

bifurcation:
Fixes #780 
Fixes #781
Fixes #797
Fixes #798
Fixes #799
Fixes #801
Fixes #814
Fixes #817

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -857,9 +857,12 @@ directly to the new member (there's no need to send it to the group). Only after
 A has received its Commit message back from the group channel does it update its
 state to reflect the new member's addition.
 
-Upon receiving the Welcome message, the new member will be able to read and send
-new messages to the group. However, messages sent before they were added to the
-group will not be accessible.
+Once A has updated its state, the new member has processed the Welcome, and any
+other group members have processed the Commit, they will all have consistent
+representations of the group state, including a group secret that is known only
+to the members the group. The new member will be able to read and send new
+messages to the group, but messages sent before they were added to the group
+will not be accessible.
 
 ~~~ aasvg
                                                                Group
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -854,7 +854,7 @@ group state containing only itself and downloads KeyPackages for B and C. For
 each member, A generates an Add and Commit message adding that member, and
 broadcasts them to the group. It also generates a Welcome message and sends this
 directly to the new member (there's no need to send it to the group). Only after
-A has received its Commit message back from the group channel does it update its
+A has received its Commit message back from the Delivery Service does it update its
 state to reflect the new member's addition.
 
 Once A has updated its state, the new member has processed the Welcome, and any
@@ -4786,7 +4786,7 @@ were delayed or reordered.
 Applications SHOULD define a policy on how long to keep unused nonce and key
 pairs for a sender, and the maximum number to keep. This is in addition to
 ensuring that these secrets are deleted according to the deletion schedule
-defined in {{deletion-schedule}} Applications SHOULD also define a policy
+defined in {{deletion-schedule}}. Applications SHOULD also define a policy
 limiting the maximum number of steps that clients will move the ratchet forward
 in response to a new message.  Messages received with a generation counter
 that's too much higher than the last message received would then be rejected.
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -854,12 +854,15 @@ group state containing only itself and downloads KeyPackages for B and C. For
 each member, A generates an Add and Commit message adding that member, and
 broadcasts them to the group. It also generates a Welcome message and sends this
 directly to the new member (there's no need to send it to the group). Only after
-A has received its Commit message back from the server does it update its state
-to reflect the new member's addition.
+A has received its Commit message back from the group channel does it update its
```
Bren2010, line 857:
```suggestion
A has received its Commit message back from the Delivery Service does it update its
```

draft-ietf-mls-protocol.md
```diff
@@ -4777,15 +4785,15 @@ were delayed or reordered.
...
-ends. Applications SHOULD also define a policy limiting the maximum number of
-steps that clients will move the ratchet forward in response to a new message.
-Messages received with a generation counter that's too much higher than the last
-message received would then be rejected. This avoids causing a denial-of-service
-attack by requiring the recipient to perform an excessive number of key
-derivations. For example, a malicious group member could send a message with
-`generation = 0xffffffff` at the beginning of a new epoch, forcing recipients to
-perform billions of key derivations.
+ensuring that these secrets are deleted according to the deletion schedule
+defined in {{deletion-schedule}} Applications SHOULD also define a policy
```
Bren2010, line 4789:
```suggestion
defined in {{deletion-schedule}}. Applications SHOULD also define a policy
```

Bren2010 (approved):

rohan-wire (approved):

PR #824, opened 2022-12-07
Credentials should be replaced before expiring

### LINKED ISSUE #758: Certificate expiration and joining groups
We ran into this case, which does not seem to have a defined solution within the RFC and I'm wondering if this ambiguity should be resolved at the RFC level, or be left up to the application.

Existing group with Alice, and Bob.

1. Bob had a valid certificate upon being added to the group by Alice.
2. Bob's certificate expires .
3. Alice does a commit to add Charlie to the group.
4. Bob accepts the valid commit to add Charlie.
5. Charlie follows 8.3 and validates the tree. In this process if certificate expiration is enforced, then Charlie would drop the Welcome, which would not be optimal.

Solution 1: Leave this behavior up to the application, consider it part of the AS. 

Solution 2: Add an explicit mention that credentials that expire should be considered valid after expiration once they are a member of the group.

Solution 2: A committer MUST validate that all existing credentials in the tree are still considered valid at the time a commit is made. If an existing credential is no longer considered valid, then the resulting commit MUST contain a Remove proposal for the leaf containing that credential.

bifurcation:
Fixes #758

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1447,6 +1447,26 @@ identifiers in the new credential is valid as a successor to the set of
 presented identifiers in the old credential, according to the application's
 policy.
 
+### Expiring Credentials
+
+In some credential schemes, the validity status of a credential can change over
+time.  For example, X.509 certificates can expire or be revoked by their issuer.
+To avoid operational problems such as new joiners rejecting expired credentials
+in a group, applications that use such credentials should ensure that all of the
+credentials in use in a group are valid.
+
+If a member finds that its credential is invalid (or will be soon), it should
+issue an Update or Commit that replaces it with a valid credential.  For this
+reason, members SHOULD accept Update proposals and Commits issued by members
+with invalid credentials, if the credential in the Update or Commit is valid.
+
+If a member finds that another member's credential is invalid, they may issue a
+Remove that removes that member.  For example, an application could require a
+member preparing to issue a Commit to check the tree for invalid credentials and
+include Remove proposals for those members in its Commit.  In situations where
+the group tree is known to the DS, the DS could also monitor the tree for
+invalid credentials and issue external Remove proposals.
+
 ### Uniquely Identifying Clients
 
 MLS implementations will presumably provide applications with a way to request


rohan-wire (approved):
Perfect.

bifurcation:
Interim 2022-12-08:
* Distinguish "time-invalid" from "invalid"
* There is a slightly more general issue: When catching up, there might be messages sent by members whose credentials were valid when the message was sent, but have since expired
* Recommend accepting messages from members with time-invalid credentials when catching up

draft-ietf-mls-protocol.md
```diff
@@ -1447,6 +1447,26 @@ identifiers in the new credential is valid as a successor to the set of
...
+In some credential schemes, the validity status of a credential can change over
+time.  For example, X.509 certificates can expire or be revoked by their issuer.
+To avoid operational problems such as new joiners rejecting expired credentials
+in a group, applications that use such credentials should ensure that all of the
+credentials in use in a group are valid.
+
+If a member finds that its credential is invalid (or will be soon), it should
+issue an Update or Commit that replaces it with a valid credential.  For this
+reason, members SHOULD accept Update proposals and Commits issued by members
+with invalid credentials, if the credential in the Update or Commit is valid.
```
rohan-wire, line 1461: s/with invalid credentials/with recently expired but otherwise valid credentials/

tomleavy (approved):
Looks great thanks @bifurcation

draft-ietf-mls-protocol.md
```diff
@@ -1447,6 +1447,42 @@ identifiers in the new credential is valid as a successor to the set of
...
+requirements of {{credential-validation}}.  Applications can apply some
+operational practices and adaptations to Authentication Service policies to
+moderate these impacts.
+
+It can sometimes be important to distinguish this type of invalidity from other
+types; we call a credential "time-invalid" that is invalid for some time-based
+reason but otherwise valid.  An expired certificate issued by a trusted
+certificate authority would be time-invalid.
+
+In general, To avoid operational problems such as new joiners rejecting expired
```
kkohbrok, line 1464:
```suggestion
In general, to avoid operational problems such as new joiners rejecting expired
```

kkohbrok (approved):
Looks good with one small nit. Should we also note that the application might want to ensure that however the SHOULDs are interpreted, validation should be consistent for all clients? Or is that too much complexity?

draft-ietf-mls-protocol.md
```diff
@@ -1447,6 +1447,42 @@ identifiers in the new credential is valid as a successor to the set of
...
+certificate authority would be time-invalid.
+
+In general, to avoid operational problems such as new joiners rejecting expired
+credentials in a group, applications that use such credentials should ensure
+that all of the credentials in use in a group are valid.
+
+If a member finds that its credential is invalid (or will be soon), it should
+issue an Update or Commit that replaces it with a valid credential.  For this
+reason, members SHOULD accept Update proposals and Commits issued by members
+with time-invalid credentials, if the credential in the Update or Commit is
```
rohan-wire, line 1471: I still find "time-invalid" to be awkward.  I think "recently expired" is clearer, narrower, and a better fit. There is no reason for example to accept an Update with a credential which is not yet valid (which sounds time-invalid to me).

rohan-wire (changes requested):

draft-ietf-mls-protocol.md
bifurcation: I had shied away from "recently expired" because a credential can also become invalid due to revocation.  I guess we could write "expired or revoked" (since you usually can't tell "recently revoked").

rohan-wire: I would argue that you do NOT want to accept revoked credentials. That client is dead to me.  (talk to the ✋)

kkohbrok (approved):
I like the separate treatment of expired vs. revoked. Looks good to me now!

PR #827, opened 2022-12-08
Clarify group creation

bifurcation:
This PR is an alternative to #757.

This PR updates the group creation process so that epoch 0 is a "full-fledged" epoch, with an epoch secret (and all the derivatives) instead of just an init secret.  One implication of that is that the interim transcript hash can now be computed in the usual way, by extending the confirmed transcript hash with a confirmation tag:

```
// creator sets epoch secret
confirmation_key = DeriveSecret(epoch_secret, "confirm")
confirmed_transcript_hash = ""
confirmation_tag = MAC(confirmation_key, confirmed_transcript_hash)
interim_transcript_hash = Hash(confirmed_transcript_hash || confirmation_tag)
```

This has the nice side effect of immediately mixing the epoch secret into the transcript hash, helping the group diverge (a concern @kkohbrok raised on the call today).

While I was editing the group creation section, I also rationalized the requirements a bit.  One reason that #757 came up is that the group creation section currently assumes you're always starting with multiple members.  So the MUST-level requirement to fetch key packages has been downgraded to a SHOULD.

 Finally, I updated the specification of GroupInfo to explain the fields, in particular explaining how you get the confirmation tag in epoch 0 or later epochs.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2888,7 +2888,7 @@ struct {
 ~~~
 
 ~~~ pseudocode
-interim_transcript_hash_[0] = ""; /* zero-length octet string */
+confirmed_transcript_hash_[0] = ""; /* zero-length octet string */
 
 confirmed_transcript_hash_[n] =
     Hash(interim_transcript_hash_[n] ||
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -3361,19 +3361,25 @@ The client verifies the validity of a KeyPackage using the following steps:
...
-participants.  These processes are described in detail in {{add}}, {{commit}},
-and {{joining-via-welcome-message}}.
+A group is always created with a single member, the "creator".  Other members
+are then added to the group using the usual Add/Commit mechanism.
 
-The creator of a group MUST take the following steps to initialize the group:
+The creator of a group is responsible for setting the group ID, ciphersuite, and
+initial extensions for the group.  If the creator intends to add other members
+at the time of creation, then it SHOULD Fetch KeyPackages for the members to be
+added, and select a version and ciphersuite according to the capabilities of the
```
mulmarta, line 3370: > select a version and ciphersuite

In the previous sentence he doesn't choose a version, only ciphersuite. I guess the version is always MLS1.0?

draft-ietf-mls-protocol.md
```diff
@@ -3383,29 +3389,27 @@ The creator of a group MUST take the following steps to initialize the group:
   * Tree hash: The root hash of the above ratchet tree
   * Confirmed transcript hash: The zero-length octet string
   * Interim transcript hash: The zero-length octet string
```
mulmarta, line 3391: Remove this line?

draft-ietf-mls-protocol.md
```diff
@@ -2888,7 +2888,7 @@ struct {
 ~~~
 
 ~~~ pseudocode
-interim_transcript_hash_[0] = ""; /* zero-length octet string */
+confirmed_transcript_hash_[0] = ""; /* zero-length octet string */
```
mulmarta, line 2891: The indexing here is a bit confusing : if `n` starts at `0`, then the computation of `confirmed_transcript_hash_[n]` below contradicts this starting point, and the computation of `interim_transcript_hash_[n+1]` means there is no interim hash at `n=0`.

How about this instead : instead of `n`, use `epoch` and say explicitly that `x_hash[epoch]` is stored in the state when epoch id is `epoch` and `XHashInput[epoch]` contains values from the commit creating `epoch`.  Then swap the computations below so that for any `epoch >= 0` we compute 1) `interim_transcript_hash[epoch]` using `confirmed_transcript_hash[epoch]` and 2) `confirmed_transcript_hash[epoch+1]` using `interim_transcript_hash[epoch]` and `ConfirmedTranscriptHashInput[epoch+1]`?

mulmarta:
Since this PR clarifies more than #757 and the solution is equally good, I'd suggest closing #757 and merging this one.

PR #828, opened 2022-12-08
Update table header dividers

ericcornelissen:
Update all tables to use hyphens for table dividers instead of equal signs. Suggesting this because using equal signs leads to the tables not being rendered correctly on GitHub (preview below).

I tried to see if there's a reason for using one over the other, but couldn't find one. If there's some reason to use equal signs, or if this is considered a non-issue, feel free to close this.

---

##### Using equal signs

| this  | is    | an    | example |
|:======|:======|:======|:========|
| using | equal | signs | `=`     |

##### Using hyphens

| this  | is      | an      | example |
|:------|:--------|:--------|:--------|
| using | hyphens | instead | `-`     |


bifurcation (approved):
I think the RFC renderer doesn't care, but it's nice to have things display properly on GitHub as well.

PR #830, opened 2022-12-09
Update security and operational considerations

### LINKED ISSUE #762: Interaction of MLS and application
_Comment by @paulwouters_

I find the interaction with the application layer not clear. Not much is mentioned in this document. How can a group have an administrator? How is that enforced at the MLS layer? How do ACLs interact with MLS? (How are DoS attacks mitigated (eg spam attempts to join any group). What hooks would be in the authentication service and/or delivery server? How would clients know the properties of these two services?

This includes how it is not clear to me how race conditions are handled. What if two members try to expel each other at the same time? What if the group is so large, members always get their epoch wrong for things like Commit messages?

### LINKED ISSUE #763: Security against malicious members?
_Comment by @paulwouters_

There seems to be no group "owner", so inviting one member X, could that member than delete members from the group until the group is empty?

Apparently, there is ACLs for this, but the document doesn't describe how ACLS would work or how these would be implemented.

### LINKED ISSUE #813: Discuss metadata privacy

### LINKED ISSUE #815: Citations to security analysis
_Comment by @paulwouters_

~~~ 
though a complete security analysis is outside of the scope of this document.
~~~
Hmmm :P

bifurcation:
Fixes #762
Fixes #763
Fixes #815
Fixes #813

* Add a paragraph to Operational Context describing additional security services that the DS can provide
* Add a citation to security analyses via the MLS Architecture
* Add a section discussing metadata protection

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -181,15 +181,24 @@ MLS is designed to operate in the context described in
 {{?I-D.ietf-mls-architecture}}.  In particular, we assume that the following
 services are provided:
 
-* A Delivery Service that routes MLS messages among the participants in the
+* A Delivery Service (DS) that routes MLS messages among the participants in the
   protocol.  The following types of delivery are typically required:
 
   * Pre-publication of KeyPackage objects for clients
   * Broadcast delivery of Proposal and Commit messages to members of a group
   * Unicast delivery of Welcome messages to new members of a group
+  * Sequencing of Commit messages (see {{sequencing}})
 
-* An Authentication Service that enables group members to authenticate the
+* An Authentication Service (AS) that enables group members to authenticate the
   credentials presented by other group members.
+
+The DS and AS may also apply additional policies to MLS operations to obtain
+additional security properties.  For example, MLS enables any participant to add
+or remove members of a group; a DS could enforce a policy that only certain
+members are allowed to perform these operations.  MLS authenticates all members
+of a group; a DS could help ensure that only clients with certain types of
+credential are admitted. MLS provides no inherent protection against denial of
+service; A DS could also enforce rate limits in order to mitigate these risks.
 
 ##  Change Log
 
@@ -4819,7 +4828,8 @@ apply limits of the type discussed above.
 The security goals of MLS are described in {{?I-D.ietf-mls-architecture}}.
 We describe here how the protocol achieves its goals at a high level,
 though a complete security analysis is outside of the scope of this
-document.
+document.  The Security Considerations section of {{?I-D.ietf-mls-architecture}}
+provides some citations to detailed security analyses.
 
 ## Confidentiality of the Group Secrets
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -4836,6 +4846,53 @@ be securely removed from a group. It also allows a member to rotate their
...
+## Confidentiality of Group Metadata
+
+MLS handles certain metadata about a group that is sometimes exposed to parties
+outside the group:
+
+* Group ID
+* Group membership
+* Group extensions
+* Frequency of epoch changes
+* Frequency of application messages within an epoch
```
kkohbrok, line 4858: If we're using public handshake messages, it would also expose who is adding whom and when.

draft-ietf-mls-protocol.md
```diff
@@ -4836,6 +4846,53 @@ be securely removed from a group. It also allows a member to rotate their
...
+
+MLS provides no mechanism to protect the group ID and epoch of a message from
+the DS, so group ID and the frequency of messages and epoch changes are not
+protected against inspection by the DS.
+
+If a group can be joined by external commits (see
+{{joining-via-external-commits}}), then a group member will need to publish a
+GroupInfo object that states the group's extensions and a commitment to the
+group's membership.  This generally results in the group's membership and
+extensions being known to the DS and prospective new joiners.
```
kkohbrok, line 4875: No necessarily. The GroupInfo could be uploaded in encrypted form and then the key material could be passed to a prospective joiner.

It might also be worth mentioning here that the same goes if the tree is not sent as an extension via the Welcome.

draft-ietf-mls-protocol.md
```diff
@@ -4836,6 +4846,53 @@ be securely removed from a group. It also allows a member to rotate their
...
+If a group sends Proposal and Commit messages unencrypted (as MLSPlaintext),
+then the DS will be able to infer the group's membership from the set of Add and
+Remove proposals covered by Commits.  If Proposal and Commit messages are sent
+in encrypted form (as MLSCiphertext), then the DS will not be able to observe
+membership changes directly, but may be able to make some less direct
+inferences.  For example, the `new_member` field in a Welcome message indicates
+the KeyPackage to the Welcome is directed.  The Welcome message does not state
+which group it belongs to in plaintext, but if the DS is able to infer this
+(e.g., from timing signals), then the Welcome will tell the DS that a given
+client has joined the group.
```
kkohbrok, line 4894: This is only true if the DS has actually seen the KeyPackage, e.g. when it was uploaded. KeyPackages could be distributed in other ways, though, that are not visible to the DS.

Also it might be worth mentioning here that (global or per-group) pseudonyms can be used in conjunction with MLSPlaintext messages to avoid or at least reduce leakage of Metadata.

kkohbrok (commented):
Thanks for writing up the metadata considerations! Do they belong here or in the architecture doc?

kkohbrok:
It also feels like the text assumes that the DS is a server as opposed to the abstract layer that we define it as in the architecture document. Maybe we should qualify the considerations accordingly?

bifurcation:
@kkohbrok @raphaelrobert - Thanks for sending #833.  I didn't merge it directly, but rewrote the metadata section trying to hit the key points from your proposal.  LMK what you think.

draft-ietf-mls-protocol.md
```diff
@@ -4836,6 +4846,93 @@ be securely removed from a group. It also allows a member to rotate their
...
+A party observing these fields might be able to infer certain metadata about the
+group:
+
+* Group ID
+* Current epoch and frequency of epoch changes
+* Frequency of messages within an epoch
+* Group extensions
+* Group membership
+
+These fields are typically protected from parties other than the DS by applying
```
kkohbrok, line 4875: This suggests that the DS is a separate party. As I noted in the other PR, that is not necessarily the case. The DS is an abstract component parts of which may run on the clients themselves. It makes sense to use the DS as adversary when it comes to metadata, but I think its abstract nature forces us to qualify a bit here.

kkohbrok (commented):
Thanks for incorporating our ideas for this section. Looks good to me generally, but we should be careful about how we frame the DS.

draft-ietf-mls-protocol.md
```diff
@@ -4836,6 +4846,93 @@ be securely removed from a group. It also allows a member to rotate their
...
+"hop by hop" transport encryption (in contrast to the "end to end" protections
+provided by MLS). Whether the DS is exposed to unencrypted fields depends on
+several aspects of the DS design, such as:
+
+* How KeyPackages are distributed
+* How the ratchet tree is distributed
+* How proespective external joiners get a GroupInfo object for the group
```
raphaelrobert, line 4882:
```suggestion
* How prospective external joiners get a GroupInfo object for the group
```

draft-ietf-mls-protocol.md
```diff
@@ -4836,6 +4846,93 @@ be securely removed from a group. It also allows a member to rotate their
...
+is encrypted.  If a party has access to a pool of KeyPackages and observes a
+Welcome message, then they can identify the KeyPackage representing the new
+member.  If the party can also associate the Welcome with a group, then the
+party can infer that the identified new member was added to that group.
+
+Note that these information leaks reveal the group's membership to the degree
+that that membership is revealed by the contents of a member's LeafNode in the
+ratchet tree.  In some cases, this may be quite direct, e.g., due to credentials
+attesting to identifiers such as email addresses.  An application could
+cosntruct a member's leaf node to be less identifying, e.g., by using a
```
raphaelrobert, line 4933:
```suggestion
construct a member's leaf node to be less identifying, e.g., by using a
```

draft-ietf-mls-protocol.md
```diff
@@ -4836,6 +4846,93 @@ be securely removed from a group. It also allows a member to rotate their
...
+about the corresponding changes to the group's membership.  A party that sees
+all of these changes can reconstruct the group membership.
+
+Welcome messages contain a hash of each KeyPackage for which the Welcome message
+is encrypted.  If a party has access to a pool of KeyPackages and observes a
+Welcome message, then they can identify the KeyPackage representing the new
+member.  If the party can also associate the Welcome with a group, then the
+party can infer that the identified new member was added to that group.
+
+Note that these information leaks reveal the group's membership to the degree
```
raphaelrobert, line 4929:
```suggestion
Note that these information leaks reveal the group's membership only to the degree
```

raphaelrobert (changes requested):
Modulo a few nits and Konrad's comment, I think this works.

kkohbrok (approved):

PR #832, opened 2022-12-11
editorial: Minor fixes in presentation language.

duesee:
This PR fixes some minor issues with the presentation language. Moving `SignContent` and `RefHashInput` to their own `~~~ tls` block is required so that the `extract-tls.py` can now detect them.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1333,9 +1333,16 @@ Where SignContent is specified as:
 
 ~~~ tls
 struct {
-    opaque label<V> = "MLS 1.0 " + Label;
-    opaque content<V> = Content;
+    opaque label<V>;
+    opaque content<V>;
 } SignContent;
+~~~
+
+And its fields set to:
+
+~~~ pseudocode
+label = "MLS 1.0 " + Label;
+content = Content;
 ~~~
 
 Here, the functions `Signature.Sign` and `Signature.Verify` are defined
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1327,9 +1327,11 @@ SignWithLabel(SignatureKey, Label, Content) =
 
 VerifyWithLabel(VerificationKey, Label, Content, SignatureValue) =
     Signature.Verify(VerificationKey, SignContent, SignatureValue)
+~~~
 
 Where SignContent is specified as:
 
+~~~ tls
```
rohan-wire, line 1334: Anytime there is an equal sign and something that looks like a function definition, that is no longer TLS Presentation Language, but pseudocode.

I would move the definition of SignContent into a block like this before line 1334:
`~~~ tls`
`struct {`
`    opaque label<V>;`
`    opaque content<V>;`
`} SignContent;`

and then keep the content in the current block like this:

`    label = "MLS 1.0 " + Label;`
`    content = Content;`

draft-ietf-mls-protocol.md
```diff
@@ -1359,9 +1361,11 @@ MakeKeyPackageRef(value) = RefHash("MLS 1.0 KeyPackage Reference", value)
 MakeProposalRef(value)   = RefHash("MLS 1.0 Proposal Reference", value)
 
 RefHash(label, value) = Hash(RefHashInput)
+~~~
 
 Where RefHashInput is defined as:
 
+~~~ tls
```
rohan-wire, line 1368: Same concept here. This is not TLS Presentation Language.

rohan-wire (changes requested):
Separate pseudocode from genuine TLS Presentation Language please.

draft-ietf-mls-protocol.md
duesee: I see, thanks! I pushed a fixup in [360dbc7](https://github.com/mlswg/mls-protocol/pull/832/commits/360dbc7133c4923a4d7d29e921dc7587a013dfe3) and hope that I got your idea right.

rohan-wire (approved):

bifurcation (approved):

PR #833, opened 2022-12-12
Proposal for changes to #830

kkohbrok:


draft-ietf-mls-protocol.md
```diff
@@ -4854,45 +4854,63 @@ outside the group:
...
 * Group membership
 * Group extensions
-* Frequency of epoch changes
+* Current epoch and frequency of epoch changes
 * Frequency of application messages within an epoch
 
+### DS Architecture
+
+The following is relevant for applications where the DS is an entity less
+trusted than the members’ clients (e.g. a serverside component).
```
bifurcation, line 4863: The DS is always less trusted than the members clients.  The whole point of MLS is to protect against the DS :)  If you're thinking of things like p2p environments, then the DS would be the p2p links and any overlay routing, which is still less trusted than the client itself.

draft-ietf-mls-protocol.md
```diff
@@ -4854,45 +4854,63 @@ outside the group:
...
+### Group ID and Epoch
+
 MLS provides no mechanism to protect the group ID and epoch of a message from
 the DS, so group ID and the frequency of messages and epoch changes are not
 protected against inspection by the DS.
 
+### Extensions
+
+Group extensions are set via GroupContextExtensions proposals and thus exposed
+to the DS. Extensions can also be exposed via {{groupinfo-objects}}.
```
bifurcation, line 4886:
```suggestion
to the DS. Extensions can also be exposed via GroupInfo objects (see {{groupinfo-objects}}).
```

Otherwise it will render as "exposed via Section X.X.X"

This will render as "exposed via Section X.X.X".

draft-ietf-mls-protocol.md
kkohbrok: This has lead to quite a bit of confusion in the past and the architecture document is a bit ambiguous in the way the DS is defined. On the one hand you're right: The role of the DS involves message routing and it's described as a less trusted component. On the other hand, the role of the DS also involves commit ordering, which in case of a p2p system happens on the clients and which can not be done by overlay routing or p2p links.

I do agree though, that even as an abstract component, we can generally say that the DS is considered less trusted, even if it's (at least partially) running within the trust boundary of the MLS client.

Issue #834, opened 2022-12-14
Faster remove?


TWal:
When removing a participant, we blank its path up to the root because it might know the private key inside these nodes.
We actually know precisely which nodes contain private key that are known to this participant: it is nodes for which the participant is not an umerged leaf.

From this remark, we could design an alternative Remove that produces less blank nodes:
- blank all nodes on the path for which the participant is unmerged
- remove the participant from the unmerged leaves list of other nodes

I think this would preserve the parent-hash invariants, but would need to check more thoroughly.
(the intuition might rely on an invariant that is not checked on unmerged leaves, which is that the set of nodes for which a node is unmerged forms a connected path from the leaf going up [note that this invariant would have been true for free with #752!])

Pros:
- less blank nodes are produced
- removing a participant just after adding it does not change the tree, it's not that useful but it's a pretty property!

Cons:
- the remove definition is more complex
- the optimization isn't great if the participant is merged for most nodes (e.g., it has issued an UpdatePath in the past)

What do you think?

kkohbrok:
It sounds nice and it really sounds like it's secure enough, but I'm not sure it's not a bit too late for this.

bifurcation:
My personal analysis here:

* If we do Remove like this, we should do Update as well.  It's the same, you just don't remove from `unmerged_leaves`
* On the one hand, this gives some better symmetry to the `unmerged_leaves` calculations and might make things slightly faster
* On the other hand, this is pretty late for a substantive change and I don't expect the performance gain to be huge.

Overall, I'm more worried about the security analysis than the practicalities.  If @TWal can get his model updated and checked in a timely manner, I would probably take this change.

(...more detailed notes below this line...)

Practically speaking, it's a small spec diff.  (Note that we should cover Update here as well, which requires slightly different logic.)

``` diff
@@ -3599,7 +3599,9 @@ A member of the group applies an Update message by taking the following steps:
 
 * Replace the sender's LeafNode with the one contained in the Update proposal
 
-* Blank the intermediate nodes along the path from the sender's leaf to the root
+* For each intermediate node along the path from the sender's leaf to the root:
+  If the member's leaf node is not listed in the `unmerged_leaves` field of the
+  intermediate node, then blank the intermediate node.
  
@@ -3621,7 +3623,10 @@ A member of the group applies a Remove message by taking the following steps:
 
 * Replace the leaf node L with a blank node
 
-* Blank the intermediate nodes along the path from L to the root
[...]

PR #835, opened 2022-12-14
Have parent hash validation also verify unmerged leaves

bifurcation:
In #834, @raphaelrobert pointed out that we could be stricter in our validation of the `unmerged_leaves` field in the tree.  This PR updates the parent hash validation algorithm to also check that the `unmerged_leaves` relationships are consistent with the nodes having been updated at the same time.  In particular, it checks that the leaf node that signs the parent-hash chain is not unmerged at any of the parent nodes in the chain.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2729,6 +2729,12 @@ itself) and S the other child:
 * D is in the resolution of C, and the intersection of P's `unmerged_leaves`
   with the subtree under C is equal to the resolution of C with D removed.
 
+* If D is a parent node, the `unmerged_leaves` field of D is a subset of the
+  `unmerged_leaves` field of P.
+
+* If D is a leaf node, then the leaf index of D does not appear in the
+  `unmerged_leaves` field of P or any of its ancestors.
+
 These checks verify that D and P were updated at the same time (in the same
 UpdatePath), and that they were neighbors in the UpdatePath because the nodes in
 between them would have omitted from the filtered direct path.


draft-ietf-mls-protocol.md
```diff
@@ -2729,6 +2729,12 @@ itself) and S the other child:
 * D is in the resolution of C, and the intersection of P's `unmerged_leaves`
   with the subtree under C is equal to the resolution of C with D removed.
 
+* If D is a parent node, the `unmerged_leaves` field of D is a subset of the
+  `unmerged_leaves` field of P.
```
TWal, line 2733: This is implied by the property just above it
See the [discussion about the property](https://github.com/mlswg/mls-protocol/pull/713#issuecomment-1146399188) in #713, in which I show that the parent-hash link between D and P implies that `intersect(P.unmerged_leaves, subtree(D)) = D.unmerged_leaves`, which implies your condition.

draft-ietf-mls-protocol.md
```diff
@@ -2729,6 +2729,12 @@ itself) and S the other child:
...
+
+* If D is a leaf node, then the leaf index of D does not appear in the
+  `unmerged_leaves` field of P or any of its ancestors.
```
TWal, line 2736: This is almost implied by the resolution equation for the same reason as the new first condition.
Almost because we can't prove that `D` is not in the unmerged leaves of nodes not authenticated by `D`.

TWal (commented):
The current parent-hash guarantees are as strong as they can be: for every subtree, the version of this subtree at the time of its last modification (i.e. when we "un-add" its unmerged leaves) is authenticated by one of its leaves.
These new conditions do not improve this result.

bifurcation:
Closing in favor of #837

PR #836, opened 2022-12-14
Add a section discussing the security of the sender data protection

### LINKED ISSUE #770: KDF.Nh
_Comment by @paulwouters_

~~~ 
   The key and nonce provided to the AEAD are computed as the KDF of the
   first KDF.Nh bytes of the ciphertext generated in the previous
   section.
~~~
I don't think I understand the security model here? What happens if the previous ciphertext has 00's for the first KDF.Nh bytes?

bifurcation:
Fixes #770

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -110,12 +110,9 @@ informative:
        - name: Trevor Perrin(ed)
        - name: Moxie Marlinspike
 
-  SECG:
-    title: "Elliptic Curve Cryptography, Standards for Efficient Cryptography Group, ver. 2"
-    target: https://secg.org/sec1-v2.pdf
-    date: 2009
-
   SHS: DOI.10.6028/NIST.FIPS.180-4
+
+  NAN: DOI.10.1007/978-3-030-26948-7_9
 
 --- abstract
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -4921,6 +4918,41 @@ be securely removed from a group. It also allows a member to rotate their
...
+compromising sender data protection by reusing an AEAD (key, nonce) pair.  For
+example, in many AEAD schemes, reusing a key and nonce reveals the exclusive OR
+of the two plaintexts.  Assuming that the AEAD acts as a PRF, if L bits are
+sampled, the odds of two ciphertext samples being identical is roughly
+2<sup>-L/2</sup>, i.e., the birthday bound.
+
+The AEAD algorithms for ciphersuites defined in this document provide this
+property; the size of the sample depends on the ciphersuite's hash function, but
+in all cases, the probability of collision is no more than 2<sup>-128</sup>.
+Any future ciphersuite MUST use an AEAD algorithm is a PRF, especially with
```
kkohbrok, line 4953:
```suggestion
Any future ciphersuite MUST use an AEAD algorithm that is also a PRF, especially with
```

kkohbrok:
I would propose we say that we need AE1 security for our AEAD scheme, which gives us random ciphertexts and if we have random ciphertexts the input to key/nonce generation for sender data encryption is unique. This in turn gives us AE1 security for sender data encryption. I don't think we need AE2 encryption because our nonces are random and don't reveal any private information. Also, if I understand correctly, AES GCM is not AE2 secure, so if we want AE2 security, we need a different scheme or the transform described in NAN.

draft-ietf-mls-protocol.md
```diff
@@ -4921,6 +4918,41 @@ be securely removed from a group. It also allows a member to rotate their
...
+that are used for AEAD encryption of the sender data.
+
+``` pseudocode
+(key, nonce) = PRF(sender_data_secret, sample)
+encrypted_sender_data =
+  AEAD.Seal(key, nonce, sender_data_aad, sender_data)
+```
+
+The only difference between this construction and HN1 as described in {{NAN}} is
+that it uses authenticated encryption instead of unauthenticated encryption.
```
kkohbrok, line 4937: Plus we don't actually use the ciphertext as nonce, but instead just use it as context in the key/nonce derivation to ensure that the nonce doesn't repeat.

kkohbrok (commented):
Generally looks good, thanks for addressing my concerns. Just two nits.

kkohbrok (approved):

PR #837, opened 2022-12-15
Validate additional properties of unmerged leaves

bifurcation:
In discussion of #834, it came up that the current tree validation rules are loose enough that they allow configurations of `unmerged_leaves` that could not come about by normal operations.  This PR adds a check to enforce that if a node is unmerged at a parent node, it is also unmerged at every node between that parent node and the unmerged leaf.  

This is clearly a necessary property of any honestly-created tree, given the way `unmerged_leaves` entries are set and cleared (including the possible changes in #834).  I think it is also sufficient, but I don't have a proof on hand.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -4419,9 +4419,14 @@ welcome_key = KDF.Expand(welcome_secret, "key", AEAD.Nk)
   * For each non-empty leaf node, validate the LeafNode as described in
     {{leaf-node-validation}}.
 
-  * For each non-empty parent node, verify that each entry in the node's
-    `unmerged_leaves` represents a non-blank leaf node that is a descendant of
-    the parent node.
+  * For each non-empty parent node and each entry in the node's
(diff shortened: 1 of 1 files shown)


bifurcation:
Note that this property is *partly* checked by parent hash validation, [as @TWal notes](https://github.com/mlswg/mls-protocol/pull/835#discussion_r1048719223), but it still allows [some unnatural trees](https://github.com/mlswg/mls-protocol/issues/834#issuecomment-1352921209).  So the main impact of this PR is to have an explicit, complete check in addition to the implicit, partial check in parent hash validation.

kkohbrok (approved):

TWal:
Independently of #834, I think this is a useful property to have: this is a key invariant that is used in the O(n log log n) algorithm for parent-hash verification described in #527 and implemented in mlspp.

I don't know how hard it would be to add enough checks to reject every "unnatural trees", for example I don't think it's possible to have a blank node whose children are non-blank internal nodes, but a tree having this thing would be accepted when joining a group?

PR #838, opened 2022-12-15
Don't blank leaves that aren't known to a removed member

### LINKED ISSUE #834: Faster remove?

bifurcation:
Fixes #834

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3675,7 +3675,9 @@ A member of the group applies an Update message by taking the following steps:
 
 * Replace the sender's LeafNode with the one contained in the Update proposal
 
-* Blank the intermediate nodes along the path from the sender's leaf to the root
+* For each intermediate node along the path from the sender's leaf to the root:
+  If the member's leaf node is not listed in the `unmerged_leaves` field of the
+  intermediate node, then blank the intermediate node.
 
 ### Remove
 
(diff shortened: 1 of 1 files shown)


mulmarta:
While it sounds like a nice idea, I'm rather reluctant to make such changes at this stage. There has been a number of analyses of MLS and most of them are unlikely to be updated. I haven't been able to find an attack in a couple of days, but I don't feel certain enough that this is secure.

Would it be possible to make this an extension instead? Or maybe (also) a part of MLS 2.0?

raphaelrobert:
I would also like to put the optimisation into perspective again: This only makes removes faster in the narrow time window after a new member joined and before said member did its initial update. Furthermore, this only applies to the removal of new members. With that in mind, the optimisation should be more accurately renamed as "More efficient removal of new joiners".
New members are strongly encouraged to do an update as quickly as possible (even when the general update policy is more relaxed), therefore I expect this time window to be relatively short in an overwhelming majority of MLS deployments.

The risk @mulmarta describes is real, and I don't think it is justified by the marginal performance gain.

seanturner:
Discussed at 12/16/2022 interim. Decided to close, but it can be revived later as an extension.

PR #842, opened 2022-12-16
Add label and context to public-key encryption

bifurcation:
As discussed in the virtual interim 2022-12-16, it is possible that extensions might be tempted to re-use HPKE public keys created for MLS.  Even within MLS, we have two different public-key encryption cases, Welcome and Commit (though these currently use different keys).

Currently, for Commit encryption, the HPKE `info` parameter is set to the group context.  The content of `info` is actually undefined for Welcome encryption (!)  This PR makes the following changes:

* Define general `EncryptWithLabel` and `DecryptWithLabel` functions that populte the HPKE `info` parameter with an encoded label and context
* Apply these to Commit and Welcome encryption:
    * Commit: label="UpdatePathNode", context=`group_context`
    * Welcome: label="Welcome", context=`encrypted_group_info`
* Add an IANA registry of public-key encryption labels (as for signing)

There is also a bit of reorganization of the Ciphersuites section to split it into three subsections.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1292,23 +1292,21 @@ strings to HPKE key pairs.  As in HPKE, MLS assumes that an AEAD algorithm
 produces a single ciphertext output from AEAD encryption (aligning with
 {{?RFC5116}}), as opposed to a separate ciphertext and tag.
 
-Ciphersuites are represented with the CipherSuite type. HPKE public keys
-are opaque values in a format defined by the underlying
-protocol (see the Cryptographic Dependencies section of the HPKE specification for more
-information).
+Ciphersuites are represented with the CipherSuite type. The ciphersuites are
+defined in {{mls-ciphersuites}}.
+
+### Public Keys
+
+HPKE public keys are opaque values in a format defined by the underlying
+protocol (see the Cryptographic Dependencies section of the HPKE specification
+for more information).
 
 ~~~ tls
 opaque HPKEPublicKey<V>;
 ~~~
 
-The signature algorithm specified in the ciphersuite is the mandatory algorithm
-to be used for signatures in FramedContentAuthData and the tree signatures.  It MUST be
-the same as the signature algorithm specified in the credentials in the leaves
-of the tree (including the leaf node information in KeyPackages used to add new
-members).
-
-Like HPKE public keys, signature public keys are represented as opaque values in
-a format defined by the ciphersuite's signature scheme.
+Signature public keys are likewise represented as opaque values in a format
+defined by the ciphersuite's signature scheme.
 
 ~~~ tls
 opaque SignaturePublicKey<V>;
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1356,7 +1362,41 @@ they should re-use the SignWithLabel construction, using a distinct label.  To
...
+  opaque label<V>;
+  opaque context<V>;
+} EncryptContext;
+~~~
+
+And its fields set to:
+
+~~~
+label = "MLS 1.0 " + Label;
+content = Context;
```
raphaelrobert, line 1392:
```suggestion
context = Context;
```

draft-ietf-mls-protocol.md
```diff
@@ -1356,7 +1362,41 @@ they should re-use the SignWithLabel construction, using a distinct label.  To
...
+~~~
+
+Here, the functions `SealBase` and `OpenBase` are defined {{RFC9180}}, using the
+HPKE algorithms specified by the group's cipehersuite.  If MLS extensions
```
raphaelrobert, line 1396:
```suggestion
HPKE algorithms specified by the group's ciphersuite.  If MLS extensions
```

draft-ietf-mls-protocol.md
```diff
@@ -2441,26 +2481,23 @@ the `encrypted_path_secret` vector MUST be equal to the length of the resolution
...
-~~~ pseudocode
-context = SetupBaseR(kem_output, node_private_key, group_context)
-path_secret = context.Open("", ciphertext)
+path_secret =
+  DecryptWithLabel(node_private_key, "UpdatePathNode",
+                   group_context, kem_output, ciphertext)
 ~~~
 
+Here `node_public_key` is the public key of the node that the path secret is
+being encrypted for, `group_context` is the provisional GroupContext object for
```
raphaelrobert, line 2497:
```suggestion
Here `node_public_key` is the public key of the node for which the path secret is
encrypted, `group_context` is the provisional GroupContext object for
```

raphaelrobert (approved):
I think this captures what we discussed during the interim.

FWIW, I'm wondering if we didn't have a domain separation issue in the past when we only had one HPKE public key in KeyPackages and used it for both the Welcome subsequently for UpdatePath as well. Either way, I think this is a good defense-in-depth mechanism.

draft-ietf-mls-protocol.md
```diff
@@ -1356,7 +1362,41 @@ they should re-use the SignWithLabel construction, using a distinct label.  To
...
+
+~~~
+label = "MLS 1.0 " + Label;
+content = Context;
+~~~
+
+Here, the functions `SealBase` and `OpenBase` are defined {{RFC9180}}, using the
+HPKE algorithms specified by the group's cipehersuite.  If MLS extensions
+require HPKE encryption operations, they should re-use the EncryptWithLabel
+construction, using a distinct label.  To avoid collisions in these labels, an
```
mulmarta, line 1398: Would be possible to recommend or even RECOMMEND that extensions use separate HPKE keys (which can be included in leaves as extensions)?

mulmarta:
> FWIW, I'm wondering if we didn't have a domain separation issue in the past when we only had one HPKE public key in KeyPackages and used it for both the Welcome subsequently for UpdatePath as well. Either way, I think this is a good defense-in-depth mechanism.

Note that any previous analysis of MLS had to take this particular reuse into account. This can't be the case for any arbitrary extensions reusing the key. So if MLS uses an extension that reuses the key, none of the existing analysis applies, which is a much bigger issue.

This said, this reuse did decrease security -- it weakened FS in that a joiner had to hold on to their HPKE leaf key, which also allowed to decrypt the init secret from the welcome message, long after joining.

PR #844, opened 2022-12-20
cleanup after plaintext and ciphertext renaming.

kkohbrok:
Just stumbled over these remaining occurrences of plaintext/ciphertext. Was this intentional?


Bren2010 (approved):
Thanks!

bifurcation (approved):

PR #849, opened 2023-01-31
IESG Review by Erik Kline

bifurcation:
https://datatracker.ietf.org/doc/draft-ietf-mls-protocol/ballot/#draft-ietf-mls-protocol_erik-kline

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -962,7 +962,7 @@ for the new epoch so that it is not known to the removed member.
 Note that this does not necessarily imply that any member
 is actually allowed to evict other members; groups can
 enforce access control policies on top of these
-basic mechanism.
+basic mechanisms.
 
 ~~~ aasvg
                                                           Group
@@ -1874,7 +1874,7 @@ and PrivateContentTBE.
 
 ### Content Encryption
 
-Content to be encrypted is encoded in an PrivateContentTBE structure.
+Content to be encrypted is encoded in a PrivateContentTBE structure.
 
 ~~~ tls
 struct {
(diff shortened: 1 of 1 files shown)


ekline:
:+1:

draft-ietf-mls-protocol.md
```diff
@@ -2217,7 +2220,9 @@ The client verifies the validity of a LeafNode using the following steps:
   * If instead the LeafNode appears in a message being received by the client, e.g.,
     a proposal, a commit, or a ratchet tree of the group the client is joining, it is
     RECOMMENDED that the client verifies that the current time is within the range
-    of the `lifetime` field.
+    of the `lifetime` field.  (This check is not required because the LeafNode
```
raphaelrobert, line 2223:
```suggestion
    of the `lifetime` field.  (This check is not compulsory because the LeafNode
```

kkohbrok (approved):

draft-ietf-mls-protocol.md
rohan-wire: "compulsory" is poorly understood. Please leave as "required"

rohan-wire (approved):

draft-ietf-mls-protocol.md
bifurcation: I can see the confusion here if you read "not required" as "not needed" as opposed to "not an absolute requirement".  Suggest "not mandatory", since this is well understood to have the same semantic as "required" (see, e.g., "mandatory to implement").

```suggestion
    of the `lifetime` field.  (This check is not mandatory because the LeafNode
```

PR #850, opened 2023-01-31
IESG Review from Roman Danyliw

bifurcation:
https://datatracker.ietf.org/doc/draft-ietf-mls-protocol/ballot/#draft-ietf-mls-protocol_roman-danyliw

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -591,6 +591,22 @@ Group Context:
 Signature Key:
 : A signing key pair used to authenticate the sender of a message.
 
+Proposal:
+: A message that proposes a change to the group, e.g., adding or removing a
+member.
+
+Commit:
+: A message that implements the changes to the group proposed in a set of
+Proposals.
+
+PublicMessage:
+: An MLS protocol message that is signed by its sender and authenticated as
+coming from a member of the group in a particular epoch, but not encrypted.
+
+PrivateMessage:
+: An MLS protocol message that is both signed by its sender and encrypted so
+that is confidential to the members of group in a particular epoch.
+
 Handshake Message:
 : A PublicMessage or PrivateMessage carrying an MLS Proposal or Commit
 object, as opposed to application data.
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -591,6 +591,22 @@ Group Context:
...
+: A message that implements the changes to the group proposed in a set of
+Proposals.
+
+PublicMessage:
+: An MLS protocol message that is signed by its sender and authenticated as
+coming from a member of the group in a particular epoch, but not encrypted.
+
+PrivateMessage:
+: An MLS protocol message that is both signed by its sender and encrypted so
+that is confidential to the members of group in a particular epoch.
```
raphaelrobert, line 608:
```suggestion
that it is confidential to the members of the group in a particular epoch.
```

kkohbrok (approved):

rohan-wire (approved):

draft-ietf-mls-protocol.md
```diff
@@ -591,6 +591,22 @@ Group Context:
...
+Commit:
+: A message that implements the changes to the group proposed in a set of
+Proposals.
+
+PublicMessage:
+: An MLS protocol message that is signed by its sender and authenticated as
+coming from a member of the group in a particular epoch, but not encrypted.
+
+PrivateMessage:
+: An MLS protocol message that is both signed by its sender and encrypted so
```
br-hale, line 607: Suggested edit
"by its sender, authenticated as coming from a member of the group in a particular epoch, and is confidential...". I.e. the definition of PrivateMessage mentions confidentiality but leaves off the authenticity mentioned under PublicMessage (albeit implemented as authenticated encryption).

draft-ietf-mls-protocol.md
bifurcation, line 608: Combining suggestions from @br-hale and @raphaelrobert ...

```suggestion
: An MLS protocol message that is both signed by its sender, authenticated as
coming from a member of the group in a particular epoch, and encrypted so
that it is confidential to the members of the group in that epoch.
```

PR #855, opened 2023-02-08
Clarify transcript hash initialization

bifurcation:
The current text on initializing the transcript hash is ambiguous.  In particular, it never tells you how to compute the first `interim_transcript_hash`. This PR provides clearer initialization instructions, and a figure that summarizes the transcript hash evolution [2].

image

[1] To be clear, I think the intent is clear and all current implementations are consistent.  The text is just bad.
[2] The diagram is a little over-the-top elaborate.  I would be OK removing it :)

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2992,21 +2992,42 @@ struct {
 ~~~
 
 ~~~ pseudocode
-confirmed_transcript_hash_[epoch] = ""; /* zero-length octet string */
+confirmed_transcript_hash_[0] = ""; /* zero-length octet string */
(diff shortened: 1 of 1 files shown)


kkohbrok:
I generally like the diagram, but the intersection between `confirmed`, commitX` and `confirm_tag` is not clear to me. What is going where? Also shouldn't `confirmed` and `confirm_tag` also get numbers? And what's the different between the `1` and [1]` suffix?

bifurcation:
OK, I rotated the diagram 90 degrees and rejiggered things to hopefully be clearer.  It now should just barely fit in the .TXT format.  What do folks think?

image

Bren2010 (approved):

kkohbrok (approved):
Beautiful!

Issue #860, opened 2023-02-10
Loosen chain requirements


ekr:
> For an X.509 credential, each entry in the chain represents a single DER-encoded X.509 certificate. The chain is ordered such that the first entry (chain[0]) is the end-entity certificate and each subsequent certificate in the chain MUST be the issuer of the previous certificate. The public key encoded in the subjectPublicKeyInfo of the end-entity certificate MUST be identical to the signature_key in the LeafNode containing this credential.

I'm sorry I missed this previously, but this requirement was a source of trouble in TLS and was removed in RFC 8446, which allows extra certificates in the middle. Was there a conscious decision to be more strict here.

Bren2010:
I don't recall this ever being discussed. I think it was just added and nobody ever objected.

kkohbrok:
@ekr can you elaborate a little bit? The way I read it the requirement is just that there is a valid signature chain, i.e. that one certificate signs the next and so on.

bifurcation:
@ekr - I deliberately reverted to the original TLS level of strictness here, in hopes that in starting over with a new ecosystem we could reset things to a cleaner state.  Do you think there are legitimate reasons for violating the strict rule, or is it just a common misconfiguration in TLS servers?  I would not really be inclined to accommodate the latter.

@kkohbrok - I believe what @ekr is envisioning are cases where there are multiple valid chains, say via old and new versions of an intermediate certificate.  So you might do something like provide a "chain" of the form (leaf, old_intermediate, new_intermediate), with the idea that the client could verify using whichever one makes it happy.

tomleavy:
@bifurcation take a look at the comment I left on the PR. I can see your point about starting over as being valid, but one thing that also caught my eye in the TLS RFC was language around allowing a root CA to be left out of the chain since those would potentially already exist in a local trust store.

kkohbrok:
I see. Thanks @bifurcation for the clarification. Maybe we can phrase it in such a way that whatever the credential includes (added intermediary certificates or left-out root certificates) in the end the local client MUST be able to create a continuous certificate chain from the leaf to the root. Do we need to define how the client does that? Since we're already pushing everything else related to authentication into the AS, I'm inclined to say no.

PR #861, opened 2023-02-10
#860: Loosen chain requirements.

Bren2010:
This updates the language around how certificate chains are constructed:
- We allow certificates to be omitted if the client knows any recipients will already possess the omitted certificate.
- We allow certificate chains to be out-of-order and potentially contain unused certificates.

This is nice because it means less data needs to be sent on the wire, and also provides a method for root certificates to be rotated gracefully (as clients can provide chains for both old and new roots).

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1501,7 +1501,7 @@ struct {
             opaque identity<V>;
 
         case x509:
-            Certificate chain<V>;
+            Certificate certificates<V>;
     };
 } Credential;
 ~~~
@@ -1509,7 +1509,7 @@ struct {
 A "basic" credential is a bare assertion of an identity, without any additional
 information.  The format of the encoded identity is defined by the application.
 
-For an X.509 credential, each entry in the chain represents a single DER-encoded
+For an X.509 credential, each entry in the `certificates` field represents a single DER-encoded
 X.509 certificate. The chain is ordered such that the first entry (chain[0]) is
 the end-entity certificate. The public key encoded in the
 `subjectPublicKeyInfo` of the end-entity certificate MUST be identical to the
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -835,8 +835,8 @@ entropy to the group.  This fresh entropy is provided only to members in the new
 epoch and not to members who have been removed. Commits thus maintain the property that
 the epoch secret is confidential to the members in the current epoch.
 
-For each Commit that adds one or more members to the group, there is a single corresponding
-Welcome message.  The Welcome message provides all the new members with the information
+For each Commit that adds one or more members to the group, there are one or more corresponding
+Welcome messages.  Each Welcome message provides new members with the information
 they need to initialize their views of the key schedule and ratchet tree, so
 that these views align with the views held by other members of the group
 in this epoch.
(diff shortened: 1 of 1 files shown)


tomleavy:
@Bren2010 it might be useful to also add this language from RFC 8446. It would make the process exactly the same as TLS which makes implementation easier.

>The sender's certificate MUST come in the first
>CertificateEntry in the list.  Each following certificate SHOULD
 >directly certify the one immediately preceding it.  Because
 >certificate validation requires that trust anchors be distributed
 >independently, a certificate that specifies a trust anchor MAY be
 >omitted from the chain, provided that supported peers are known to
 >possess any omitted certificates.

tomleavy (approved):

draft-ietf-mls-protocol.md
```diff
@@ -1511,10 +1511,11 @@ information.  The format of the encoded identity is defined by the application.
 
 For an X.509 credential, each entry in the chain represents a single DER-encoded
 X.509 certificate. The chain is ordered such that the first entry (chain[0]) is
-the end-entity certificate and each subsequent certificate in the chain MUST be
-the issuer of the previous certificate.  The public key encoded in the
+the end-entity certificate and each subsequent certificate in the chain SHOULD be
```
rohan-wire, line 1514: I am very uncomfortable with this change for two reasons. Every SHOULD needs to have a statement explaining when you would not do the recommended behavior and why. Also, since this is regarding credential validation I think this is a pretty terrible idea. 

I would prefer to leave this as a MUST. I will comment on another way to accomplish the stated goal in the Conversation part of the PR.

Bren2010:
> Don't we already have a better mechanism for rotating certificates? As soon as a member has a new certificate chain, it can update its LeafNode in an UpdatePath. This semantics of X.509 certificate bundles are already a bit fuzzy. Such a change will make validation more complicated and more error prone, when this seems completely unnecessary.

The scenario that this would address is: say that a root CA is compromised and needs to be replaced. Distributing the new CA certificate likely requires a software release, which not everyone will install simultaneously. During the rollout, some users will know only the old compromised CA and some users will know only the new CA and not the compromised one. To preserve interop during the rollout, users need to include two intermediates in their certificate chain: an intermediate signed by the old root that signs their leaf, and an intermediate signed by the new root that signs their leaf. (In the exact scenario I'm thinking of, the intermediate private key is held by the user, so it would not be compromised with the root. Though there may be other arrangements where this flexibility helps.)

In terms of implementation complexity: the X509 chain validation libraries I'm aware of already support path building because it's necessary for TLS, so I would expect those could be used

tomleavy:
> In terms of implementation complexity: the X509 chain validation libraries I'm aware of already support path building because it's necessary for TLS, so I would expect those could be used

This is actually the reason why I think it makes sense to adopt the same mechanism. For example, if you use OpenSSL to validate the chain it does not take a chain as input. It takes a collection of certificates, a set of trusted root certs, and the leaf you want to validate. I believe not making this change has a higher complexity.

draft-ietf-mls-protocol.md
```diff
@@ -1511,10 +1511,10 @@ information.  The format of the encoded identity is defined by the application.
 
 For an X.509 credential, each entry in the chain represents a single DER-encoded
```
bifurcation, line 1512:
```suggestion
For an X.509 credential, each entry in the `certificates` field represents a single DER-encoded
```

bifurcation: ... and the corresponding change to the struct definition above.

bifurcation (approved):
One minor comment, otherwise this looks good to me.

bifurcation (approved):

PR #864, opened 2023-02-16
Change MLS-Exporter label from "exporter" to "exported"

TWal:
Currently the "exporter" label is used two times : to derive `exporter_secret`, and in `MLS-Exporter`.
I'm not sure it's really problematic, but using different labels doesn't cost much, hence this PR.


bifurcation (approved):

duesee (commented):
I think this should be adapted, too: https://github.com/mlswg/mls-protocol/pull/864/files#diff-7c369b85b26a746a7e70cd2884037896defe03b3098bec976f79c994d22a604aR2937

bifurcation:
@duesee, no the point is for the two labels to be different :)

duesee:
Argh... confused myself :-)

PR #866, opened 2023-02-21
Allow group context extensions proposal without path

mulmarta:
Requiring paths to change group context seems too strict.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -4096,6 +4096,7 @@ The only proposal types defined in this document that do not require a path are:
 * `add`
 * `psk`
 * `reinit`
+* `group_context_extensions`
 
 New proposal types MUST state whether they require a path. If any instance of a
 proposal type requires a path, then the proposal type requires a path. This


Bren2010:
I'm slightly concerned about this because it covers extensions, so we don't know exactly what all will be happening here. You're saying there are /no/ extensions possible where it makes sense to maintain the stronger security of a path-populated Commit

mulmarta:
> I'm slightly concerned about this because it covers extensions, so we don't know exactly what all will be happening here. You're saying there are /no/ extensions possible where it makes sense to maintain the stronger security of a path-populated Commit

An extension can still specify that a commit that changes its value in the group context requires a path. So I think what this is doing is changing the default behavior to allow not updating the path (since most extensions won't need it). However, this behavior can be overridden by extensions.

bifurcation:
I think this was discussed and the feeling was that sending the path was more conservative, since you don't know in advance whether a future extension would require a path.  It also lets you decide at the proposal-type level, without having to get into the internals of the GroupContextExtensions.

If we are going to do something here, I would suggest being more formal than this PR is, and defining for extensions a notion analogous to "path required" for Proposals.  That way at least there's an algorithm defined.

I would probably prefer to leave this as-is, though.

bifurcation:
I think this was discussed and the feeling was that sending the path was more conservative, since you don't know in advance whether a future extension would require a path.  It also lets you decide at the proposal-type level, without having to get into the internals of the GroupContextExtensions.

If we are going to do something here, I would suggest being more formal than this PR is, and defining for extensions a notion analogous to "path required" for Proposals.  That way at least there's an algorithm defined.

I would probably prefer to leave this as-is, though.

mulmarta:
No strong opinions.

PR #839, opened 2022-12-15
Remove "MLS" prefix on structs

bifurcation:
Fixes #759 

As discussed in #759, this PR makes the following terminology changes:

| Before                      | After                     |
|:----------------------------|:--------------------------|
| MLSMessage                  | MLSMessage                |
| MLSContent                  | GroupContent              |
| MLSContentTBS               | GroupContentTBS           |
| MLSContentAuthData          | GroupContentAuthData      |
| MLSAuthenticatedContent     | AuthenticatedContent      |
| MLSContentTBM               | AuthenticatedContentTBM   |
| MLSSenderData               | SenderData                |
| MLSSenderDataAAD            | SenderDataAAD             |
| MLSCiphertextContent        | AuthenticatedContentTBE   |
| MLSCiphertextContentAAD     | ContentAAD                |
| MLSPlaintext                | PublicMessage             |
| MLSCiphertext               | PrivateMessage            |

It might be useful to review the two commits here separately.  The first is just a global search-and-replace.  The second is a light review for grammar ("a" vs. "an"), and makes a few small copy edits.

PR #867, opened 2023-02-21
Rename `SenderData` to `SenderDataTBE`

TWal:
With the renaming in #839, some message formats used before encryption / after decryption were suffixed with `TBE` (with the same spirit as `TBS`).
However `SenderData` is only used before encryption / after decryption, so it makes sense to add `TBE` in the name, hence this pull request.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -1899,7 +1899,7 @@ struct {
 ~~~
 
 `encrypted_sender_data` and `ciphertext` are encrypted using the AEAD function
-specified by the ciphersuite in use, using as input the structures SenderData
+specified by the ciphersuite in use, using as input the structures SenderDataTBE
 and PrivateContentTBE.
 
 ### Content Encryption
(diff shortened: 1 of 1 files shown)


Bren2010:
Honestly I'd prefer not to have the extra cruft

TWal:
I'm not opposed to leave `SenderData` as-is, but I think the whole document should have a consistent naming convention: either every message format used as input for encryption/decryption has `TBE` suffixed, or none of them has.
A middle-ground where some are named with `TBE` and some aren't is not great I think.

bifurcation:
I am also disinclined to do this.  We have the following structs that are inputs:

* `GroupSecrets`
* `GroupInfo`
* `SenderData`
* `PrivateMessageTBE`

If anything, I would be inclined to remove the `TBE`.  Maybe call it `PrivateMessageContent`.

TWal:
Thanks for the exhaustive list!
Dropping the TBE makes sense. I think the difference between `TBE` and `TBS`/`TBM` is that `xxxTBS` and `xxxTBM` are always linked to another message format `xxx`, which is not the case of `xxxTBE`.

TWal:
Closing in favor of #870.

PR #872, opened 2023-03-03
Comments from Nick Sullivan

bifurcation:
Nick sent me some late comments, which I have reflected in this PR.  The most significant changes are:

* A new section in the overview covering external commits, so that that concept has been introduced before it is encountered further down.

* A new subsection in the security considerations elaborating why key uniqueness mattes, and some slightly stronger requirements around it in the appropriate spots.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -190,13 +190,13 @@ services are provided:
   * Sequencing of Commit messages (see {{sequencing}})
 
 The DS and AS may also apply additional policies to MLS operations to obtain
-additional security properties, especially in cases where MLS handshake messages
+stronger security properties, especially in cases where MLS handshake messages
 are transmitted unencrypted (as PublicMessage objects).  For example, MLS
 enables any participant to add or remove members of a group; a DS could enforce
 a policy that only certain members are allowed to perform these operations.  MLS
 authenticates all members of a group; a DS could help ensure that only clients
 with certain types of credential are admitted. MLS provides no inherent
-protection against denial of service; A DS could also enforce rate limits in
+protection against denial of service; a DS could also enforce rate limits in
 order to mitigate these risks.
 
 ##  Change Log
@@ -587,7 +587,7 @@ Key Package:
 
 Group Context:
 : An object that summarizes the shared, public state of the group. The group
-  context is typically distributed in a signed GroupInfo message, which provided
+  context is typically distributed in a signed GroupInfo message, which is provided
   to new members to help them join a group.
 
 Signature Key:
@@ -1045,7 +1045,7 @@ well as a public key that can be used to encrypt a secret to the existing
 members of the group.  When the new member Z wishes to join, they download the
 GroupInfo object and use it to form a Commit of a special form that adds Z to
 the group (as detailed in {{joining-via-external-commits}}).  The existing
-members of the group process this external Commit in a similar way to normal
+members of the group process this external Commit in a similar way to a normal
 Commit, advancing to a new epoch in which Z is now a member of the group.
 
 ~~~
@@ -1065,7 +1065,7 @@ A              B              Z          Directory        Channel
 |              |              |<----------------------------+
 |              |              |              |              |
 ~~~
-{: #groupinfo-flow title"Client A publishes a GroupInfo object and Client Z uses
+{: #groupinfo-flow title="Client A publishes a GroupInfo object and Client Z uses
 it to join the group"}
 
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -148,10 +148,10 @@ as well as post-compromise security, but are nonetheless efficient
 enough for heavy use over low-bandwidth networks.
 
 For a group of size greater than two, a common strategy is to
-unilaterally broadcast symmetric "sender" keys over existing shared
-symmetric channels, and then for each member to send messages to the
+distribute symmetric "sender keys" over existing 1:1
+secure channels, and then for each member to send messages to the
 group encrypted with their own sender key. On the one hand, using sender keys
-improves efficiency relative to pairwise broadcast of individual messages, and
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -1010,15 +1014,76 @@ A              B     ...      Z          Directory       Channel
...
 resumption pre-shared key (PSK) from each epoch to allow entropy extracted from
-one epoch to be injected into a future epoch. This link guarantees that members
-entering the new epoch agree on a key if and only if they were members of the group
-during the epoch from which the resumption key was extracted.
+one epoch to be injected into a future epoch.  A group member that wishes to
+inject a PSK issues a PreSharedKey proposal ({{presharedkey}}) describing the
+PSK to be injected.  When this proposal is committed, the corresponding PSK will
+be incorporated into the key schedule as described in {{pre-shared-keys}}.
+
+Linking epochs in this way
```
raphaelrobert, line 1083: strange line break

raphaelrobert (approved):
Looks good modulo the existing suggestions and comments

draft-ietf-mls-protocol.md
```diff
@@ -1010,15 +1014,76 @@ A              B     ...      Z          Directory       Channel
...
+Note that the flows in this section are examples; applications can arrange
+message flows in other ways.  For example:
+
+* Welcome messages don't necessarily need to be sent directly to new joiners.
+  Since they are encrypted to new joiners, they could be distributed more
+  broadly, say if the application only had access to a broadcast channel for the
+  group.
+
+* Proposal messages don't necessarily need to be sent to all group members. They
+  only need to reach a member who will commit them.
```
bifurcation, line 1026:
```suggestion
* Proposal messages don't need to be immediately sent to all group members.  They need to
  be available to the committer before generating a commit, and to other members before
  processing the commit.
```

bifurcation: @Bren2010 look good?

draft-ietf-mls-protocol.md
bifurcation: Doesn't change the rendering.  I'll fix it if I end up having to do edits locally, vs in the GitHub UI :)

draft-ietf-mls-protocol.md
```diff
@@ -5390,6 +5469,23 @@ be mitigated by having clients regularly generate new KeyPackages and upload
...
 to a new group is more likely to be fresh and less likely to be compromised.
 
+## Uniqueness of Ratchet Tree Key Pairs
+
+The encryption and signature keys stored in ratchet tree nodes MUST be distinct
+from one another.  If two members' leaf nodes have the same signature key, then
+the data origin authentication properties afforded by signatures within the
+group are degraded.  If two nodes have the same encryption key, it creates a
+"double join" scenario where a removed member may still be able to compute the
+group's secrets.
```
mulmarta, line 5479: I'm not sure I get this argument. First, keys could be related in different ways, e.g. `pk2 = pk1 ^ 2`. If I know `sk1` corresponding to `pk1`, I can decrypt ciphertexts for `sk2 = 2 * sk1` corresponding to `pk2` as well. In general, such relations can't be detected, so I don't see how this check _can_ improve security. On the other hand, it's a bit annoying to implement.

Another reason I don't think this check prevents double joins by malicious insiders. Say leaves A and B have the same HPKE key. This could be because:

* A is honest and a malicious B copied A's key. Since B doesn't know the secret key, this is not an attack and B can be removed.
* A is honest and generated the same key as a malicious B. This clearly can't happen because keys are random.
* A and B are malicious, they collude and they introduced the same key. In this case security is restored only after removing both insiders. (Since A and B collude, they are one entity in a way)

draft-ietf-mls-protocol.md
```diff
@@ -2301,9 +2371,12 @@ The client verifies the validity of a LeafNode using the following steps:
...
 
-* Verify that the following fields are unique among the members of the group:
-    * `signature_key`
-    * `encryption_key`
+* Verify that the leaf node's `signature_key` value does not appear in any other
+  leaf node in the group's ratchet tree.
+
+* Verify that the leaf node's `encryption_key` value does not appear in any
+  other node of the group's ratchet tree (including both leaf nodes and parent
+  nodes).
```
mulmarta, line 2379: What exactly does it mean to "appear in any other node"? Is it OK e.g. if a signature or an HPKE key of a leaf appears in an intermediate certificate in the X509 chain in the credential of another leaf? Should we also check that the keys don't appear in other places e.g. in the external signers extension or as the HPKE key for external joiners?

draft-ietf-mls-protocol.md
```diff
@@ -1166,15 +1228,22 @@ at `0` from the left to `2^d - 1` at the right (for a tree with `2^d` leaves). A
...
 * An ordered list of "unmerged" leaves (see {{views}})
 * A hash of certain information about the node's parent, as of the last time the
   node was changed (see {{parent-hashes}}).
 
+As described in {{views}}, different members know different subsets of the set
+of private keys corresponding to the public keys in nodes in the tree.  The
+private key corresponding to a parent node is known only to members at leaf
+nodes that are descedants of that node.  The private key corresponding to a leaf
+node is known only to the member at that leaf node.  A leaf node is _unmerged_
+relative to one of its ancestor node if the member at the leaf node does not
```
Bren2010, line 1244:
```suggestion
relative to one of its ancestor nodes if the member at the leaf node does not
```

draft-ietf-mls-protocol.md
```diff
@@ -1166,15 +1228,22 @@ at `0` from the left to `2^d - 1` at the right (for a tree with `2^d` leaves). A
...
+know the private key corresponding the ancestor node.
```
Bren2010, line 1245:
```suggestion
know the private key corresponding to the ancestor node.
```

Bren2010 (approved):

PR #873, opened 2023-03-07
Security considerations

ekr:
This PR adds a few more key points to the security considerations that probably deserve to be in this document.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -5233,13 +5233,16 @@ was compromised at some point in the past.
 ~~~
 {: title="Forward secrecy and post-compromise security" }
 
-Post-compromise security is provided between epochs by members regularly
-updating their leaf key in the ratchet tree. Updating their leaf key prevents
-group secrets from continuing to be encrypted to public keys whose private keys
-had previously been compromised. Note that sending an Update proposal does not achieve
-PCS until another member includes it in a Commit. Members can achieve immediate
-PCS by sending their own Commit and populating the `path` field, as
-described in {{commit}}.
+Post-compromise security is provided between epochs by members
+regularly updating their leaf key in the ratchet tree. Updating their
+leaf key prevents group secrets from continuing to be encrypted to
+public keys whose private keys had previously been compromised. Note
+that sending an Update proposal does not achieve PCS until another
+member includes it in a Commit. Members can achieve immediate PCS by
+sending their own Commit and populating the `path` field, as described
+in {{commit}}. To be clear, in all these cases, the PCS guarantees
+come into effect when the members of the group process the relevant
+Commit, not when the sender creates it.
 
 Forward secrecy between epochs is provided by deleting private keys from past
 versions of the ratchet tree, as this prevents old group secrets from being
@@ -5289,10 +5292,11 @@ DS to influence which Commit is applied, even to the point of
 preventing a member from ever having its Commits applied.
 
 When put together, these abilities potentially allow a DS to collude
-with a compromised member to defeat PCS by suppressing the valid
-Update and Commit messages sent by the uncompromised endpoint for the
-same user. Aside from the generation, MLS leaves loss detection up to the
-application.
+with an attacker who has compromised a member's state to defeat PCS by
+suppressing the valid Update and Commit messages from the member that
+would lock out the attacker and update the member's leaf to a new,
+uncompromised state. Aside from the SenderData.generation value, MLS
+leaves loss detection up to the application.
 
 
 ## Authentication Service Compromise
@@ -5307,7 +5311,7 @@ compromised AS to join the group as an existing user, for instance as
 if it were a new device associated with the same user. If
 the application uses a transparency mechanism such as CONIKS
 {{CONIKS}} or Key Transparency {{KT}}, then it may be possible for end
-users to detect this kind of misbehavior by the AS.  It is also possible to 
+users to detect this kind of misbehavior by the AS.  It is also possible to
 construct schemes in which the various clients owned by a user vouch
 for each other, e.g., by signing each others' keys.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -5203,12 +5236,21 @@ was compromised at some point in the past.
 Post-compromise security is provided between epochs by members regularly
 updating their leaf key in the ratchet tree. Updating their leaf key prevents
 group secrets from continuing to be encrypted to public keys whose private keys
-had previously been compromised.
+had previously been compromised. Note that sending an Update does not achieve
+PCS until another member sends a Commit. Members can achieve immediate
```
kkohbrok, line 5240:
```suggestion
had previously been compromised. Note that sending an Update proposal does not achieve
PCS until another member includes it in a Commit. Members can achieve immediate
```

draft-ietf-mls-protocol.md
```diff
@@ -5231,6 +5273,43 @@ groups by exhausting all available KeyPackages. To prevent such a denial-of-serv
 attack, the KeyPackage publication system SHOULD rate-limit KeyPackage
 requests, especially if not authenticated.
 
+## Delivery Service Compromise
+
+MLS is designed to protect the confidentiality and integrity of
+the group data even in the face of a compromised DS. However, a compromised
+DNS can still mount some attacks. While it cannot forge messages,
```
kkohbrok, line 5280:
```suggestion
DS can still mount some attacks. While it cannot forge messages,
```

draft-ietf-mls-protocol.md
```diff
@@ -5231,6 +5273,43 @@ groups by exhausting all available KeyPackages. To prevent such a denial-of-serv
...
+it can selectively delay or remove them. This can in some cases be
+observed by detecting gaps in the per-sender generation counter,
+though it may not always be possible to distinguish attack from message
```
kkohbrok, line 5283:
```suggestion
though it may not always be possible to distinguish an attack from message
```

draft-ietf-mls-protocol.md
```diff
@@ -5231,6 +5273,43 @@ groups by exhausting all available KeyPackages. To prevent such a denial-of-serv
...
+loss. In addition, the DS can permanently block messages to and from
+a group member. This will not always be detectable by other members.
+If an application uses the DS to resolve conflicts between
+simultaneous Commits (see {{sequencing}}), it is also possible for the
+DS to influence which Commit is applied, even to the point of
+preventing a member from ever having its Commit applied.
```
kkohbrok, line 5289:
```suggestion
preventing a member from ever having its Commits applied.
```

draft-ietf-mls-protocol.md
```diff
@@ -5231,6 +5273,43 @@ groups by exhausting all available KeyPackages. To prevent such a denial-of-serv
...
+
+When put together, these abilities potentially allow a DS to collude
+with a compromised member to defeat PCS by suppressing the valid
+Update and Commit messages sent by the uncompromised endpoint for the
+same user. Aside from the generation, MLS leaves loss detection up the
```
kkohbrok, line 5294: If I understand correctly, the "compromised member" and "the uncompromised endpoint" are more or less the same here, aren't they? It might make sense to spell the scenario out a bit more.

Also "generation" refers to the generation in the messages, right? Since it's a relatively generic word, it might make sense to be explicit about that context.

draft-ietf-mls-protocol.md
```diff
@@ -5231,6 +5273,43 @@ groups by exhausting all available KeyPackages. To prevent such a denial-of-serv
...
+of the Delivery Service. A compromised AS can assert a binding for a
+signature key and identity pair of its choice, thus allowing
+impersonation of a given user. This ability is sufficient to allow the
+AS to join new groups as if it were that user. Depending on the
+application architecture, it may also be sufficient to allow the
+compromised AS to join the group as an existing user, for instance as
+if it were a new device associated with the same user. If
+the application uses a transparency mechanism such as CONIKS
+{{CONIKS}} or Key Transparency {{KT}}, then it may be possible for end
+users to detect this kind of misbehavior by the AS.
```
kkohbrok, line 5310: Since we're naming mitigation measures here, cross-signing of clients would also be worth mentioning here, I think.

kkohbrok (changes requested):
Great additions overall! Just a few nits and questions.

draft-ietf-mls-protocol.md
raphaelrobert, line 5294:
```suggestion
same user. Aside from the generation, MLS leaves loss detection up to the
```

draft-ietf-mls-protocol.md
```diff
@@ -5203,12 +5236,21 @@ was compromised at some point in the past.
 Post-compromise security is provided between epochs by members regularly
 updating their leaf key in the ratchet tree. Updating their leaf key prevents
 group secrets from continuing to be encrypted to public keys whose private keys
-had previously been compromised.
+had previously been compromised. Note that sending an Update does not achieve
+PCS until another member sends a Commit. Members can achieve immediate
+PCS by sending their own Commit and populating the `path` field, as
+described in {{commit}}.
```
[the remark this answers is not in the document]
bifurcation: This seems like a worthwhile distinction.  Something like:

> To be clear, in all these cases, the PCS guarantees come into effect when the members of the group process the relevant Commit, not when the sender creates it.

draft-ietf-mls-protocol.md
bifurcation: Agree that this is a little ambiguous.  Suggest something like:

> When put together, these abilities potentially allow a DS to collude with an attacker who has compromised a member's state to defeat PCS by suppressing the valid Update and Commit messages from the member that would lock out the attacker and update the member's leaf to a new, uncompromised state.

draft-ietf-mls-protocol.md
bifurcation:
```suggestion
users to detect this kind of misbehavior by the AS.  It is also possible to 
construct schemes in which the various clients owned by a user vouch
for each other, e.g., by signing each others' keys.
```

@kkohbrok - If you had a citation, that would be helpful.

bifurcation (approved):
Overall, these are good additions.  If we can get the comments addressed, I think it's good to merge.

bifurcation (approved):

PR #874, opened 2023-03-07
Add some more explanatory text to the beginning.

ekr:
This removes the following graf:

   The DS and AS may also apply additional policies to MLS operations to obtain
   additional security properties.  For example, MLS enables any participant to add
   or remove members of a group; a DS could enforce a policy that only certain
   members are allowed to perform these operations.  MLS authenticates all members
   of a group; a DS could help ensure that only clients with certain types of
   credential are admitted. MLS provides no inherent protection against denial of
   service; A DS could also enforce rate limits in order to mitigate
   these risks.

My sense is this belongs in Security Considerations, not here.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -784,7 +784,7 @@ KeyPackage and Welcome messages are used to initiate a group or introduce new
 members, so they are exchanged between group members and clients not yet in the
 group. A client publishes a KeyPackage via the DS, thus enabling other
 clients to add it to groups. When a group member wants to add a new member
-to a group it uses the new member's KeyPackage to add the new member to 
+to a group it uses the new member's KeyPackage to add the new member to
 the group and construct a Welcome message with which the new member can
 initialize its local state.
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -783,9 +783,10 @@ shared cryptographic state from one epoch to another by exchanging MLS messages.
 KeyPackage and Welcome messages are used to initiate a group or introduce new
 members, so they are exchanged between group members and clients not yet in the
 group. A client publishes a KeyPackage via the DS, thus enabling other
-clients to add it to groups. When a group member wants to add a client
-to a group it uses its KeyPackage to construct a Welcome message to
-send to the new member.
+clients to add it to groups. When a group member wants to add a new member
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -734,10 +709,31 @@ such an overflow condition occurs.
...
 
+MLS is designed to operate in the context described in
+{{?I-D.ietf-mls-architecture}}. In particular, we assume that the following
+services are provided:
+
+* An Authentication Service (AS) that enables group members to authenticate the
+  credentials presented by other group members.
+
+* A Delivery Service (DS) that routes MLS messages among the participants in the
+  protocol.
```
bifurcation, line 720: I would probably keep the sub-bullets describing the usual functions of the DS.

draft-ietf-mls-protocol.md
```diff
@@ -768,7 +764,10 @@ shared cryptographic state from one epoch to another by exchanging MLS messages:
 
 KeyPackage and Welcome messages are used to initiate a group or introduce new
 members, so they are exchanged between group members and clients not yet in the
-group.
+group. A client publishes a KeyPackage via the DS, thus enabling other
+clients to add it to groups. When a group member wants to add a client
+to a group it uses its KeyPackage to construct a Welcome message to
+send to the new member.
```
bifurcation, line 770:
```suggestion
clients to add it to groups. When a group member wants to add a new member
to a group it uses the new member's KeyPackage to add the new member to 
the group and construct a Welcome message with which the new member can
initialize its local state.
```

draft-ietf-mls-protocol.md
bifurcation:
```suggestion
* A Delivery Service (DS) that routes MLS messages among the participants in the
  protocol.
  
      * Pre-publication of KeyPackage objects for clients
      * Delivery of Proposal and Commit messages to members of a group
      * Delivery of Welcome messages to new members of a group
      * Sequencing of Commit messages (see {{sequencing}})
```

(Also incorporating some changes from #872)

ekr: I mean it's up to you, but I actually think this is a regression, because it has a lot of forward references.

bifurcation (approved):
This looks good to me.  Please do move the removed paragraph to the security considerations, either here or in #873.

raphaelrobert (approved):

PR #876, opened 2023-03-14
Remove protocol version from KeyPackage.

Bren2010:
The protocol version is already explicitly provided by the enclosing MLSMessage, and signed by SignWithLabel.


bifurcation:
Closing this after discussion with @raphaelrobert.  While in principle, a few bytes could be saved, it's a very minor improvement, and not enough to justify changing a signed structure this late in the process.

PR #877, opened 2023-03-24
Resolve ambiguities found in interop testing

bifurcation:
@mulmarta and I have completed interop testing that covers all of the functionality in the spec.  We found a couple of places where there were ambiguities in the spec that led to different implementation decisions, and in the process of resolving those, found a couple of errors / omissions in drarft-19.

* The current PSK proposal validation forbids `resumption` with `usage` other than `applications`.  The branch and reinit sections call for the creation of PSK proposals with types `branch` and `reinit`.  Given the text in the PSK section, these proposals would be invalid.  This PR explicitly allows `branch` and `reinit` in the specific context of those operations.  (This is a little pro-forma, since the PSK proposals are only seen by the committer.  But it seems good to allow the committer to have consistent PSK proposal validation logic.)

* The text says that Add, Remove, and ReInit proposals can be sent by external senders, but is silent about other proposal types.  This PR explicitly indicates which proposal types may be originated by external senders, and adds an IANA registry column to track.

* GREASE values for proposal types, credential types, and extension types are defined for use in Capabilities, but nothing currently forbids their use elsewhere.  This PR forbids their use in actual Proposal or Credential objects, and in GroupContext extensions.

* The `msg_type` field of the Proposal object was inconsistently referred to as both `msg_type` and `proposal_type`.  This PR changes it to be `proposal_type` everywhere.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -4013,8 +4013,13 @@ struct {
 
 A PreSharedKey proposal is invalid if any of the following is true:
 
-* The `psktype` in the PreSharedKeyID struct is set to `resumption` and
-  the `usage` is `reinit` or `branch`.
+* The PreSharedKey proposal is not being processed as part of a reinitialization
+  of the group (see {{reinitialization}}), and the PreSharedKeyID has `psktype`
+  set to `resumption` and `usage` set to `reinit`.
+
+* The PreSharedKey proposal is not being processed as part of a subgroup
+  branching operation (see {{subgroup-branching}}), and the PreSharedKeyID has
+  `psktype` set to `resumption` and `usage` set to `branch`.
 
 * The `psk_nonce` is not of length `KDF.Nh`.
 
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3914,9 +3914,6 @@ struct {
 } Proposal;
 ~~~
 
-A Proposal object whose `proposal_type` is one of the GREASE values defined in
-{{grease}} MUST be rejected as malformed.
-
 On receiving a FramedContent containing a Proposal, a client MUST verify the
 signature inside FramedContentAuthData and that the `epoch` field of the enclosing
 FramedContent is equal to the `epoch` field of the current GroupContext object.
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-protocol.md
```diff
@@ -5139,6 +5159,15 @@ selection of values chosen from these GREASE values:
 * `KeyPackage.extensions`
 * `GroupInfo.extensions`
 
+GREASE values MUST NOT be included in the `proposal_type` field of a Proposal
+object or the `credential_type` field of a Credential object.  A Proposal or
+Credential containing a GREASE value in one of these fields MUST be rejected as
```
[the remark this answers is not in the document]
bifurcation: In order for that to be true, though, it has to be the case the the proposal will be rejected.  This text just specifies the sender and receiver sides ("MUST NOT be included [by senders]" ... "MUST reject").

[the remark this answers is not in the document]
DavidSchinazi: What's the receiver behavior with unknown values here? Ignore or reject? If it's reject then you don't need to say anything about rejecting GREASE

draft-ietf-mls-protocol.md
```diff
@@ -5880,39 +5910,42 @@ Template:
...
-| 0xF000  - 0xFFFF | Reserved for Private Use | - | -    | RFC XXXX |
+| Value            | Name                     | R | Ext | Path | Ref      |
+|:-----------------|:-------------------------|:--|:----|:-----|:---------|
+| 0x0000           | RESERVED                 | - | -   | -    | RFC XXXX |
+| 0x0001           | add                      | Y | Y   | N    | RFC XXXX |
+| 0x0002           | update                   | Y | N   | Y    | RFC XXXX |
+| 0x0003           | remove                   | Y | Y   | Y    | RFC XXXX |
+| 0x0004           | psk                      | Y | Y   | N    | RFC XXXX |
+| 0x0005           | reinit                   | Y | Y   | N    | RFC XXXX |
+| 0x0006           | external_init            | Y | N   | Y    | RFC XXXX |
```
raphaelrobert, line 5931: This one is problematic because an external init is technically an external proposal. This is at odds with 12.1.8, but I think the problem is there, not here. In my mind, the new member proposal is not an external proposal and should have its own category. That would also reflect that it has its own sender type.
Tl;DR: Leave this as is but rewrite 12.1.8.

raphaelrobert (changes requested):

draft-ietf-mls-protocol.md
Bren2010: ExternalInit doesn't have the `external` SenderType, it should be `new_member_commit` if I'm understanding correctly

bifurcation: @Bren2010 is correct.  The concept here is:

* `SenderType::member` can send any proposal
* `SenderType::new_member_proposal` can send Add
* `SenderType::new_member_commit` can only send an External Commit (with proposals inline)
* `SenderType::external` can send only what is specified here

raphaelrobert (approved):

seanturner (approved):

PR #878, opened 2023-04-26
Minor clarifications for AUTH48

bifurcation:
This PR includes clarifications on a few minor points that were found during interoperability testing.

1. The prose description of the `info` parameter to HPKE conflicts with the pseudocode.  This PR deletes the conflicting prose.
2. When `LeafNode.extensions` contains GREASE extensions, those extensions have to be reflected in `LeafNode.capabilities.extensions`, or else the leaf note is invalid.  This PR adds a cautionary note to this effect.
3. #875 allows GREASE extensions in `LeafNode.extensions`, but did not add `LN` to the IANA registrations for the GREASE extension values.  This PR adds that notation.

We plan to hold this PR open until AUTH48 begins, then apply the changes in AUTH48.

### DIFF
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -3284,9 +3284,6 @@ context = SetupBaseR(kem_output, external_priv, "")
 init_secret = context.export("MLS 1.0 external init secret", KDF.Nh)
 ~~~
 
-In both cases, the `info` input to HPKE is set to the GroupInfo for the
-previous epoch, encoded using the TLS serialization.
-
 ## Pre-Shared Keys
 
 Groups that already have an out-of-band mechanism to generate
--- a/draft-ietf-mls-protocol.md
+++ b/draft-ietf-mls-protocol.md
@@ -2321,13 +2321,34 @@ described in {{credentials}}.
 
 The `capabilities` field indicates the protocol features that the client
 supports, including protocol versions, ciphersuites, credential types,
-non-default proposal types, and non-default extension types.
-Proposal and extension types defined in this document are considered "default" and thus MAY
-not be listed, while any credential types the application wishes to use MUST
-be listed. The types of any non-default extensions that appear in the `extensions` field of a LeafNode
+non-default proposal types, and non-default extension types.  The following
+proposal and extension types are considered "default" and MUST NOT be
+listed:
+
+* Proposal types:
+  * 0x0001 - `add`
+  * 0x0002 - `update`
+  * 0x0003 - `remove`
+  * 0x0004 - `psk`
+  * 0x0005 - `reinit`
+  * 0x0006 - `external_init`
+  * 0x0007 - `group_context_extensions`
+* Extension types:
+  * 0x0001 - `application_id`
+  * 0x0002 - `ratchet_tree`
+  * 0x0003 - `required_capabilities`
+  * 0x0004 - `external_pub`
+  * 0x0005 - `external_senders`
+
+There are no default values for the other fields of a capabilities object.  The
+client MUST list all values for the respective parameters that it supports.
+
+The types of any non-default extensions that appear in the `extensions` field of a LeafNode
 MUST be included in the `extensions` field of the `capabilities` field, and the
 credential type used in the LeafNode MUST be included in the `credentials` field
-of the `capabilities` field.  As discussed in {{extensibility}}, unknown values
+of the `capabilities` field.
+
+As discussed in {{extensibility}}, unknown values
 in `capabilities` MUST be ignored, and the creator of a `capabilities` field
 SHOULD include some random GREASE values to help ensure that other clients correctly
 ignore unknown values.
@@ -5161,6 +5182,11 @@ For the KeyPackage and GroupInfo extensions, the `extension_data` for GREASE
 extensions MAY have any contents selected by the sender, since they will be
 ignored by a correctly-implemented receiver.  For example, a sender might
 populate these extensions with a randomly-sized amount of random data.
+
+Note that any GREASE values added to `LeafNode.extensions` need to be reflected
+in `LeafNode.capabilities.extensions`, since the LeafNode validation process
+described in {{leaf-node-validation}} requires that these two fields be
+consistent.
 
 GREASE values MUST NOT be sent in the following fields, because an unsupported
 value in one these fields (including a GREASE value), will cause the enclosing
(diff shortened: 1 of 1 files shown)


Bren2010 (approved):

dconnolly:
> 2\. When `LeafNode.extensions` contains GREASE extensions, those extensions have to be reflected in `LeafNode.capabilities.extensions`, or else the leaf note is invalid.  This PR adds a cautionary note to this effect.

Not blocking these changes but does this indicate duplication amongst `LeafNode`? It's probably too late to change but

bifurcation:
@dconnolly - It's a good question, but the duplication is small.  The only thing that's duplicated is the 2-byte extension type value.  In addition to sending the GREASE extension with type 0xXXXX, you also need to say "I support extensions of type 0xXXXX".

dconnolly:
> @dconnolly - It's a good question, but the duplication is small.  The only thing that's duplicated is the 2-byte extension type value.  In addition to sending the GREASE extension with type 0xXXXX, you also need to say "I support extensions of type 0xXXXX".

Got it, that makes sense and is as you say small. Thanks!

dconnolly (approved):

bifurcation:
While responding to @dconnolly, though, I got to looking at the relevant text, and suspect we might have yet another minor issue:

> Proposal and extension types defined in this document are considered "default" and thus need not be listed [...]  Extensions that appear in the extensions field of a LeafNode MUST be included in the extensions field of the capabilities field [...]

That second sentence reads to me like extensions types must appear in `capabilities` **even if they are default types**.  That seems incorrect to me.  Propose clarifying to "Extensions of non-default types that appear in the extensions field..."  You would still have to GREASE consistently, but it would save a few bytes for default extension.  But maybe adding an exception to the logic isn't worth it.  @mulmarta @Bren2010 wdyt?

mulmarta:
@bifurcation I agree that the clarification you propose reflects what I always understood. Even more clear may be something like "Non-default types of extensions that appear in the extensions field of a LeafNode..." -- this indicates that types, not full extensions are copied. (FWIW the impact is quite small, because the only default leaf node extension is application id.)

Bren2010:
I agree it's inconsistent, though I don't have a strong opinion on which way we go. This sentence also stands out:

> The capabilities field indicates what protocol versions, ciphersuites, extensions, credential types, and non-default proposal types are supported by a client.

(Extensions isn't covered by "non-default")

kkohbrok:
The PR looks good to me. I support the solution approach proposed by @mulmarta. The inconsistency pointed out by @Bren2010 we can probably get rid of by rewriting as follows:

> The capabilities field indicates what protocol versions, ciphersuites, credential types, and non-default proposal and extension types are supported by a client.

bifurcation:
Thanks @mulmarta @Bren2010 @kkohbrok.  I updated the relevant sentences to clarify, and I also removed the paragraph about GREASE-ing consistently.  Under the theory that since GREASE extensions are defined in this document, they are "default", and thus don't have to be listed in `capabilities.extensions` (though they MAY be listed).

franziskuskiefer:
> Under the theory that since GREASE extensions are defined in this document, they are "default", and thus don't have to be listed in `capabilities.extensions` (though they MAY be listed).

We ran into this issue interopping between OpenMLS and mlspp because OpenMLS was checking for the GREASE values. But I agree that it is more consistent when GREASE values don't need to be listed as capabilities (since they're "default").

raphaelrobert:
> Under the theory that since GREASE extensions are defined in this document, they are "default", and thus don't have to be listed in `capabilities.extensions` (though they MAY be listed).

Wouldn't that defeat the purpose of GREASE values? If they are considered "default", it means they need to be hard-coded on the receiving side. Isn't that what should be avoided?

DavidSchinazi:
(Hi! GREASE enthusiast here) I agree with @raphaelrobert. For GREASE codepoints to work as intended, they need to behave exactly like an unknown optional extension. It would be best to specify that the clause `Proposal and extension types defined in this document are considered "default"` does not apply to GREASE values.

franziskuskiefer:
I generally with the GREASE sentiment. But any more exceptions here make this really hard to get right.
I'd be in favour of dropping the notion of "default" extensions all together instead.

bifurcation:
I'm inclined to agree, but it just seems like bloat to have to list, e.g., the ratchet tree extension.  I don't think we should consider support for those extensions optional, so they would all appear there.  What if instead we were explicit that implementations are REQUIRED to support extension types 0x0001-0x0005, and thus these are considered default?

> The following proposal types and extension types MUST be supported by implementations.  They are considered "default" and MAY be omitted from the corresponding fields in a Capabilities object.
> * [proposal types 0x0001-0x0007]
> * [extension types 0x0001-0x0005]

... and then we'll need to re-add the "consistent GREASE" paragraph.

draft-ietf-mls-protocol.md
```diff
@@ -2319,11 +2319,12 @@ member's public signing key. The `credential` field contains information
...
 described in {{credentials}}.
 
-The `capabilities` field indicates what protocol versions, ciphersuites,
-extensions, credential types, and non-default proposal types are supported by a client.
-Proposal and extension types defined in this document are considered "default" and thus need
+The `capabilities` field indicates the protocol features that the client
+supports, including protocol versions, ciphersuites, credential types,
+non-default proposal types, and non-default extension types.
+Proposal and extension types defined in this document are considered "default" and thus MAY
 not be listed, while any credential types the application wishes to use MUST
```
rohan-wire:
```suggestion
Proposal and extension types defined in this document are considered "default" and thus are not listed, while any credential types the application wishes to use MUST
```
If we don't have any reason to include default proposals and extensions, then they should never be included. There should be exactly one way to say that an MLS client implements the default.

rohan-wire (approved):

bifurcation:
In the latest commit:

* Be explicit about which proposals/extensions are considered "default"
* As @rohan-wire suggests, require that default values MUST NOT be listed
* Re-add the paragraph about consistent GREASE

I think this reflects all the feedback received so far, so I will plan to merge in the next 24hr or so unless there are objections.

mulmarta (approved):

rohan-wire (approved):

raphaelrobert (approved):

kkohbrok (approved):

DavidSchinazi (approved):

PR #879, opened 2023-05-30
AUTH48 review of RFC Editor's Proposed Changes

bifurcation:
This branch has my changes processing the RFC Editor's proposals.  

* Diffs on `draft-ietf-mls-protocol.md` reflect accepted changes (plus novel changes)
* Diffs on `rfc9420.authors.xml` to see rejected changes (plus novel changes)
* `QUESTIONS.md` shows my proposed answers to the RFC Editor's questions

The diff between the XML generated from Markdown and `rfc9420.authors.xml` is not trivial, but it reflects things that are difficult / impossible to reproduce in Markdown:

* Coalesced cross-references, e.g., "Sections 2 and 3" instead of "Section 2 and Section 3"
* One instance of `<blockquote>`
* Some references to email addresses that are styled to appear in a certain way
* Some minutia in the media type registration and references

To review the diff yourself (on a Mac, with vimdiff):

```
> git pull origin auth48-through-sec5
> git checkout auth48-through-sec5
> cd auth48
> make diff
```


draft-ietf-mls-protocol.md
```diff
@@ -1049,8 +641,7 @@ A              B     ...      Z          Directory        Channel
 |              |              |              |              |
 ~~~
 {: #update-flow title="Client B proposes to update its key, and client A commits the
-proposal.  As a result, the keys for both B and A updated, so the group has
-post-compromise security with respect to both of them."}
```
Bren2010, line 1053: "As a result" was supposed to be moved into the body of the text I believe

draft-ietf-mls-protocol.md
```diff
@@ -1236,23 +827,23 @@ replaced with a freshly generated instance.
...
+children, and the descendants of its children.  We say a tree
 _contains_ a node if that node is a descendant of the root of the tree,
 or if the node itself is the root of the tree. Nodes are _siblings_ if they share the same parent.
 
 A _subtree_ of a tree is the tree given by any node (the _head_ of the
 subtree) and its descendants. The _size_ of a tree or subtree is the
 number of leaf nodes it contains.  For a given parent node, its _left
-subtree_ is the subtree with its left child as head (respectively
-_right subtree_).
+subtree_ is the subtree with its left child as head and the its
```
Bren2010:
```suggestion
subtree_ is the subtree with its left child as head and its
```

auth48/rfc9420.authors.xml
```diff
@@ -1127,37 +1043,21 @@ replaced with a freshly generated instance.</t>
...
-   For a given parent node, its left subtree is the subtree with its 
-   left child as the head (and respectively, its right subtree has its 
-   right child as the head).
--->
-For a given parent node, its <em>left
-subtree</em> is the subtree with its left child as head (respectively
-	<em>right subtree</em>).</t>
-        <!-- Note: subscript or superscript may be needed -->
+number of leaf nodes it contains.  For a given parent node, its <em>left
+subtree</em> is the subtree with its left child as head and the its
```
Bren2010:
```suggestion
subtree</em> is the subtree with its left child as head and its
```

draft-ietf-mls-protocol.md
```diff
@@ -1269,22 +860,22 @@ algorithms, as long as they produce correct protocol messages.
 ### Ratchet Tree Nodes
 
 Each leaf node in a ratchet tree is given an _index_ (or _leaf index_), starting
-at `0` from the left to `2^d - 1` at the right (for a tree with `2^d` leaves). A tree
-with `2^d` leaves has `2^(d+1) - 1` nodes, including parent nodes.
+at 0 from the left to 2<sup>d - 1</sup> at the right (for a tree with 2<sup>d</sup> leaves). A tree
```
Bren2010:
```suggestion
at 0 from the left to 2<sup>d</sup>-1 at the right (for a tree with 2<sup>d</sup> leaves). A tree
```

auth48/rfc9420.authors.xml
```diff
@@ -1171,9 +1071,11 @@ trees.  An implementation may use any tree representation and associated
 algorithms, as long as they produce correct protocol messages.</t>
         <section anchor="ratchet-tree-nodes">
           <name>Ratchet Tree Nodes</name>
-          <!-- Note: subscript or superscript may be needed -->
-          <t>Each leaf node in a ratchet tree is given an <em>index</em> (or <em>leaf index</em>), starting at <tt>0</tt> from the left to <tt>2^d - 1</tt> at the right (for a tree with <tt>2^d</tt> leaves). A tree with <tt>2^d</tt> leaves has <tt>2^(d+1) - 1</tt> nodes, including parent nodes.</t>
-          <t>Each node in a ratchet tree is either <em>blank</em> (containing no value) or holds an HPKE public key with some associated data:</t>
+          <t>Each leaf node in a ratchet tree is given an <em>index</em> (or <em>leaf index</em>), starting
+at 0 from the left to 2<sup>d - 1</sup> at the right (for a tree with 2<sup>d</sup> leaves). A tree
```
Bren2010:
```suggestion
at 0 from the left to 2<sup>d</sup>-1 at the right (for a tree with 2<sup>d</sup> leaves). A tree
```

draft-ietf-mls-protocol.md
```diff
@@ -1552,56 +1146,60 @@ DecryptWithLabel(PrivateKey, Label, Context, KEMOutput, Ciphertext) =
...
 label = "MLS 1.0 " + Label;
 context = Context;
 ~~~
 
-Here, the functions `SealBase` and `OpenBase` are defined {{RFC9180}}, using the
-HPKE algorithms specified by the group's ciphersuite.  If MLS extensions
-require HPKE encryption operations, they should re-use the EncryptWithLabel
-construction, using a distinct label.  To avoid collisions in these labels, an
-IANA registry is defined in {{mls-public-key-encryption-labels}}.
+Here, the functions `SealBase` and `OpenBase` are defined {{Section 6.1 of
```
Bren2010:
```suggestion
Here, the functions `SealBase` and `OpenBase` are defined in {{Section 6.1 of
```

draft-ietf-mls-protocol.md
```diff
@@ -4942,18 +4540,18 @@ traversal of the ratchet tree. Each node is listed between its left subtree and
...
-right until it has a length of the form `2^(d+1) - 1`, adding the minimum number
+right until it has a length of the form 2<sup>d+1</sup> - 1, adding the minimum number
 of blank values possible.  (Obviously, this may be done "virtually", by
 synthesizing blank nodes when required, as opposed to actually changing the
 structure in memory.)
 
 The leaves of the tree are stored in even-numbered entries in the array (the
 leaf with index `L` in array position `2*L`). The root node of the tree is at
-position `2^d - 1` of the array. Intermediate parent nodes can be identified by
+position 2<sup>d - 1</sup> of the array. Intermediate parent nodes can be identified by
```
Bren2010:
```suggestion
position 2<sup>d</sup>-1 of the array. Intermediate parent nodes can be identified by
```

draft-ietf-mls-protocol.md
```diff
@@ -5703,7 +5301,7 @@ All of these registries should be under a heading of "Messaging Layer Security",
 and assignments are made via the Specification Required policy {{!RFC8126}}. See
 {{de}} for additional information about the MLS Designated Experts (DEs).
 
-RFC EDITOR: Please replace XXXX throughout with the RFC number assigned to
+RFC EDITOR: Please replace 9420 throughout with the RFC number assigned to
 this document
```
Bren2010:
```suggestion
```

Bren2010 (commented):
This is a huge amount of work, thank you!

draft-ietf-mls-protocol.md
bifurcation: Really good catch, thanks.

draft-ietf-mls-protocol.md
bifurcation: This got reworded as it moved.

katrielalex:
Wow this is a nice set of changes. Thank you both!

I’ll do an independent pass on the full protocol doc but let me not tie that to this review

draft-ietf-mls-protocol.md
```diff
@@ -3754,23 +3352,24 @@ are then added to the group using the usual Add/Commit mechanism.
...
+Group IDs SHOULD be constructed in such a way that there is an overwhelmingly low
 probability of honest group creators generating the same group ID, even without
-assistance from the Delivery Service. For example, by making the group ID a
+assistance from the Delivery Service. This can be done, for example, by making the group ID a
 freshly generated random value of size `KDF.Nh`. The Delivery Service MAY
 attempt to ensure that group IDs are globally unique by rejecting the creation
 of new groups with a previously used ID.
 
-The creator of a group MUST take the following steps to initialize the group:
+To initialize a group, the creator of the group <bcp14>MUST</bcp14> take the
```
Bren2010: Why is this the only place where `<bcp14>` happened?

draft-ietf-mls-protocol.md
```diff
@@ -5552,14 +5156,14 @@ versions of the ratchet tree, as this prevents old group secrets from being
 re-derived. Forward secrecy *within* an epoch is provided by deleting message
 encryption keys once they've been used to encrypt or decrypt a message.
 Note that group secrets and message encryption keys are shared by the
-group, and thus there is a risk to forward secrecy as long as any
+group.  There is  thus a risk to forward secrecy as long as any
```
Bren2010:
```suggestion
group.  There is thus a risk to forward secrecy as long as any
```

draft-ietf-mls-protocol.md
```diff
@@ -5573,12 +5177,12 @@ If two members' leaf nodes have the same signature key, for example, then the
 data origin authentication properties afforded by signatures within the group
 are degraded.
 
-Uniqueness of keys in leaf nodes is assured by explicit checks on leaf nodes
-being added to the tree by Add or Update proposals, or in the `path` field of a
+Uniqueness of keys in leaf nodes is assured by explicitly checking each leaf node
+being added to the tree by in Add proposal, in an Update proposals, or in the `path` field of a
```
Bren2010:
```suggestion
being added to the tree by in Add proposals, in Update proposals, or in the `path` field of a
```

Bren2010 (approved):

draft-ietf-mls-protocol.md
```diff
@@ -948,11 +538,11 @@ A                B                C            Directory       Channel
...
 
-{{create-flow}} shows how these prepublished KeyPackages are used to create a group.
-When a client A wants to establish a group with B and C, it first initializes a
+{{create-flow}} shows how these pre-published KeyPackages are used to create a group.
+When client A wants to establish a group with clients B and C, it first initializes a
 group state containing only itself and downloads KeyPackages for B and C. For
-each member, A generates an Add and Commit message adding that member, and
-broadcasts them to the group. It also generates a Welcome message and sends this
+each member, A generates an Add and Commit message to add that member and then
+broadcasts them to the group. Client A also generates a Welcome message and sends it
```
rohan-wire:
```suggestion
broadcasts it to the group. Client A also generates a Welcome message and sends it
```

"it" is referring to the Commit message with Add proposal that A generates.

In general, I find it awkward talking about Add and Commit messages. This is a single message (the Commit) with an enclosed Add proposal.

draft-ietf-mls-protocol.md
```diff
@@ -999,27 +589,29 @@ A              B              C          Directory            Channel
 {: #create-flow title="Client A creates a group with clients B and C"}
 
 Subsequent additions of group members proceed in the same way.  Any
-member of the group can download a KeyPackage for a new client
-and broadcast Add and Commit messages that the current group will use to update
-their state, and a Welcome message that the new client can use to
+member of the group can download a KeyPackage for a new client,
+broadcast Add and Commit messages that the current group will use to update
```
rohan-wire, line 593: Again, "Add and Commit messages" is awkward

draft-ietf-mls-protocol.md
```diff
@@ -1611,28 +1209,28 @@ value = value;
...
 ## Credentials
 
 Each member of a group presents a credential that provides one or more
-identities for the member, and associates them with the member's signing key.
+identities for the member and associates them with the member's signing key.
 The identities and signing key are verified by the Authentication Service in use
 for a group.
 
-It is up to the application to decide which identifier or identifiers to use at
+It is up to the application to decide which identifier(s) to use at
```
rohan-wire: I think "identifier or identifiers" is clearer.

draft-ietf-mls-protocol.md
```diff
@@ -3655,17 +3253,17 @@ Other values may be retained (not consumed):
 # Key Packages
 
 In order to facilitate the asynchronous addition of clients to a
-group, key packages are pre-published that
+group, clients can pre-publish KeyPackage objects that
```
rohan-wire, line 3256:
```suggestion
group, clients pre-publish KeyPackage objects that
```
The "in order to facilitate" already makes the "can" redundant here.

draft-ietf-mls-protocol.md
```diff
@@ -4119,26 +3720,26 @@ group agree on the extensions in use.
 ### External Proposals
 
 Proposals can be constructed and sent to the group by a party
-that is outside the group in two cases. One case, indicated by an `external` SenderType
-is useful in cases where, for example, an automated service might propose to
-remove a member of a group who has been inactive for a long time, or propose adding
+that is outside the group in two cases. One case, indicated by the `external` SenderType
+is useful, for example, if an automated service might propose
+removing a member of a group who has been inactive for a long time, or propose adding
```
rohan-wire: this is still a mouthful of a sentence and reads worse than the original IMO. 

```suggestion
could be used by an automated service, for example, to remove a member
of a group who has been inactive for a long time, or to propose adding
```

draft-ietf-mls-protocol.md
```diff
@@ -5067,12 +4672,12 @@ all the members of the group that will process the Commit.
...
-In order to ensure that MLS provides meaningful authentication it is important
+In order to ensure that MLS provides meaningful authentication, it is important
 that each member is able to authenticate some identity information for each
 other member.  Identity information is encoded in Credentials, so this property
 is provided by ensuring that members use compatible credential types.
 
-The types of credential that may be used in a group is restricted to what all
+The only types of credential that may be used in a group are those that all
 members of the group support, as specified by the `capabilities` field of each
 LeafNode in the ratchet tree. An application can introduce new credential types
```
rohan-wire, line 4683: This is an unnecessary imposition and makes it impossible to add support for new credential types for some members of a group, and makes KeyPackage more complicated. If a specific credential type is required by all the members of the group, it should be included in `required_capabilities`.  The specific use case of unknown credentials was discussed in various interim meetings and was a major motivator for moving the signature out of the credential and into the LeafNode.

```suggestion
If a specific credential type is included in the `required_capabilities` extension
in the GroupContext, then all members of the group MUST support that credential
type, as specified by the `capabilities` field of each LeafNode in the ratchet tree.
Credential types that are unknown to a member and not included the
`required_capabilities` extension are ignored by the member. This allows a subset
of members to introduce additional functionality without requiring universal
support for a new credential type. An application can introduce new credential types
```

draft-ietf-mls-protocol.md
```diff
@@ -5180,24 +4785,24 @@ selection of values chosen from these GREASE values:
...
 described in {{leaf-node-validation}} requires that these two fields be
 consistent.
 
 GREASE values MUST NOT be sent in the following fields, because an unsupported
-value in one these fields (including a GREASE value), will cause the enclosing
+value in one these fields (including a GREASE value) will cause the enclosing
 message to be rejected:
 
 * `Proposal.proposal_type`
 * `Credential.credential_type`
```
rohan-wire, line 4802: suggest removing credential_type from the list if my previous change is accepted.

```suggestion
```

draft-ietf-mls-protocol.md
rohan-wire:
```suggestion
Uniqueness of keys in leaf nodes is assured by explicitly checking each leaf node
being added to the tree in any Add proposals, in any Update proposals, or in the `path` field of a
```

draft-ietf-mls-protocol.md
```diff
@@ -6085,22 +5687,23 @@ the column layout of the registry.
...
 |:------|:------------|:----------|
 | (N/A) | (N/A)       | (N/A)     |
+{: title="MLS Exporter Labels Registry" }
 
 ## MLS Designated Expert Pool {#de}
 
 Specification Required {{RFC8126}} registry requests are registered
-after a three-week review period on the MLS DEs' mailing list:
-<mls-reg-review@ietf.org>, on the advice of one or more of the MLS DEs. However,
+after a three-week review period on the MLS DEs' mailing list
```
rohan-wire:
```suggestion
after a three-week review period on the MLS Designated Expert (DE) mailing list
```

rohan-wire (changes requested):
Some small changes proposed, and the issue of optional credential types in a group.

draft-ietf-mls-protocol.md
raphaelrobert: I remember the discussions a bit differently, namely that we included the signature key in the leaf node so that we can have easier extensibility for credentials, but *not* that we can now have arbitrary combinations of credentials. I'm worried that this weakens authentication guarantees when clients can only verify a signature but no longer correlate it to an identity. It's also not clear to me why we can't do what we do with other extensions: first make sure every client in the group supports it as advertized in its capabilities and *only then* make it a required capability before using it.

draft-ietf-mls-protocol.md
raphaelrobert: Agreed, I think this terminology got carried over from the times when we had actual Add messages

draft-ietf-mls-protocol.md
mulmarta: The part I'm worried about here is `Credential types that are unknown to a member and not included the
required_capabilities extension are ignored by the member.` This seems to conflict with the fact that every credential ` MUST be validated with the AS`, since a member can't verify a credential it doesn't support? Note that required capabilities are optional, so credentials must be verified even without them.

raphaelrobert (approved):
Thanks @bifurcation for setting all this up and answering the questions. Thanks also to the other reviewers, unless where I said otherwise I agree with the suggestions.

This looks good to me, I only found one nit: The document is inconsistent regarding ciphersuites. In prose we use "ciphersuite", but the C struct is spelled "CipherSuite" as if it was a contraction of two words. The same applies to fields spelled "cipher_suite". I propose we harmonize it by renaming the C struct to "Ciphersuite" and the fields to "ciphersuite". I can do a PR against the `auth48` or `auth48-through-sec5` branch, @bifucraction let me know what you prefer.

draft-ietf-mls-protocol.md
kkohbrok: I second @raphaelrobert and @mulmarta. Allowing credentials that are not supported by a subset of group members and which they can thus only ignore seems like a very bad idea from an authentication standpoint. @rohan-wire can you maybe elaborate what kind of use-case you had in mind? Because I'm having trouble thinking of one. It would have to be a situation, where different applications are interoperating and where clients are okay with group members that they cannot authenticate at all. For such a group, one might was well use basic credentials.

rohan-wire: MLS currently supports `basic` and `x509`. It does not require `x509`. So, the "default" case is that groups support `basic` which has no authentication (and is similar to what the industry does now for DoubleRatchet protocols). This can be mitigated with key transparency or manual verification, but is mostly just swept under the rug.

Now imagine a vendor/provider introduces support for x509 credentials tomorrow. Do clients have to wait until they are in a group when 100% of the members support this credential type before they include their credentials? 
- What harm does it cause to clients which only support `basic`? I cannot see any. They had no authentication before. They still have no authentication?
- Does it create an opportunity for bid-down?  No. because credential capabilities are signed in KeyPackages.
- Can clients who *do* support `x509` do something useful with those credentials? Yes. the subset of clients which support x509 can indicate to the end user which clients in the roster have been validated and can even put more prominent indications when everyone is using the credentials or when suddenly everyone is not.
- Do we lose the ability to require a specific credential type? No. a client can set `required_capabilities`.

Does that help?

raphaelrobert: You provide an example in which you claim there's no downside to your proposed changes. Here's a counter-example: Suppose a group is to use X.509 and X.509 only. With the proposed changes, a subset of the clients could start using BasicCredentials and the remainder of the clients would have to comply with that, since BasicCredentials would simply be an unknown credential to them.
We would need better reasoning about the relaxation, however I'm doubtful that it's a good idea to do such a potentially impactful change during AUTH48, without the possibility to fully consider all consequences.

rohan-wire: Why wouldn't those clients just use `required_capabilities` if they only want to use X.509?
I also find it a bit preposterous the idea that an implementation would not understand what `basic` is, as opposed to administratively refusing to use it.

I am sympathetic about the AUTH48 comment, but that would indicate this is better fixed as an errata than an extension.

draft-ietf-mls-protocol.md
bifurcation: I think I added it before I realized that it could get added automatically.  In any case, removed.

draft-ietf-mls-protocol.md
bifurcation: Reworded to:

> ... each leaf node as it is added to the tree, whether in an Add proposal, in an Update proposal, or in the `path` field of a Commit.

That way everything is singular.  Also got rid of "by in".

draft-ietf-mls-protocol.md
bifurcation: Not necessarily!  You could send an Add Proposal alongside a Commit.  At this stage, I think it is a little clearer to address them separately, with inlining as an optimization discussed later.  Reworded to be a little more explicit, though:

> A generates an Add message and a Commit message to add that member and then broadcasts the two messages to the group

draft-ietf-mls-protocol.md
bifurcation: Reduced to "identifiers", seems cleaner.

draft-ietf-mls-protocol.md
bifurcation: Disagree.  Omitting the "can" implies that pre-publication is mandatory, which it isn't really.  In a more real-time setting, you could provide the KeyPackage at join time, to prompt the Add+Commit.  If you want to stretch things, you say that this is still "pre" in the sense that it's before the Add.  But that seems like a stretch.

bifurcation:
@raphaelrobert Good point.  I would actually go the other direction.  I would like to be consistent with TLS here, which appears to use "cipher suite" as two words throughout.  Thus "cipher suite" in prose, `CipherSuite`, and `cipher_suite`.  (I'm pretty sure we got the `cipher_suite` variable names via that copy/paste.)  I will implement the change in this direction, but in its own commit so it's easier to revert if we disagree :)

draft-ietf-mls-protocol.md
bifurcation: Without getting into the details of the discussion here -- this is a significant change for this late.  I would have been skeptical but open earlier in the process; at this point, it's clearly too big.  And it doesn't seem like anyone is in support besides @rohan-wire.  I'm going to decline to include this change unless the chairs tell me otherwise.

FWIW, I would also argue that this change should not be accepted as an erratum.  It is not an error in the spec, it is a deliberate decision by the working group.

That said, you could definitely write a small, no-content `allow_unknown_credentials` GroupContext extension that just says, "In this group, it's OK to use credentials that not everyone supports".  Then you just switch on the presence of that extension to turn off the credential support checking.  You still have an everyone-has-to-upgrade problem, but only with regard to that extension, not every credential type.

raphaelrobert:
> @raphaelrobert Good point. I would actually go the other direction. I would like to be consistent with TLS here, which appears to use "cipher suite" as two words throughout. Thus "cipher suite" in prose, `CipherSuite`, and `cipher_suite`. (I'm pretty sure we got the `cipher_suite` variable names via that copy/paste.) I will implement the change in this direction, but in its own commit so it's easier to revert if we disagree :)

I've been meaning to check what TLS does, since the term was coined in the context of SSL/TLS originally anyway. Makes perfect sense to align with that.

draft-ietf-mls-protocol.md
rohan-wire: Huh?
We are describing the canonical flow here, not prescribing normative behavior. Even if we were, A sends a Commit message which contains or references A's Add proposal. The important thing for the clarity of this sentence is the Commit message, and in 99% of cases, A will never **separately** send an Add proposal. The proposed rewording misleads the reader otherwise.

draft-ietf-mls-protocol.md
rohan-wire: I think the earlier phrase "In order to facilitate ..." makes it clear this is optional, but no biggie here.

